PR #10, opened 2018-02-12
enable tls

bajacondor:


bajacondor:
@buchgr:  Here's my initial shot at this.  Let me know what you think.

main.go
```diff
@@ -18,6 +18,10 @@ func main() {
 	maxSize := flag.Int64("max_size", -1,
 		"The maximum size of the remote cache in GiB. This flag is required.")
 	htpasswd_file := flag.String("htpasswd_file", "", "Path to a .htpasswd file. This flag is optional. Please read https://httpd.apache.org/docs/2.4/programs/htpasswd.html.")
+	tls_enabled := flag.Bool("tls_enabled", false, "Bool specifying wheather or not to start the server with tls.  If true, default port changes to 8443, and server_cert and server_key flags are requred.")
```
buchgr: I wonder if a flag magically changing the default value of another flag is a good idea, it seems it could confuse people? What do you think?

main.go
```diff
@@ -18,6 +18,10 @@ func main() {
...
+	server_cert := flag.String("server_cert", "", "Path to a pem encoded certificate file.  Required if tls_enabled is set to true.")
```
buchgr: `tls_cert`? It should be clear that this is a server certificate. Also I think `PEM` is typically spelled all uppercase.

main.go
```diff
@@ -18,6 +18,10 @@ func main() {
...
+	server_key := flag.String("server_key", "", "Path to a pem encoded key file.  Required if tls_enabled is set to true.")
```
buchgr: Same as above?

bajacondor:
Good comments.  Thank you.

buchgr:
Thanks @bajacondor

PR #13, opened 2018-02-14
Close request body

nicolov:
To avoid leaking file descriptors.

https://stackoverflow.com/questions/23928983/defer-body-close-after-receiving-response


buchgr:
Awesome as always! Thanks!

PR #17, opened 2018-03-07
add a test for LoadFilesIntoCache

nicolov:
Also move the function itself to the `Cache` type, since I think it makes more sense there.

Also add a `NumFiles()` method to the `Cache` interface to make the test easier (will probably also use it in the upcoming status page.

This should go in after my other PR.


buchgr:
thanks <3 LGTM

PR #18, opened 2018-03-07
two-stage docker build

nicolov:
This removes the go sdk from the final container, making it smaller. I also moved the inner port to 80, which is the default for HTTP. Users can remap it to whatever they want using Docker.

Basically just copy-pasted the Docker docs: https://docs.docker.com/develop/develop-images/multistage-build


buchgr:
sweet! thanks @nicolov. I ll add Bazel docker rules next, should make building and pushing images easier.

PR #20, opened 2018-03-07
linux-build.sh should get dependencies

buchgr:
building the docker container fails without go get.

@nicolov


nicolov:
We should probably use `dep` here, to get the correct versions of dependencies. I'll push a commit that does that instead.

buchgr:
ur right!

PR #22, opened 2018-03-07
install dependencies in docker build

nicolov:
I put the `dep` command in the docker file, instead of `linux-build.sh`, since people might want to handle dependencies in their own way? I don't really care a lot either way though.


buchgr:
good idea :-)

PR #25, opened 2018-03-09
Add a basic status page

nicolov:
`/status` prints some basic json with stats about the cache server.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,7 +32,7 @@ func main() {
 	e := cache.NewEnsureSpacer(0.95, 0.5)
 	h := cache.NewHTTPCache(*dir, *maxSize*1024*1024*1024, e)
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -32,6 +32,7 @@ func main() {
 	e := cache.NewEnsureSpacer(0.95, 0.5)
 	h := cache.NewHTTPCache(*dir, *maxSize*1024*1024*1024, e)
 
+	http.HandleFunc("/status", maybeAuth(h.StatusPageHandler, *htpasswd_file, *host))
```
buchgr: I wonder whether the status page should be password protected? What's a good argument for it?

nicolov: I thought it would be less surprising to users / secure by default. Will remove password protection.

PR #30, opened 2018-03-15
Add LRU implementation

nicolov:
Add a size-aware LRU cache that we can use to limit the on-disk cache size. This PR only adds the generic cache, but does not wire it in yet.

This PR includes my logging changes, do you want to review those first?


cache/utils_test.go
```diff
@@ -34,3 +35,10 @@ func createTmpCacheDirs(t *testing.T) string {
 
 	return path
 }
+
+// newSilentLogger returns a cheap logger that doesn't print anything, useful
+// for tests.
+func newSilentLogger() (l log.Logger) {
+	l = *log.New(ioutil.Discard, "", 0)
+	return
+}
```
buchgr: new line?

main.go
```diff
@@ -29,8 +30,10 @@ func main() {
 		return
 	}
 
+	accessLogger := log.New(os.Stdout, "", log.Ldate | log.Ltime | log.LUTC)
+	errorLogger := log.New(os.Stdout, "", log.Ldate | log.Ltime | log.LUTC)
 	e := cache.NewEnsureSpacer(0.95, 0.5)
-	h := cache.NewHTTPCache(*dir, *maxSize*1024*1024*1024, e)
+	h := cache.NewHTTPCache(*dir, *maxSize*1024*1024*1024, e, *accessLogger, *errorLogger)
```
buchgr: pass by reference?

cache/sizedlru/sizedlru.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"container/list"
+)
+
+type SizedItem interface {
+	Size() int64
+}
+
+// SizedLRU is an LRU cache that will keep its total size below maxSize by evicting
+// items. If maxSize is zero, eviction is disabled.
```
buchgr: I am thinking we might not even want to support this. IMHO there's no use case for an unbounded cache and one can always set maxSize to a really big value (or disk size). Wdyt?

cache/sizedlru/sizedlru.go
```diff
@@ -0,0 +1,111 @@
...
+	return &SizedLRU{
+		maxSize: maxSize,
+		ll:      list.New(),
+		cache:   make(map[interface{}]*list.Element),
+	}
+}
+
+// Add adds a (key, value) to the cache, evicting items as necessary. Add returns false (
+// and does not add the item) if the item size is larger than the maximum size of the cache.
+func (c *SizedLRU) Add(key Key, value SizedItem) (ok bool) {
```
buchgr: I am thinking ahead when using this actually implement eviction. We would need to return a slice of SizedItem that needs to be deleted from the disk?

cache/http.go
```diff
@@ -90,22 +105,36 @@ func ensureDirExists(path string) {
 func (h *httpCache) CacheHandler(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close()
 
+	// Helper function for logging responses
+	logResponse := func(code int) {
+		// Parse the client ip
+		clientIp, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			h.errorLogger.Fatal("Failed parsing RemoteAddr %s: %s", r.RemoteAddr, err.Error())
```
buchgr: Use `h.errorLogger.Fatalf`

buchgr: Actually, I am thinking using `.Fatal` here is wrong. A single request would crash the whole server (this is why go test fails on CI). Can we just log `r.RemoteAddr` as is and thus not even encounter this error case?

buchgr:
@nicolov can you please run `bazel run //:gazelle` to automatically generated a `BUILD` file for the `sizedlru` package.

cache/sizedlru/sizedlru.go
```diff
@@ -0,0 +1,111 @@
...
+	"container/list"
+)
+
+type SizedItem interface {
+	Size() int64
+}
+
+// SizedLRU is an LRU cache that will keep its total size below maxSize by evicting
+// items. If maxSize is zero, eviction is disabled.
+// SizedLRU is not thread-safe.
```
buchgr: Are you planning on wrapping every call with a mutex? I could see this becoming a bottleneck. Wdyt?

main.go
```diff
@@ -29,8 +30,10 @@ func main() {
 		return
 	}
 
+	accessLogger := log.New(os.Stdout, "", log.Ldate | log.Ltime | log.LUTC)
+	errorLogger := log.New(os.Stdout, "", log.Ldate | log.Ltime | log.LUTC)
```
buchgr: `os.Stderr`?

cache/sizedlru/sizedlru.go
```diff
@@ -0,0 +1,111 @@
+// Package sizedlru implements an LRU cache that bounds the total space occupied
+// by the items it contains.
+
+package sizedlru
```
buchgr: Maybe just call it lru?

cache/sizedlru/sizedlru.go
nicolov: Correct, but the critical section should only contain a few pointer swaps for the linked list and an hashtable lookup/update. This pales in comparison with all the work for setting up the HTTP connection and actually transfering the file. I'll be sure to test this with 3-4 gbit/s throughput.

cache/sizedlru/sizedlru.go
nicolov: a callback?

cache/sizedlru/sizedlru.go
buchgr: Yeah it's probably premature optimization. I am thinking we need benchmarks :-).

cache/http.go
```diff
@@ -40,13 +43,25 @@ type statusPageData struct {
 	ServerTime int64
 }
 
-// NewHTTPCache ...
-func NewHTTPCache(cacheDir string, maxBytes int64, ensureSpacer EnsureSpacer) HTTPCache {
+// NewHTTPCache returns a new instance of the cache.
+// accessLogger will print one line for each HTTP request to the cache.
+// errorLogger will print unexpected server errors. Inexistent files and malformed URLs will not
+// be reported.
+func NewHTTPCache(cacheDir string, maxBytes int64, ensureSpacer EnsureSpacer, accessLogger log.Logger, errorLogger log.Logger) HTTPCache {
```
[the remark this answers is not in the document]
nicolov: I tried to match what you did for the EnsureSpacer. I still haven't figured out a good rule of thumb.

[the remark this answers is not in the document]
nicolov: I'll just create an interface for it.

cache/http.go
nicolov: removed the Fatal call

main.go
nicolov: I had one job..

cache/sizedlru/sizedlru.go
nicolov: I actually moved it into the main package.

cache/sizedlru/sizedlru.go
nicolov: removed

main.go
nicolov: made it an interface

cache/sizedlru/sizedlru.go
nicolov: Each request would:
1. lock the mutex
2. get the element from the lru, updating the eviction list
3. unlock the mutex
4. actually serve the file from disk

I would expect that the locking would be relatively unexpensive.

nicolov:
I addressed the points in the review, with the exception of the eviction callback. I'll add it in a follow-up PR, since this is already a good amount of code to merge. Btw, I'm on ubuntu 18.04/Bazel 0.11 and can't get the tests to pass using bazel. I get an empty log, and nothing is streamed to stdout. go test works. Which versions are you on?

buchgr:
@nicolov I am on Google's Debian based Linux [1]. We are testing 14.04 and 16.04 but not 18.04 on our CI. I ll open an issue for Bazel to support 18.04.


[1] http://www.zdnet.com/article/google-moves-to-debian-for-in-house-linux-desktop/

Issue #34, opened 2018-04-03
Purging does not work


Yannic:
I was running your amazing cache with a very small cache size and noticed that new files were no longer added to the cache after some time. After reading the code, I think there are a few bugs in `EnsureSpacer`:

```go
d, err := os.Open(cache.Dir())
// ...
files, err := d.Readdir(-1)
// ...
for _, fileinfo := range files {
  name := fileinfo.Name()
  // ...
}
```

- It seems like we're leaking file descriptors here because `d` is never closed. I think I've seen some errors related to too many files open in log files, but I'm not 100% sure that it is related to this.
- `files` will always be a slice containing exactly `ac` and `cas`. I think what we want to do here is find the oldest files in the cache and delete them. This could be done with `filepath.Walk`, similar to what we do in `LoadExistingFiles()`, but since we know what files are stored in the cache, we could maybe find a more efficient solution.
- `fileinfo.Name()` returns the base name of a file, not the path (https://golang.org/pkg/os/#FileInfo). Unfortunately, there is no way to get the full path from FileInfo.

nicolov:
#36 should fix this, please reopen if it doesn't.

PR #35, opened 2018-04-03
Fix implementation of purge in EnsureSpacer

Yannic:
Closes: #34

### DIFF
--- a/cache/ensurespacer.go
+++ b/cache/ensurespacer.go
@@ -1,7 +1,6 @@
 package cache
 
 import (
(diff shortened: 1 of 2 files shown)


nicolov:
Thanks for taking a look at this. I have reworked the EnsureSpacer implementation to maximize disk utilization. I'll wait for Jakob to decide whether we want to merge this or my implementation.

buchgr:
Ty @Yannic ! This was an unfortunate race condition with @nicolov. :(

PR #36, opened 2018-04-05
Use LRU to evict individual items

nicolov:
Previously, files were deleted in bulk once the total cache size grew over the high water mark set in EnsureSpacer. This PR changes this behavior, and instead checks for the cache size limit before each requests. Just enough files are deleted to make space for the incoming items, thus improving the average disk utilization.

I moved the actual cache implementation away from the HTTP handler and into a separate `Cache` interface, with an eye towards making alternate backends (such as HTTP/multilevel/etc..).


cache/lru.go
```diff
@@ -20,27 +34,27 @@ type SizedLRU struct {
 	// SizedLRU will evict items as needed to maintain the total size of the cache
 	// below maxSize.
 	maxSize int64
+	onEvict EvictCallback
```
Yannic: Just one quick note here:
maps in go aren't safe to use from multiple goroutines https://golang.org/doc/faq#atomic_maps
Could we add a mutex (RW?) here and make sure we access the map only after acquiring a lock.

nicolov: Correct, as noted in the comment for `SizedLRU`. Since it's a basic datastructure, I think it should be the job of the client to protect it with locks for more flexibility. `fsCache` does so.

cache/http.go
```diff
@@ -131,49 +112,50 @@ func (h *httpCache) CacheHandler(w http.ResponseWriter, r *http.Request) {
...
 			return
 		}
-		http.ServeFile(w, r, cacheItem.absFilePath)
+
 		logResponse(http.StatusOK)
 	case http.MethodPut:
-		if h.cache.ContainsFile(cacheItem.absFilePath) {
-			h.discardUpload(w, r.Body)
-			logResponse(http.StatusOK)
+		if r.ContentLength == -1 {
```
buchgr: We might want to check for `Transfer Encoding: chunked` and simply disallow it. This would seem to be the most likely cause of a missing `Content Length` header.

cache/http.go
```diff
@@ -131,49 +112,50 @@ func (h *httpCache) CacheHandler(w http.ResponseWriter, r *http.Request) {
...
+			// We need the content-length header to make sure we have enough disk space.
+			msg := "put without content-length header"
```
buchgr: `Chunked transfer encoding is not supported for PUT`?

cache/http_test.go
```diff
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
-	"encoding/json"
+	//"encoding/json"
```
buchgr: delete?

cache/http_test.go
```diff
@@ -4,12 +4,14 @@ import (
...
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	//"sync"
```
buchgr: delete?

cache/cache.go
```diff
@@ -1,95 +1,258 @@
...
+	}
+
+	blobPath := c.pathForKey(key)
+	f, err := os.Open(blobPath)
+	if err != nil {
+		return
+	}
+	defer f.Close()
+
+	_, err = io.Copy(w, f)
```
buchgr: I think it's easiest if you just use `http.ServeFile` here as it also sets all the necessary headers. This is also why the `Content Length` header is not set and you had to uncomment the testing code.

cache/http.go
```diff
@@ -131,49 +112,50 @@ func (h *httpCache) CacheHandler(w http.ResponseWriter, r *http.Request) {
 
 	switch m := r.Method; m {
 	case http.MethodGet:
-		if !h.cache.ContainsFile(cacheItem.absFilePath) {
-			w.WriteHeader(http.StatusNotFound)
+		found, err := h.cache.Get(cacheKey, w)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			h.errorLogger.Printf("get %s: %s", cacheKey, err)
```
buchgr: `GET`?

cache/http.go
```diff
@@ -131,49 +112,50 @@ func (h *httpCache) CacheHandler(w http.ResponseWriter, r *http.Request) {
...
-				"The disk is full (%d/%d bytes used)", h.cache.CurrSize(), h.cache.MaxSize())
-			return
-		}
-		written, err := h.saveToDisk(r.Body, *cacheItem)
+
+		err := h.cache.Put(cacheKey, r.ContentLength, expectedHash, r.Body)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
-			h.errorLogger.Printf("Error saving file: %s", err.Error())
+			h.errorLogger.Printf("put %s: %s", cacheKey, err)
```
buchgr: `PUT`?

cache/http.go
```diff
@@ -131,49 +112,50 @@ func (h *httpCache) CacheHandler(w http.ResponseWriter, r *http.Request) {
...
-		w.WriteHeader(http.StatusOK)
+
 		logResponse(http.StatusOK)
 	case http.MethodHead:
-		if !h.cache.ContainsFile(cacheItem.absFilePath) {
-			http.Error(w, err.Error(), http.StatusNotFound)
+		ok, err := h.cache.Contains(cacheKey)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			h.errorLogger.Printf("head %s: %s", cacheKey, err)
```
buchgr: `HEAD`?

cache/cache.go
```diff
@@ -1,95 +1,258 @@
...
+		}
+	}
+
+	cache := &fsCache{
+		dir: dir,
+		mux: &sync.RWMutex{},
+		lru: NewSizedLRU(maxSizeBytes, onEvict),
+	}
+
+	cache.loadExistingFiles()
```
Yannic: Could we log here how many files we loaded into the cache?

cache/cache.go
nicolov: I added the missing header back. I'd rather not use `serveFile` since it couples the cache to the http request and has a bunch of magic for autodetecting types, etc..

cache/cache.go
nicolov: done

cache/http.go
nicolov: Mmm, how would that be better than complaining about the `Content-Length` header directly?

nicolov:
Thanks for the review. I've corrected all the points except the Content-Length bit.

buchgr:
Thanks for the work @nicolov :-)

It seems that garbage collection doesn't work with this CL. I am getting errors like
```
2018/04/12 12:28:07 cache.go:70: remove /usr/local/google/home/buchgr/.bazel-remote-cache/c/65d692a2366f6eedfa0ff959cf001e3b40e3c4b25d739c3c9bec1ac76da51d23: no such file or directory
2018/04/12 12:28:07 cache.go:70: remove /usr/local/google/home/buchgr/.bazel-remote-cache/as/c36e53c6be5d6f084c598b574c564b36baed64ca97a70f2c600847b268a6ae5e: no such file or directory
2018/04/12 12:28:07 cache.go:70: remove /usr/local/google/home/buchgr/.bazel-remote-cache/as/0ff649f2a7ffe122cbe9080bf9487bda152ecfddc9c5d61fc0fc80322148e982: no such file or directory
2018/04/12 12:28:07 cache.go:70: remove /usr/local/google/home/buchgr/.bazel-remote-cache/c/087fd5f8eee9c195ccdd7bcf97d4b06a0086757fef1588cf3b98f82ccc2b610d: no such file or directory
2018/04/12 12:28:07 cache.go:70: remove /usr/local/google/home/buchgr/.bazel-remote-cache/c/fafcf8ceff147793c6f04975c5c4631d3064e3cbe6287a7d1cc030fb2bdf0685: no such file or directory
2018/04/12 12:28:07 cache.go:70: remove /usr/local/google/home/buchgr/.bazel-remote-cache/as/f16bb7a37f919a6e1bc9f5a64c96e7a0e185c10de9048e05c57a456fd3c0ed65: no such file or directory
```

The path seems to be constructed incorrectly?

buchgr:
Thanks for the awesome work @nicolov and @Yannic !

cache/cache.go
```diff
@@ -1,95 +1,270 @@
...
+			blobPath := filepath.Join(dir, key.(string))
+			err := os.Remove(blobPath)
+			if err != nil {
+				log.Println(err)
+			}
+		}
+	}
+
+	cache := &fsCache{
+		dir: dir,
```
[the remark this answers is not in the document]
nicolov: I saw you did this already, thanks.

PR #40, opened 2018-04-26
Use port 8080 in Dockerfile

jerrymarino:
First off, thanks for putting this sweet project on GitHub - I've been testing it out quite a bit 😃 

Here's a small PR: The previous Dockerfile and readme point to port 8080, so patch uses 8080 instead of 80.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Usage of ./bazel-remote:
   -max_size int
 	The maximum size of the remote cache in GiB. This flag is required. (default -1)
   -port int
-	The port the HTTP server listens on (default 8080)
+	The port the HTTP server listens on (default 80)
   -tls_cert_file string
 	Path to a PEM encoded certificate file.  Required if tls_enabled is set to true.
   -tls_enabled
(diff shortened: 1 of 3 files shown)


jerrymarino:
@nicolov cool that makes sense to me. I've updated everything to point at port 80

README.md
```diff
@@ -42,7 +42,7 @@ We publish docker images to [DockerHub](https://hub.docker.com/r/buchgr/bazel-re
 
 ```bash
 $ docker pull buchgr/bazel-remote-cache
-$ docker run -v /path/to/cache/dir:/data -p 9090:8080 buchgr/bazel-remote-cache
+$ docker run -v /path/to/cache/dir:/data -p 9090:80 buchgr/bazel-remote-cache
```
jerrymarino: One minor suggestion, once this lands, it could be useful to update the related container on dockerhub.

buchgr:
I think it would be problematic to switch to port 80, as this would break existing users and apparently this thing has been pulled 3100 times from docker hub. So I think we should stick with `8080`, if there's no strong technical reason.

Issue #39, opened 2018-04-23
Usage as library


Yannic:
It would be great if we could add a way to use this cache as a library (i.e. to make integration into existing projects easier). As a side effect, this would mostly implement #12, except the configuration part, and make it easier for folks to implement their own storage mechanism on top of GCS, S3, ...

We would create a new package `httpcache` (to avoid conflicts with `net/http`), which has a function `New` that returns a `http.Handler`.

```go
type Options struct {
  // Logging, ...
}

// New creates a new Bazel cache which uses the provided StorageMechanism
// to store and retrieve cached files.
func New(StorageMechanism, Options) http.Handler {
  // ...
}
```
(I think `StorageMechanism` will look mostly like `LRU` @nicolov implemented recently).

WDYT?

Yannic:
I‘m not completely sure that this would be sufficient. `httpCache` seems to have a very tight coupling to the stats, which might cause issues in alternative implementations of `Cache` (I only have an immediate use case for alternative stogage).
I guess it‘s the best if I investigate some possible solutions and then report the results here.

buchgr:
I have pushed a change that goes into this direction https://github.com/buchgr/bazel-remote/commit/700d941c1a083f87e2b11d4437bf28f931e454f7

sorry for the large commit :(

PR #46, opened 2018-06-28
Split off and inject the blob store implementation

nicolov:
Right now, we currently only have a single cache backend implementation, using files on a local filesystem. However, it would be nicer to have alternative backends, such as S3.

Most of the code was already structured with this concern in mind, and this PR completes the job by renaming the `Cache` interface to `BlobStore` (which better reflects its actual role), and injecting implementations into `HTTPCache` (rather than constructing them there). This will make it easy to plug in multiple backends.

I've also tweaked the `BlobStore` tests so that they can run against multiple implementations without code duplication.

I believe this also addresses the concerns in https://github.com/buchgr/bazel-remote/issues/39


buchgr:
Hi @nicolov! I ll try to take a look tomorrow! Sorry for the delay!

buchgr:
FYI I have also started working on the same thing I guess :-) https://github.com/buchgr/bazel-remote/tree/cache-delegate

PR #47, opened 2018-07-13
Set default configuration in Dockerfile using environment variables

AlexandreCarlton:
This will allow users to easily change the port and directory without having to override the entrypoint.

I also updated the README to document the usage of these environment variables.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,4 +20,6 @@ FROM alpine:latest
 WORKDIR /root
 EXPOSE 80
 COPY --from=0 /go/src/github.com/buchgr/bazel-remote/bazel-remote .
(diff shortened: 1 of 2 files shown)


nicolov:
LGTM, or we could change `ENTRYPOINT` to `CMD` IIRC.

AlexandreCarlton:
I considered this, but if we had:

```Dockerfile
CMD ["--port=80", "--dir=/data"]
```
then a user already providing arguments to the `docker run` invocation (for example `--max-size 1`) would override these options instead of appending to them. From [the docs](https://docs.docker.com/engine/reference/builder/#cmd):

> If the user specifies arguments to `docker run` then they will override the default specified in `CMD`.

 Specifying these values as environment variables seemed better to avoid breaking changes.

PR #49, opened 2018-08-06
Use a simple sentinel style ErrTooBig error

mostynb:
This simplifies code that needs to recognise this particular error.


nicolov:
Many thanks.

PR #52, opened 2018-08-08
Document the REST API

### LINKED ISSUE #51: document the REST API
I think it would be helpful if the README.md described the REST API, and the difference between the action cache and the CAS cache.

After poking around a bit I found some info in #5, if you like I can have a go at documenting this?

mostynb:
Resolves #51.


buchgr:
thank you @mostynb

PR #57, opened 2018-10-31
Add s3 as a cache backend 

bshashank:
This change adds a new cache back-end to the project.

In order to add the new back-end, I have added a dependency to the minio go-sdk. I also had to update the versions of rules_go and gazelle that we point to. Let me know if you'd like to see that as a separate pull-request, and I'll be happy to split it.

I also have a simple unittest for the configuration check. I couldn't think of an easy way to unittest the s3 functionality (since it's just calling s3 APIs), see some functional tests I ran locally below.

This is my first PR against this project, let me know if you need me to change something, I'll be happy to do it.

----
Functional testing done:

Started minio via docker:

```
$ docker run -p 9000:9000 --name minio1                           \
  -v /home/sbharadwaj/local/bazel-data:/data                      \
  -v /home/sbharadwaj/.minio:/root/.minio                         \
   minio/minio server /data
```

Then run the bazel-remote project like so:

```
❯ bazel run :bazel-remote -- --config_file test.yaml 
```

where the `test.yaml` file is:
```
host: localhost
dir: /home/sbharadwaj/local/data
port: 8080
max_size: 100
s3_proxy:
  endpoint: '127.0.0.1:9000'
  bucket: test-bucket
  location: test-location
  access_key_id: <KEY_ID>
  secret_access_key: <SEC_KEY>
```

To test, I try to upload a simple file:

```
    $ sha256sum README.md | \
      cut -f1 -d' '       | \
      xargs -I% curl --upload-file README.md http://localhost:8080/cas/%
```

Then download using:
```
    $ sha256sum README.md | \
      cut -f1 -d' '       | \
      xargs -I% curl http://localhost:8080/cas/%
```
Resolves: buchgr#56

### DIFF
--- a/cache/s3/s3.go
+++ b/cache/s3/s3.go
@@ -61,8 +61,22 @@ func (c *s3Cache) Put(key string, size int64, expectedSha256 string, r io.Reader
 // Get writes the content of the cache item stored under `key` to `w`. If the item is
 // not found, it returns ok = false.
 func (c *s3Cache) Get(key string, actionCache bool) (data io.ReadCloser, sizeBytes int64, err error) {
+	objInfo, err := c.mclient.StatObject(
+		c.bucket, // bucketName
+		key,      // objectName
+		minio.StatObjectOptions{}, // opts
+	)
+	if err != nil {
+		return nil, 0, err
+	}
 
-	return nil, 0, nil
+	object, err := c.mclient.GetObject(
+		c.bucket, // bucketName
+		key,      // objectName
+		minio.GetObjectOptions{}, // opts
+	)
+
+	return object, objInfo.Size, err
 }
 
 // Contains returns true if the `key` exists.
--- a/cache/s3/s3.go
+++ b/cache/s3/s3.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/buchgr/bazel-remote/cache"
+	"github.com/buchgr/bazel-remote/config"
 	"github.com/minio/minio-go"
 )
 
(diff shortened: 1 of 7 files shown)


cache/s3/s3.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type s3Cache struct {
+	mclient  *minio.Client
+	location string
+	bucket   string
+}
+
+// New erturns a new instance of the S3-API based cached
+func New(endpoint string, bucket string, location string,
```
prestonvanloon: This information looks like S3CloudStorageConfig, why not pass in that struct rather than having so many parameters?

bshashank: Yeah makes sense. Done. Also fixed code to pass tests.

cache/s3/s3.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Initialize minio client object.
+	minioClient, err := minio.New(
+		s3Config.Endpoint, s3Config.AccessKeyID, s3Config.SecretAccessKey,
+		useSSL)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	log.Printf("%#v\n", minioClient) // minioClient is now setup
```
prestonvanloon: I think you should pass in an errorLogger like other cache backends do.

cache/s3/s3.go
```diff
@@ -0,0 +1,111 @@
...
+		c.bucket, // bucketName
+		key,      // objectName
+		r,        // reader
+		size,     // objectSize
+		minio.PutObjectOptions{
+			ContentType: "application/octet-stream",
+		}, // opts
+	)
+	if err != nil {
+		log.Fatalln(err)
```
prestonvanloon: Is fatal the correct level? This will kill the process. 

Also, I think you should use cache.Logger and pass in an errorLogger from main.go as other backends do.

buchgr: +1 to both suggestions.

buchgr:
Thanks for the PR! Please apologize for the delay in review as I am just coming back from vacation. Did you also test this with a real Bazel build?

bshashank:
Thanks for the reviews.

Will address the comments and update.

> Did you also test this with a real Bazel build?

Good point, haven't done a full blown test yet, will do and update.

cache/s3/s3.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/minio/minio-go"
+)
+
+type s3Cache struct {
+	mclient  *minio.Client
+	location string
+	bucket   string
+}
+
+// New erturns a new instance of the S3-API based cached
```
jdanekrh:
```suggestion
// New returns a new instance of the S3-API based cached
```

PR #68, opened 2019-01-14
disk.go: Don't sync after writing a file.

philwo:
The sync protects against data loss in case the power goes out suddenly, but I don't think that level of protection is needed for a cache server - in the worst case, if it loses a file, the clients can regenerate it anyway.

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -220,10 +220,6 @@ func (c *diskCache) Put(kind cache.EntryKind, hash string, size int64, r io.Read
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


nicolov:
If we remove the sync, are we still safe from incomplete files being stored on disk, and served back at a later time?

buchgr:
I have a change in flight that parallelies loading and verifies the integrity of the values stored. Let's merge your change after that one?

mostynb:
@philwo: is this still an issue for you? Maybe it could be a config option for people willing to take this risk?

Issue #61, opened 2018-12-07
Allow server to daemonize


ob:
One of the uses of the remote cache is as a proxy on local machines (Bazel waits for all the artifacts to upload before completing the build so a slow network connection can impact build times). 

I've been playing with the idea of having my Bazel wrapper automatically spawn a `bazel-remote` subprocess on a local port and rewire the Bazel command to prefer the local cache server than the remote one. 

In order for this to work, there are two features that I think would be needed from `bazel-remote`:

1. Detect when a `bazel-remote` process is already serving objects from a directory. 
2. Exit after an idle timeout.

I think for 1. the easiest would be to write the PID and port number in a file inside the cache directory. On startup, `bazel-remote` could check if that PID is still running, if not, it would start on the same port.

I could manage this from the Bazel wrapper, but I think it would be better to add it to `bazel-remote` so I'm opening an issue to start the discussion.

buchgr:
Hi Oscar,

that sounds like a great idea! I am happy to hear that you want to use it like this. In fact, there are early plans on adding functionality for Bazel to be able to launch a remote caching process locally. That is, so that 1) we can support the use you mentioned and 2) we'd be able to move authentication libraries from the Bazel binary into a remote cache binary. This binary could be lazily fetched from the internet and support proxying cache/execution requests to a remote system. This should make maintenance of such libraries easier and reduce Bazel's binary size.

Sounds good?

ob:
@buchgr sounds perfect! In my case, we have a go-written bazel-wrapper that already handles versioning, fetching, and caching of the Bazel main binary. Our plan is to add Bazel-remote to the wrapper and have it be spawned automatically, on demand. 

I'm wondering if bazel is going the route of using Bazel-wrapper like I intend to, maybe it'd be better to add this functionality straight to bazel?

We have an internal design document on how we are planning on doing the integration, I'll see if I can share the details if that is interesting.

buchgr:
> I'm wondering if bazel is going the route of using Bazel-wrapper like I intend to, maybe it'd be better to add this functionality straight to bazel?

I think so but I'd guess that's 9-12 months out at least, because other things are higher priority. However, I'd be happy to work with you to make the necessary changes to bazel-remote so that they integrate well with your Bazel-wrapper. I think this could also give us valuable feedback when ultimately adding this functionality to Bazel.

buchgr:
[...]

PR #71, opened 2019-01-24
Add max_idle_time flag to exit bazel-remote when idling for certain time

bayareabear:
This is for https://github.com/buchgr/bazel-remote/issues/61.
The max_idle_time flag is useful when using bazel-remote as a local proxy to PUT cache objects to remote cache server and exit when idle for a time specified by max_idle_time flag. This allows for local build finish early and leave bazel-remote to finish the PUT request queue, instead of waiting for the PUT requests to finish.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -30,6 +30,7 @@ type Config struct {
 	TLSKeyFile         string                    `yaml:"tls_key_file"`
 	GoogleCloudStorage *GoogleCloudStorageConfig `yaml:"gcs_proxy"`
 	HTTPBackend        *HTTPBackendConfig        `yaml:"http_proxy"`
+	IdleTimeout        int64                     `yaml:"idle_timeout"`
 }
 
 // New ...
@@ -40,6 +41,7 @@ func New(dir string, maxSize int, host string, port int, htpasswdFile string,
 		Port:               port,
 		Dir:                dir,
 		MaxSize:            maxSize,
+		IdleTimeout:        0,
 		HtpasswdFile:       htpasswdFile,
 		TLSCertFile:        tlsCertFile,
 		TLSKeyFile:         tlsKeyFile,
--- a/main.go
+++ b/main.go
@@ -145,14 +145,18 @@ func main() {
 		httpServer := &http.Server{Addr: c.Host+":"+strconv.Itoa(c.Port), Handler: nil}
 
 		if c.IdleTimeout > 0 {
+			cache.CacheMutex.Lock()
 			cache.LastRequestTime = time.Now()
+			cache.CacheMutex.Unlock()
 			ticker := time.NewTicker(time.Second)
 			go func() {
 				for {
 					select {
 					case <-ticker.C:
 						currentTime := time.Now()
+						cache.CacheMutex.Lock()
 						lastRequestTime := cache.LastRequestTime
+						cache.CacheMutex.Unlock()
 						if int64(currentTime.Sub(lastRequestTime).Seconds()) > c.IdleTimeout {
 							accessLogger.Printf("Shutting down server after idling for more than %d seconds", c.IdleTimeout)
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -30,6 +30,7 @@ type Config struct {
 	TLSKeyFile         string                    `yaml:"tls_key_file"`
 	GoogleCloudStorage *GoogleCloudStorageConfig `yaml:"gcs_proxy"`
 	HTTPBackend        *HTTPBackendConfig        `yaml:"http_proxy"`
+	MaxIdleTime        int64                     `yaml:"max_idle_time"`
```
buchgr: how about `idle_timeout`?

main.go
```diff
@@ -81,6 +81,12 @@ func main() {
 			Usage:  "Path to a pem encoded key file.",
 			EnvVar: "BAZEL_REMOTE_TLS_KEY_FILE",
 		},
+		cli.Int64Flag{
+			Name:   "max_idle_time",
+			Value:  -1,
```
buchgr: I think we don't need to use a negative value, as `0` would do just fine no?

server/http.go
```diff
@@ -84,6 +92,18 @@ func parseRequestURL(url string) (cache.EntryKind, string, error) {
...
 	defer r.Body.Close()
 
+	// Exit bazel-remote when there is no request for MaxIdleTime seconds
+	if MaxIdleTime > 0 {
+		if idleTimer != nil {
+			idleTimer.Stop()
+		}
+		idleTimer := time.AfterFunc(time.Duration(MaxIdleTime)*time.Second, func() {
+			h.accessLogger.Printf("Exit bazel-remote server after no request for %d seconds", MaxIdleTime)
+			os.Exit(0)
```
buchgr: One issue with calling `os.Exit()` is that it doesn't properly terminate open network connections (bazel-remote uses http keep-alive) and another problem with `os.Exit()` is that it's difficult to test (I think).

An alternative might be to use the https://godoc.org/net/http#Server.Shutdown function (see the example in godoc). The timer could then be implemented in `main.go` to periodically (once a second) check whether the last request has been served longer than `idle_timeout` seconds ago, if so it calls server. In order to do so simply add a lastRequestTime() function to the HTTPCache interface.

What do you think?

buchgr:
FYI @ob

ob:
Nice! One small issue I see is that the timer is not being reset when uploading to the remote (in the case `bazel-remote` is acting as a proxy). Other that that 👍

config/config.go
bayareabear: updated, thanks!

main.go
bayareabear: yeap

server/http.go
bayareabear: updated, thanks!

bayareabear:
Good point @ob , every request is going through "disk", PUT, GET, CONTAINS api, so I just put the reset timer there. Let me know if it makes sense to you.
Also, different from my first implementation, this latest idle timeout count time from the start of bazel-remote; while the previous implementation count time after it served some request.

main.go
```diff
@@ -134,12 +142,31 @@ func main() {
...
+
+		if c.IdleTimeout > 0 {
+			cache.LastRequestTime = time.Now()
+			ticker := time.NewTicker(time.Second)
+			go func() {
+				for {
+					select {
+					case <-ticker.C:
+						currentTime := time.Now()
+						lastRequestTime := cache.LastRequestTime
```
ob: I think you need either a Mutex or a Channel here since this var is being accessed from multiple threads?

bayareabear: absolutely!

buchgr:
Thanks @bayareabear and @ob! Merged as 57a18163783d3d0cb199ad93bcc788c864ec4140

PR #73, opened 2019-01-28
Create a pid file under bazel-remote cache directory

bayareabear:
This is for managing local bazel-remote processes. With pid files under the cache folder, bazel-remote can check previous invocations of bazel-remote and take corresponding action according to either flag or config file specification. 
For CI box, when there is a new build job, bazel-remote can check if there is already a previous bazel-remote process running, and if yes, it can either select to kill all previous invocations to use a new version of bazel-remote; or select to just select to continue to use previous invocations.
For dev box, when user run bazel build, it can spawn bazel-remote process with the same options as CI box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ GLOBAL OPTIONS:
    --tls_cert_file value  Path to a pem encoded certificate file. [$BAZEL_REMOTE_TLS_CERT_FILE]
    --tls_key_file value   Path to a pem encoded key file. [$BAZEL_REMOTE_TLS_KEY_FILE]
    --idle_timeout value   The maximum period of having received no request after which the server will shut itself down. Disabled by default. (default: 0s) [$BAZEL_REMOTE_IDLE_TIMEOUT]
+   --kill_old_pid value   This will kill the existing running bazel-remote process before starting a new bazel-remote process. This is when user want to upgrade with a new version
    --help, -h             show help
 ```
 
--- a/main.go
+++ b/main.go
@@ -3,12 +3,16 @@ package main
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	auth "github.com/abbot/go-http-auth"
@@ -90,6 +94,12 @@ func main() {
 			Usage:  "The maximum period of having received no request after which the server will shut itself down. Disabled by default.",
 			EnvVar: "BAZEL_REMOTE_IDLE_TIMEOUT",
 		},
+		cli.BoolFlag{
+			Name:   "kill_old_pid",
+			Hidden: false,
+			Usage:  "This will kill the existing running bazel-remote process before starting a new bazel-remote process. This is when user want to upgrade with a new version",
+			EnvVar: "BAZEL_REMOTE_KILL_OLD",
+		},
 	}
 
 	app.Action = func(ctx *cli.Context) error {
@@ -106,7 +116,8 @@ func main() {
 				ctx.String("htpasswd_file"),
 				ctx.String("tls_cert_file"),
 				ctx.String("tls_key_file"),
-				ctx.Duration("idle_timeout"))
+				ctx.Duration("idle_timeout"),
+				ctx.Bool("kill_old_pid"))
 		}
 
 		if err != nil {
@@ -117,6 +128,8 @@ func main() {
 
 		accessLogger := log.New(os.Stdout, "", log.Ldate|log.Ltime|log.LUTC)
 		errorLogger := log.New(os.Stderr, "", log.Ldate|log.Ltime|log.LUTC)
+
+		writePidFile(c, accessLogger)
 
 		diskCache := disk.New(c.Dir, int64(c.MaxSize)*1024*1024*1024)
 
@@ -201,4 +214,38 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
 	secrets := auth.HtpasswdFileProvider(htpasswdFile)
 	authenticator := auth.NewBasicAuthenticator(host, secrets)
 	return auth.JustCheck(authenticator, handler)
+}
+
+func writePidFile(c *config.Config, accessLogger cache.Logger) error {
+	// create a "pid" directory under cache directory and clean up inactive pid file
+	pidPath := filepath.Join(c.Dir, "pid")
+	err := os.MkdirAll(pidPath, os.FileMode(0744))
+	if err != nil {
+		log.Fatal(err)
+		return err
+	}
+	err = filepath.Walk(pidPath, func(name string, fileInfo os.FileInfo, err error) error {
+		if !fileInfo.IsDir() {
+			pid, _ := strconv.Atoi(fileInfo.Name())
+			bazelRemoteProcess, err := os.FindProcess(pid)
+			if err != nil {
+				accessLogger.Printf("Error to find pid: %d", bazelRemoteProcess)
+			} else {
+				err = bazelRemoteProcess.Signal(syscall.Signal(0))
+				if err != nil && strings.Contains(err.Error(), "process already finished") {
+					accessLogger.Printf("Removing pid file: %s", fileInfo.Name())
+					err = os.Remove(filepath.Join(c.Dir, "pid", fileInfo.Name()))
+				} else if c.KillOldPid {
+					accessLogger.Printf("Killing existing bazel remote process: %d", pid)
+					err = bazelRemoteProcess.Signal(syscall.Signal(syscall.SIGKILL))
+				}
+			}
+		}
+		return err
+	})
+	// create a file with pid number as the name and write critical server info into the file
+	pidFile := filepath.Join(c.Dir, "pid", strconv.Itoa(os.Getpid()))
+	port := []byte("port: " + strconv.Itoa(c.Port) + "\n")
+	err = ioutil.WriteFile(pidFile, port, 0744)
+	return err
 }
--- a/config/config.go
+++ b/config/config.go
@@ -32,11 +32,12 @@ type Config struct {
 	GoogleCloudStorage *GoogleCloudStorageConfig `yaml:"gcs_proxy"`
 	HTTPBackend        *HTTPBackendConfig        `yaml:"http_proxy"`
 	IdleTimeout        time.Duration             `yaml:"idle_timeout"`
+	KillOldPid         bool                      `yaml:"kill_old_pid"`
 }
 
 // New ...
 func New(dir string, maxSize int, host string, port int, htpasswdFile string,
-	tlsCertFile string, tlsKeyFile string, idleTimeout time.Duration) (*Config, error) {
+	tlsCertFile string, tlsKeyFile string, idleTimeout time.Duration, killOldPid bool) (*Config, error) {
 	c := Config{
 		Host:               host,
 		Port:               port,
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -51,6 +51,7 @@ GLOBAL OPTIONS:
    --tls_cert_file value  Path to a pem encoded certificate file. [$BAZEL_REMOTE_TLS_CERT_FILE]
    --tls_key_file value   Path to a pem encoded key file. [$BAZEL_REMOTE_TLS_KEY_FILE]
    --idle_timeout value   The maximum period of having received no request after which the server will shut itself down. Disabled by default. (default: 0s) [$BAZEL_REMOTE_IDLE_TIMEOUT]
+   --kill_old_pid value   This will kill the existing running bazel-remote process before starting a new bazel-remote process. This is when user want to upgrade with a new version
```
ob: I'm not sure this is a good name for this option. I would decouple killing old `bazel-remote` processes from writing down the PID/PORT in the cache directory. I think it'd be easier to look at the change if the two were separate as there is no need for an option for the former and handling the latter properly can be tricky. (See other comments).

main.go
```diff
@@ -118,6 +129,8 @@ func main() {
 		accessLogger := log.New(os.Stdout, "", log.Ldate|log.Ltime|log.LUTC)
 		errorLogger := log.New(os.Stderr, "", log.Ldate|log.Ltime|log.LUTC)
 
+		writePidFile(c, accessLogger)
```
ob: I find it surprising that a function called `writePidFile` would have the side-effect of killing other processes. Better to decouple that logic.

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
 	authenticator := auth.NewBasicAuthenticator(host, secrets)
 	return auth.JustCheck(authenticator, handler)
 }
+
+func writePidFile(c *config.Config, accessLogger cache.Logger) error {
+	// create a "pid" directory under cache directory and clean up inactive pid file
+	pidPath := filepath.Join(c.Dir, "pid")
```
ob: If there can be more than one PID, the director should be `pids` no?

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
...
+	err := os.MkdirAll(pidPath, os.FileMode(0744))
+	if err != nil {
+		log.Fatal(err)
+		return err
+	}
+	err = filepath.Walk(pidPath, func(name string, fileInfo os.FileInfo, err error) error {
```
ob: `filepath.Walk()` is probably more expensive than `ioutil.ReadDir()` or even `filepath.Glob()`. It also seems overkill to recursively traverse the directory.

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
...
+		if !fileInfo.IsDir() {
```
ob: It'd be better to check for `IsRegular()` since other things that are not directories are not necessarily files (e.g. symlinks, devices, named pipes, etc).

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
...
+			pid, _ := strconv.Atoi(fileInfo.Name())
+			bazelRemoteProcess, err := os.FindProcess(pid)
```
ob: If `strconv.AtoI()` fails `pid` can be null.

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
...
+			if err != nil {
+				accessLogger.Printf("Error to find pid: %d", bazelRemoteProcess)
+			} else {
+				err = bazelRemoteProcess.Signal(syscall.Signal(0))
+				if err != nil && strings.Contains(err.Error(), "process already finished") {
```
ob: Better to check `errno` than pattern match on the error string.

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
...
+					accessLogger.Printf("Removing pid file: %s", fileInfo.Name())
+					err = os.Remove(filepath.Join(c.Dir, "pid", fileInfo.Name()))
```
ob: Where is the `pid` file removed if the server exits by itself? For example because of an idle timeout? Or Ctrl-C'd? Or killed from a different terminal?

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
...
+				} else if c.KillOldPid {
+					accessLogger.Printf("Killing existing bazel remote process: %d", pid)
+					err = bazelRemoteProcess.Signal(syscall.Signal(syscall.SIGKILL))
```
ob: `SIGKILL` does not terminate the server gracefully. See [this article](https://medium.com/@kpbird/golang-gracefully-stop-application-23c2390bb212) for ways to do that.

main.go
njuliyu: Per the latest design doc, the pid file name is fixed to <bazel-remote version>.d, so no need to search for all pid file unless we want to kill the daemon of a different version.

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
...
+				}
```
njuliyu: If a daemon is already running and we don't want to kill it, this process should exit immediately. It seems the current code doesn't support this behavior. It just continue to check the next pid file instead.

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
 	authenticator := auth.NewBasicAuthenticator(host, secrets)
 	return auth.JustCheck(authenticator, handler)
 }
+
+func writePidFile(c *config.Config, accessLogger cache.Logger) error {
```
njuliyu: What if two bazel-remote daemon start at the same time, and find there is no existing pid file at the same time? It could cause two running daemons of the same version. I think we'd better have a file lock at the beginning of the function to prevent the race condition.

Also checkRunningProcess might be a function name.

main.go
```diff
@@ -202,3 +215,37 @@ func wrapAuthHandler(handler http.HandlerFunc, htpasswdFile string, host string)
...
+					err = bazelRemoteProcess.Signal(syscall.Signal(syscall.SIGKILL))
+				}
+			}
+		}
+		return err
+	})
+	// create a file with pid number as the name and write critical server info into the file
+	pidFile := filepath.Join(c.Dir, "pid", strconv.Itoa(os.Getpid()))
+	port := []byte("port: " + strconv.Itoa(c.Port) + "\n")
+	err = ioutil.WriteFile(pidFile, port, 0744)
```
njuliyu: if failed to write the pid file, the process should exit.

main.go
bayareabear: strconv.AtoI() always returns an int

main.go
bayareabear: That's a good point. I'll see how we can better deal with it.

main.go
bayareabear: The pid file is written at bazel-remote start up and is removed by another instance start up. The purpose of a pid file is just to let bazel-remote gain knowledge of other PIDs

main.go
bayareabear: That's good idea. But the Signal call only returns error string. Also os.FindProcess(pid) will always returns a pid, so it will always exist

main.go
bayareabear: Good point!

main.go
bayareabear: Not sure how ioutil.ReadDir() can be less expensive? filepath.Walk() offers convenient api to process the file at the same time. This is a directory with just a few files and this only runs once.

main.go
bayareabear: agreed

main.go
bayareabear: yeah, it could be multiple files. but I think 'pid' is better than 'pids' as it conforms the convention with 'ac', 'cas'

README.md
bayareabear: agreed

main.go
bayareabear: This will be in a seperate PR, and will be supported through a flag like:
--kill_old_pid

main.go
bayareabear: as long as we specify a port, this won't happen.

main.go
bayareabear: agreed

main.go
bayareabear: That is needed as the main purpose is to manage version upgrade. File name without .d looks simpler

main.go
ob: Sorry, I should've been more clear. If there is a file that can't be converted to a number, for example a file called "junk", then `strong.Atoi()` will return 0, which is a valid argument to `os.FindProcess()` as it'll just wrap it in a `*Process`. Below, in the call to `Signal`, it'll use PID 0 which means send the signal to all the processes with the same group id as the sender, effectively killing the sender itself.

main.go
ob: `errno` is set after any system call, regardless of what the return value is.

`os.FindProcess()` will always succeed because it's just doing a type-conversion from `int` (the PID) to `*Process`. It makes no guarantees about the existence of the PID.

main.go
ob: What if there isn't another instance and the only `bazel-remote` instance just exits due to an `idle_timeout`? Who cleans up the PID file then?

njuliyu: I agree with ob that the current process should delete its pid when gracefully terminated.  Cleaning up at the next start is a way to do garbage collection, but this is not the normal case.  The normal case should be: no processing running, no pid files.

main.go
njuliyu: Per the design doc, for the same version of bazel-remote daemon, there is only one pid file name  <version>.d.  Do you want to kill the daemon of another version? Please note we are talking about the process of ANOTHER version, not the same version.

main.go
njuliyu: Why that will be in another PR?  Checking the pid file, if one is running, exit yourself, that is the whole process.

--kill_pod_pid is about to kill other process, different thing.

main.go
njuliyu: I don't see how specifying a port can help in this case.

Firstly, two daemons can use different ports.

Even if you specify the same port for both, one daemon cannot start because port binding will fail. But that happens after writing the pid files. Regardless you using one pid file or multiple pid files, they have been messed up.

main.go
bayareabear: I don't think it is good idea to rely on the file to assume if there is a pid file the process is running. The pid file is just to tell bazel-remote a process might be running. It has to verify if it is actually running, so I don't see it is necessary to hold this: "no processing running, no pid files."

main.go
bayareabear: what race condition can it cause? If two bazel-remote start at the same time, they will have different pid, and write to different files. no?

main.go
bayareabear: There might be use cases where we may want to keep the original process running and spawn a new process with a different port

main.go
njuliyu: The whole purpose of this PR is to use pid file to prevent two bazel-remote process of the same version running. Now you allow two process to run, I am confused what you want to do.

main.go
njuliyu: We need to clarify the design requirement, and the give solid use cases.

Even if that is a supported use case, it is still the most common use case that that the daemon terminate itself if there is another process of the same version already running.

main.go
njuliyu: No one has to RELY on the file to assume if there is a pid file the process is running. But it is a good indication. It is a common practice of using pid files

tomosulo:
Lao Zhang V5!

Issue #69, opened 2019-01-17
Support read-only proxy


uri-canva:
Related to https://github.com/buchgr/bazel-remote/issues/59, I have a setup where developer machines have a local cache they write to, as discussed in https://github.com/bazelbuild/bazel/issues/5139, however to proxy to the shared cache that CI writes to one must also enable writing to it, as there is no configuration knob for that. I could disable writing to the cache on the client side, but then we wouldn't be using the cache running locally as a directory cache.

shmish111:
This also makes it safer to provide a publicly available cache, it provides the benefits of caching anything that builds on CI without the risk of malicious users polluting or PUT DOSing the cache.

uri-canva:
Duplicate of #59.

PR #76, opened 2019-01-30
Add configs for readonly and writeonly proxies.

uri-canva:
Implements #69.

---
This change is [Reviewable](https://reviewable.io/reviews/buchgr/bazel-remote/76)

### DIFF
--- a/cache/http/http.go
+++ b/cache/http/http.go
@@ -17,7 +17,7 @@ import (
 const numUploaders = 100
 const maxQueuedUploads = 1000000
 
-type Mode int
+type AccessMode int
 
 const (
 	Read = iota + 1
@@ -34,7 +34,7 @@ type remoteHTTPProxyCache struct {
 	remote       *http.Client
 	baseURL      *url.URL
 	local        cache.Cache
-	mode         Mode
+	accessMode   AccessMode
 	uploadQueue  chan<- (*uploadReq)
 	accessLogger cache.Logger
 	errorLogger  cache.Logger
@@ -68,7 +68,7 @@ func uploadFile(remote *http.Client, baseURL *url.URL, local cache.Cache, access
 }
 
 // New creates a cache that proxies requests to a HTTP remote cache.
-func New(baseURL *url.URL, local cache.Cache, remote *http.Client, mode Mode, accessLogger cache.Logger,
+func New(baseURL *url.URL, local cache.Cache, remote *http.Client, accessMode AccessMode, accessLogger cache.Logger,
 	errorLogger cache.Logger) cache.Cache {
 	uploadQueue := make(chan *uploadReq, maxQueuedUploads)
 	for uploader := 0; uploader < numUploaders; uploader++ {
@@ -83,7 +83,7 @@ func New(baseURL *url.URL, local cache.Cache, remote *http.Client, mode Mode, ac
 		remote:       remote,
 		baseURL:      baseURL,
 		local:        local,
-		mode:         mode,
+		accessMode:   accessMode,
 		uploadQueue:  uploadQueue,
 		accessLogger: accessLogger,
 		errorLogger:  errorLogger,
@@ -102,7 +102,7 @@ func (r *remoteHTTPProxyCache) Put(kind cache.EntryKind, hash string, size int64
 	}
 	r.local.Put(kind, hash, size, data)
 
-	if r.mode&Write == 0 {
+	if r.accessMode&Write == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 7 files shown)


uri-canva:
This is configured via separate `disable_reads` and `disable_writes` fields in the yaml so that the default values match the current behavior of having both enabled. I could have used an enum as well but I didn't want to complicate the yaml parsing code, it might be worth it?

In the code it's represented as a `Mode` enum because that way the invalid configuration of disabling both cannot be represented at the type level.

Happy to change the config to be a mode enum in yaml as well if that is preferable.

cache/http/http.go
```diff
@@ -26,6 +34,7 @@ type remoteHTTPProxyCache struct {
 	remote       *http.Client
 	baseURL      *url.URL
 	local        cache.Cache
+	mode         Mode
```
bayareabear: maybe 'accessMode'?

cache/http/http.go
```diff
@@ -92,6 +102,10 @@ func (r *remoteHTTPProxyCache) Put(kind cache.EntryKind, hash string, size int64
 	}
 	r.local.Put(kind, hash, size, data)
 
+	if r.mode&Write == 0 {
```
bayareabear: adding spaces: r.mode & Write?

config/config.go
```diff
@@ -14,10 +14,14 @@ type GoogleCloudStorageConfig struct {
 	Bucket                string `yaml:"bucket"`
 	UseDefaultCredentials bool   `yaml:"use_default_credentials"`
 	JSONCredentialsFile   string `yaml:"json_credentials_file"`
+	DisableReads          bool   `yaml:"disable_reads"`
```
bayareabear: just curious, what is the use case for disable_reads?

uri-canva: Write through from a cache that is less likely to get poisoned to one that is more likely to: for example having the cache shared by CI write through to the cache shared by developers.

The advantage over setting them up the other way around, with the developer shared cache reading through to the CI cache is that the developer cache doesn't need to wait for the transfers: for example if your CI cache is in a data center and your developer cache is in the office, the link between them might not be as fast as the link between developers and the developer cache. In that case waiting to download the artefact from the data center cache might be generally slower than building locally. But if you set up write through the data center cache can slowly send artefacts to the developer cache without blocking any builds.

uri-canva:
```
Developer machine <===> developer cache <---> CI cache
                    ^                     ^
            fast connection        slow connection
```

Developer machines can quickly read and write to developer cache, developer cache can slowly write to CI cache without blocking developers because uploads are asynchronous, but it cannot slowly read from CI cache without blocking developers because downloads are synchronous.

cache/http/http.go
uri-canva: `gofmt` formats this without spaces.

uri-canva:
ping @buchgr

ob:
@buchgr is [OOO until Feb 25](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/bazel-dev/u5UCzUN3pPk/Ik61XZb8BQAJ)

uri-canva:
ping @buchgr

uri-canva:
ping @buchgr

brentleyjones:
We could really use this feature.

@uri-canva would you reopen this if @buchgr would accept it?

How about it @buchgr 🙏?

uri-canva:
We're not using this patch anymore ourselves, as it didn't fit our purpose as well as I thought, you're welcome to base a new PR on it though.

mostynb:
Since there are a few interrelated issues/PRs on this topic, I want to make sure that I understand the use case...

Is this the scenario:
* There's a central CI system with its own cache. This is trustworthy because the CI setup is locked down.
* There is a developer-local cache, which is more easily poisoned.
* The link between the central and developer-local cache is slow enough that you don't want synchronous lookups between them.
* You want the central cache to asynchronously populate the developer-local cache, but not to read from it.
* (Not sure if this was requested, but do you potentially want multiple developer-local caches to work like this, with a single central cache?)

brentleyjones:
I’ll let @uri-canva answer for their situation, but for us it is much simpler, we only really need the `writes_disabled` setting:

- Central CI cache. Written to by CI, since it’s trusted to not poison it.
- Local developer cache on each machine. Used since `disk_cache` doesn’t clean up, and implementing naive cleanup in a non-AC aware way breaks Builds without the Bytes.
- Local developer cache points it’s proxy to CI cache with `writes_disabled` to get remote cache benefits without possibility of poisoning it.

uri-canva:
Disregard the note around `disable_reads` above, we have it set up differently now because it wasn't working that well anyway. We were using `disable_writes` exactly the same way as @brentleyjones: to use `bazel-remote` as a local disk cache, given the discussion in https://github.com/bazelbuild/bazel/issues/5139.

> Essentially, if `disk_cache` could set a size limit and clean up in an LRU way, we wouldn’t need this solution at all. This allows `bazel-remote` to fill that purpose instead.

We use it to fall back to reading from a remote cache too, which Bazel doesn't do.

Issue #78, opened 2019-02-05
Why is the cache directory 744?


ob:
Is there any reason the cache directory defaults to 0744 instead of the more common 0755 or even respecting `umask`?

I'm talking about [this code](https://github.com/buchgr/bazel-remote/blob/master/cache/disk/disk.go#L291).

mostynb:
I'm not sure why 0744 was chosen, but it has one possibly useful quality: it prevents other users from listing the directory, which can be a very slow operation if the cache is large.

ob:
True, but it also makes managing the cache harder on servers where you might not have root and bazel-remote is run under some other user.

PR #79, opened 2019-02-11
Change default mode for dirs from 0744 to 0777 and files from 0600 to 0666

ob:
This makes it easier to administer the cache directory when
bazel-remote is not running as the same user as the server admin (and
the server admin doesn't have root).

Note that ioutil.MkdirAll() respects umask so if you still want 0744,
you can just set your umask to 0033.

Fixes issue #78


ob:
I'm tempted to change these to `0777` since Go respects `umask`.

ob:
Superseded by #128

PR #83, opened 2019-02-27
Explicitly specify goarch and goos to build the binary for Linux

rabadin:
Since the container (distroless go container) on which the binary
runs is Linux, the binary needs to be built for Linux.
Without this, on MacOSX, the binary is built for darwin and thus
fails to run in the Linux container.


buchgr:
That's been merged ashttps://github.com/buchgr/bazel-remote/commit/70d244db641ad4c20664f9b18a7b4e3674b030b3

PR #87, opened 2019-04-02
Allow cross-platform Docker builds

nugget:
This PR explicitly sets `GOOS` and `GOARCH` to `linux/amd64` when creating the docker image.  This enables developers on a non-Linux platform to produce functional docker images instead of inadvertently embedding a binary built for their local architecture into the image.


buchgr:
thanks!

PR #89, opened 2019-05-09
Don't test on Ubuntu 14.04

philwo:
Ubuntu 14.04 is about to be end-of-life and Bazel CI will stop supporting it shortly afterwards.

Context: https://groups.google.com/d/msg/bazel-dev/_D6XzfNkQQE/8TNKiNmsCAAJ


gjasny:
Build failure is fixed by #90

PR #90, opened 2019-05-16
Fix for Bazel 0.25.x

gjasny:
Make project compatible with Bazel 0.25.1 by raising Go, Gazelle, and Docker rules.


buchgr:
thanks!

Issue #92, opened 2019-06-14
In-memory storage of metadata (AC)


excitoon:
I feel like storing metadata in memory would be super efficient. Is it worth implementing? @buchgr

PR #93, opened 2019-06-15
In-memory AC.

excitoon:
This fix adds in-memory cache for AC records.

Solves #92

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -1,6 +1,7 @@
 package disk
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


excitoon:
What I would also like to make in this PR:
- [ ] move from AC cache to entries-smaller-than-N-bytes cache;
- [ ] add size limit for these entries.

mostynb:
I wonder if the kernel's page cache already performs an approximation of this optimization transparently? Have you benchmarked this compared to the filesystem backend?

PR #98, opened 2019-08-13
fix the cache interface documentation

mostynb:
The previous comments were out of date, and refer to variables that
have since been renamed, and to older return types.


mostynb:
Applied a correction to the Get comment (how to correctly recognise when the item was not found).

nicolov:
thanks!

PR #101, opened 2019-08-28
Support LDAP authentication

### LINKED ISSUE #100: Support LDAP authentication
This is desirable for enterprise use.

HackAttack:
This PR adds LDAP as a credential backend, with a caching layer in front of it so as not to try to authenticate every single request.

Resolves #100

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -20,6 +20,16 @@ type HTTPBackendConfig struct {
 	BaseURL string `yaml:"url"`
 }
 
+type LDAPConfig struct {
+	BaseURL           string        `yaml:"url"`
+	BaseDN            string        `yaml:"base_dn"`
+	BindUser          string        `yaml:"bind_user"`
+	BindPassword      string        `yaml:"bind_password"`
+	UsernameAttribute string        `yaml:"username_attribute"`
+	Groups            []string      `yaml:"groups,flow"`
+	CacheTime         time.Duration `yaml:"cache_time"`
+}
+
 // Config provides the configuration
 type Config struct {
 	Host               string                    `yaml:"host"`
@@ -31,6 +41,7 @@ type Config struct {
 	TLSKeyFile         string                    `yaml:"tls_key_file"`
 	GoogleCloudStorage *GoogleCloudStorageConfig `yaml:"gcs_proxy"`
 	HTTPBackend        *HTTPBackendConfig        `yaml:"http_proxy"`
+	LDAP               *LDAPConfig               `yaml:"ldap"`
 	IdleTimeout        time.Duration             `yaml:"idle_timeout"`
 }
 
@@ -47,6 +58,7 @@ func New(dir string, maxSize int, host string, port int, htpasswdFile string,
 		TLSKeyFile:         tlsKeyFile,
 		GoogleCloudStorage: nil,
 		HTTPBackend:        nil,
+		LDAP:               nil,
 		IdleTimeout:        idleTimeout,
 	}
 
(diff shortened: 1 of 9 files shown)


ldap/ldap.go
```diff
@@ -0,0 +1,156 @@
...
+		} else {
+			ar := &auth.AuthenticatedRequest{Request: *r, Username: username}
+			wrapped(w, ar)
+		}
+	}
+}
+
+type key int
+
+var infoKey key
```
buchgr: Why does this need to be package global? It's never written?

ldap/ldap.go
```diff
@@ -0,0 +1,156 @@
...
+		if username := c.CheckAuth(r); username == "" {
+			c.RequireAuth(w, r)
+		} else {
+			ar := &auth.AuthenticatedRequest{Request: *r, Username: username}
+			wrapped(w, ar)
+		}
+	}
+}
+
+type key int
```
buchgr: this looks unused?

buchgr:
Thanks @HackAttack ! Apologies for the delay on the review. Overall looks great :-)

ldap/ldap.go
```diff
@@ -0,0 +1,156 @@
...
+	if ce.authed != nil {
+		return *ce.authed
+	}
+	// Not initialized; actually do the query and record the result
+	authed := c.query(user, password)
+	ce.authed = &authed
+	timeout := c.config.CacheTime
+	// Don't cache a negative result for a long time; likely wrong password
+	if !authed {
+		timeout = 5 * time.Second
```
buchgr: Out of Curiosity: Why cache a negative result at all?

ldap/ldap.go
```diff
@@ -0,0 +1,156 @@
...
+// Either query LDAP for a result or retrieve it from the cache
+func (c *Cache) checkLdap(user, password string) bool {
+	k := [2]string{user, password}
+	v, _ := c.m.LoadOrStore(k, &cacheEntry{})
+	ce := v.(*cacheEntry)
+	ce.Lock()
+	defer ce.Unlock()
+	if ce.authed != nil {
+		return *ce.authed
+	}
```
buchgr:
```
ce.Lock()
defer ce.Unlock()
if ce.authed != nil {
  return *ce.authed
}
```

Could this be changed to the below?

```
authed := ce.authed
if authed != nil {
  return authed
}
ce.Lock()
defer ce.Unlock()
```

That way you don't have to acquire a lock in the case where it was cached?

config/config.go
```diff
@@ -120,5 +132,30 @@ func validateConfig(c *Config) error {
...
+		if c.LDAP.BindUser == "" {
+			return errors.New("The 'bind_user' field is required for 'ldap'")
+		}
+		if c.LDAP.BindPassword == "" {
+			return errors.New("The 'bind_password' field is required for 'ldap'")
+		}
+		if c.LDAP.UsernameAttribute == "" {
+			c.LDAP.UsernameAttribute = "uid"
+		}
+		if c.LDAP.CacheTime == 0 {
```
mostynb: Maybe check <= 0 here?

ldap/ldap.go
mostynb: If you build with invalid auth and large -j value, I guess you don't want to make several hundred LDAP requests before the client notices and stops/cancels the remaining jobs.

ldap/ldap.go
```diff
@@ -0,0 +1,156 @@
...
+	defer conn.Close()
+	// Test the configured bind credentials
+	if err = conn.Bind(config.BindUser, config.BindPassword); err != nil {
+		return nil, err
+	}
+	return &Cache{
+		config: config,
+		BasicAuth: &auth.BasicAuth{
+			Realm: "Bazel remote cache",
```
mostynb: Should this be a configuration option?

ldap/ldap.go
mostynb: I also think this should have a nice fast-path, but this suggestion might need some tweaking to avoid multiple identical queries in quick succession: after obtaining the lock, ce.authed should be re-checked.

buchgr:
If anyone is watching this PR, I'd be really interested if this works for you. I am happy to rebase and merge it if this brings value to people.

JonasScharpf:
are there any show stoppers from preventing a merge besides the review comments?

HackAttack:
Wow, I completely and utterly forgot about this! This was one or two companies ago so I don’t know that I have a good way to test this now. I assume I verified it to work at the time, but it’s been a while…

JonasScharpf:
No worries. May you could at least rebase your branch to the latest master? In best case some of the review comments could be fixed right away, but this won't be necessary for testing I assume.

mostynb:
If anyone wants to rebase and/or resubmit a newer version of this, please also add some automated tests.

PR #107, opened 2019-10-18
Check PUT request content size

allada:
Ensure that the PUT requests size is the same as the actual size of the data received.

This was discovered in some cases bazel is in the middle of uploading the file, the user terminates bazel, the cache server blindly accepts the file and bazel now redistributes incomplete data.


buchgr:
Good catch @allada !

Issue #106, opened 2019-10-18
add some validation for (http) action cache lookups, and a new REST API endpoint with no validation


mostynb:
When looking up Action messages stored in the ActionCache, the results are only usable if all the referenced output blobs are also available in the CAS. When the client performs a GET or HEAD request, bazel-remote should confirm that these blobs exist (thereby moving them to the "safe" end of the LRU queue), and otherwise return 404 (even if the Action exists).

Since adding extra validation might break existing users (eg if using bazel-remote's ActionCache as a generic blob store), perhaps we should add a new raw/ REST API endpoint that behaves like ac/ does currently (ie no validation). I'm not sure if there's a reasonable automatic migration path, but sysadmins can at least rename preexisting ac/ directories to raw/ before upgrading bazel-remote.

Another option would be to add a command line flag to turn on extra validation for HTTP ac/.

buchgr:
I'd be in favor of adding a flag to turn it off and have it on by default. My reasoning is that most users of bazel-remote will be using it with Bazel and it will be a clear benefit to them. Does that make sense?

mostynb:
How about:
* always validate in gRPC ActionCache calls (because this API can't be used for generic data)
* validate HTTP ac actions by default
* `--disable-http-ac-validation` disable all validation of HTTP ac actions (because the REST API can be used for generic data)

Maybe consider at a later point (but hopefully not):
* `--disable-http-ac-pool-validation=name` disable validation of HTTP ac actions in the `name` pool (if some users want some but not all HTTP ac actions to be validated

buchgr:
sounds like a plan :rocket:

PR #108, opened 2019-10-23
disk cache tests: return errors from helpers, fail at the call site

mostynb:
If we call t.Fatal inside test helpers, then the log message says that
the test failed inside that function. But what we really want to know
is where the helper was called from. So return errors and call t.Fatal
from the call site.

This increases the amount of boilerplate around these helpers, but
gives better error messages.

The tempDir function has no test-specific failures, and can be excluded.

Cleanup before implementing #106.


buchgr:
thanks @mostynb !

PR #109, opened 2019-10-25
rdr is a better variable name than data for cache.Get / cache.Put

mostynb:
I tended to assume that these "data" variables were byte slices.
rdr seems like a more obvious name.

Cleanup before implementing #106.


buchgr:
thanks @mostynb ! Excuse my Go :-)

mostynb:
No problem, I have a couple of other small improvements like this on the way.

PR #110, opened 2019-10-26
update README.txt with gRPC and s3 info

mostynb:
These changes sync README.md with the latest --help flags, and mentions the experimental gRPC support.

PR #111, opened 2019-10-28
validate ActionCache items before returning them for http also

mostynb:
A common bug is the scenario where an ActionResult exists in the
ActionCache and is returned to the client, but not all the blobs
that it depends on are available in the CAS when the client tries
to get them. The client typically errors out at this point, and
might also continue failing when re-run, until the ActionResult
is removed from the cache.

To avoid this problem, we should verify that all the required
blobs are available before returning a given ActionCache. Doing
so should move these items to the "safe" end of the LRU queue,
and increase the odds that the CAS blobs will still exist in the
cache when the client requests them.

We previously had this check in the gRPC interface, but not for
the REST interface. This change refactors the check so it's
available to the REST interface. Since some clients might be
relying on bazel-remote to store arbitrary non-ActionResult blobs
in the ActionCache via the REST interface, we provide a command
line flag to disable this validation:
--disable_http_ac_validation

When --disable_http_ac_validation is specified, ActionCache blobs
uploaded and downloaded via REST are stored in a new "raw"
keyspace to avoid clashes with validated gRPC ActionResult blobs
(which are always validated and always stored in the "ac" keyspace).

If you want to use --disable_http_ac_validation and not lose access
to existing cache items, stop bazel-remote, rename the "ac" dir to
"raw", then restart bazel-remote with --disable_http_ac_validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ GLOBAL OPTIONS:
    --s3.access_key_id value      The S3/minio access key to use when using S3 cache backend. [$BAZEL_REMOTE_S3_ACCESS_KEY_ID]
    --s3.secret_access_key value  The S3/minio secret access key to use when using S3 cache backend. [$BAZEL_REMOTE_S3_SECRET_ACCESS_KEY]
    --s3.disable_ssl              Whether to disable TLS/SSL when using the S3 cache backend.  Default is false (enable TLS/SSL). [$BAZEL_REMOTE_S3_DISABLE_SSL]
+   --disable_http_ac_validation  Whether to disable ActionResult validation for HTTP requests.  Default is false (enable validation). [$BAZEL_REMOTE_DISABLE_HTTP_AC_VALIDATION]
    --help, -h                    show help
 ```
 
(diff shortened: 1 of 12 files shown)


mostynb:
TODO before this is ready:
* Decide if we should validate CAS blob availability on PUT.
* Add the new flag to README.md (https://github.com/buchgr/bazel-remote/pull/110 should land first).
* Decide if we should expose a "raw" HTTP endpoint, which would allow both validated and unvalidated blob uploads via HTTP.

buchgr:
> Decide if we should validate CAS blob availability on PUT.

I think having this check on GET is sufficient for now. I think ideally we'd have a GC with reference counting. That is, we'd only evict an output if no action references it or similarly if we evict and output also evict all actions that reference it.  That way we would only need to do the check on upload and not on download and thus also not have any performance cost on GET.

mostynb:
Updated the readme, removed a TODO note in the PUT section. I like the idea of having a separate "raw/" REST endpoint, but this version is also OK for my use cases.

PR #120, opened 2019-11-05
log git commit earlier in the startup process

mostynb:
I also had to dig around a bit to figure out how to make bazel inject this value, I found the answer over in https://github.com/bazelbuild/rules_go/issues/2224

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -34,7 +34,7 @@ go_binary(
     pure = "on",
     static = "on",
     visibility = ["//visibility:public"],
-    x_defs = {"github.com/buchgr/bazel-remote/server.GitCommit": "{STABLE_GIT_COMMIT}"},
+    x_defs = {"main.gitCommit": "{STABLE_GIT_COMMIT}"},
 )
 
 go_image(
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,3 @@
+# Include git version info
+build --stamp
+build --workspace_status_command 'echo STABLE_GIT_COMMIT $(git rev-parse HEAD)'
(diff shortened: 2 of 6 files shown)


BUILD.bazel
```diff
@@ -34,7 +34,7 @@ go_binary(
     pure = "on",
     static = "on",
     visibility = ["//visibility:public"],
-    x_defs = {"github.com/buchgr/bazel-remote/server.GitCommit": "{STABLE_GIT_COMMIT}"},
```
buchgr: why rename this?

README.md
```diff
@@ -99,7 +99,7 @@ $ bazel run :bazel-remote-image
 ## Build a standalone Linux binary
 
 ```bash
-$ bazel build :bazel-remote
+$ bazel build --stamp --workspace_status_command=$(pwd)/git-commit.sh :bazel-remote
```
buchgr: Hmm. Consider putting this in an extra section as optional? I am somewhat worried that this will make Bazel look more complex than it is?

mostynb: Maybe I could move it into a workspace bazelrc file? Not sure if that's possible

BUILD.bazel
mostynb: I moved the variable to the main package because that's where it's first used, and it seems mostly unrelated to the server package. I'm not sure how to refer to the main package in the "import" naming style, but this seems to work.

The rules_go documentation seems to suggest creating a version library, would you prefer that?
https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#defines-and-stamping

README.md
mostynb: The .bazelrc file setup seems to work. I couldn't figure out a way to specify <the workspace path> in the file so I embedded the one-line script there instead.

PR #126, opened 2019-11-25
use sync.Mutex instead of sync.RWMutex

mostynb:
Apart from CurrentSize and NumItems (which are used by the http status
page), we only lock this mutex for rw, never for reading only. So let's
use the simpler sync.Mutex type.

Also, use the de-facto standard variable name for mutexes in Go code, "mu".

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -31,7 +31,7 @@ func (i *lruItem) Size() int64 {
 // diskCache is an implementation of the cache backed by files on a filesystem.
 type diskCache struct {
 	dir string
-	mux *sync.RWMutex
+	mu  *sync.Mutex
 	lru SizedLRU
 }
 
@@ -71,7 +71,7 @@ func New(dir string, maxSizeBytes int64) cache.Cache {
 
 	cache := &diskCache{
 		dir: filepath.Clean(dir),
-		mux: &sync.RWMutex{},
+		mu:  &sync.Mutex{},
 		lru: NewSizedLRU(maxSizeBytes, onEvict),
 	}
 
@@ -158,7 +158,7 @@ func (c *diskCache) loadExistingFiles() error {
 }
 
 func (c *diskCache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.Reader) error {
-	c.mux.Lock()
+	c.mu.Lock()
 
 	key := cacheKey(kind, hash)
 
@@ -168,7 +168,7 @@ func (c *diskCache) Put(kind cache.EntryKind, hash string, expectedSize int64, r
 	// condition on the bazel side) but it's useful to overwrite poisoned items.
 	if existingItem, found := c.lru.Get(key); found {
 		if !existingItem.(*lruItem).committed {
-			c.mux.Unlock()
+			c.mu.Unlock()
 			io.Copy(ioutil.Discard, r)
 			return nil
 		}
@@ -180,7 +180,7 @@ func (c *diskCache) Put(kind cache.EntryKind, hash string, expectedSize int64, r
 		committed: false,
 	}
 	ok := c.lru.Add(key, newItem)
-	c.mux.Unlock()
+	c.mu.Unlock()
 	if !ok {
 		return &cache.Error{
 			Code: http.StatusInsufficientStorage,
@@ -193,8 +193,8 @@ func (c *diskCache) Put(kind cache.EntryKind, hash string, expectedSize int64, r
 	// but this stuff is really easy to get wrong without defer().
 	shouldCommit := false
 	defer func() {
-		c.mux.Lock()
-		defer c.mux.Unlock()
+		c.mu.Lock()
+		defer c.mu.Unlock()
 
 		if shouldCommit {
 			newItem.committed = true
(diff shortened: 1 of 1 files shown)


buchgr:
Good catch. I think we should probably use a different lock for reading and writing though?

mostynb:
> Good catch. I think we should probably use a different lock for reading and writing though?

CurrentSize and NumItems are the only methods that would be able to use a read-only lock, and they're only called from the http status page. I assume the status page is loaded relatively infrequently so it might be better to just use a simple Mutex, which I assume is slightly faster than RWMutex for write-heavy workloads like we have.

MaxSize acquires the lock but only reads from a field that is never modified, it may as well just read
the field without locking. I will push a new commit to remove it.

mostynb:
I tried to benchmark sync.Mutex vs sync.RWMutex's write lock to confirm my assumption (RWMutex should be slower in write-heavy scenarios). I used BenchmarkMutexWrite from this stack overflow question:
https://stackoverflow.com/questions/57562606/why-does-sync-mutex-largely-drop-performance-when-goroutine-contention-is-more-t

Then I changed it to use sync.RWMutex instead and compared the results with benchstat. The "old" column here uses sync.Mutex and "new" uses sync.RWMutex:

name                          old time/op  new time/op  delta
MutexWrite/goroutines-2400-8   163ns ± 3%   261ns ± 2%   +59.68%  (p=0.008 n=5+5)
MutexWrite/goroutines-2480-8   165ns ± 3%   273ns ± 3%   +65.45%  (p=0.008 n=5+5)
MutexWrite/goroutines-2560-8   170ns ± 4%   292ns ± 2%   +71.88%  (p=0.008 n=5+5)
MutexWrite/goroutines-2640-8   181ns ± 7%   366ns ± 1%  +102.10%  (p=0.008 n=5+5)
MutexWrite/goroutines-2720-8   317ns ± 2%   372ns ± 1%   +17.34%  (p=0.008 n=5+5)
MutexWrite/goroutines-2800-8   322ns ± 2%   376ns ± 6%   +16.75%  (p=0.008 n=5+5)
MutexWrite/goroutines-2880-8   318ns ± 2%   366ns ± 6%   +15.03%  (p=0.008 n=5+5)
MutexWrite/goroutines-2960-8   318ns ± 2%   370ns ± 3%   +16.50%  (p=0.008 n=5+5)
MutexWrite/goroutines-3040-8   326ns ± 3%   362ns ± 2%   +11.17%  (p=0.008 n=5+5)
MutexWrite/goroutines-3120-8   332ns ± 1%   367ns ± 1%   +10.61%  (p=0.008 n=5+5)
MutexWrite/goroutines-3200-8   339ns ± 1%   373ns ± 1%   +10.15%  (p=0.008 n=5+5)
MutexWrite/goroutines-3280-8   354ns ± 2%   383ns ± 1%    +8.43%  (p=0.008 n=5+5)
MutexWrite/goroutines-3360-8   356ns ± 1%   394ns ± 3%   +10.72%  (p=0.008 n=5+5)
MutexWrite/goroutines-3440-8   369ns ± 2%   411ns ± 2%   +11.33%  (p=0.008 n=5+5)
MutexWrite/goroutines-3520-8   375ns ± 3%   409ns ± 2%    +9.00%  (p=0.008 n=5+5)
MutexWrite/goroutines-3600-8   376ns ± 4%   431ns ± 1%   +14.69%  (p=0.008 n=5+5)
MutexWrite/goroutines-3680-8   383ns ± 6%   421ns ± 1%    +9.86%  (p=0.008 n=5+5)
MutexWrite/goroutines-3760-8   364ns ± 7%   431ns ± 1%   +18.40%  (p=0.008 n=5+5)
MutexWrite/goroutines-3840-8   356ns ± 6%   437ns ± 2%   +22.88%  (p=0.008 n=5+5)
MutexWrite/goroutines-3920-8   357ns ± 2%   428ns ± 1%   +19.76%  (p=0.008 n=5+5)
MutexWrite/goroutines-4000-8   362ns ± 4%   396ns ± 2%    +9.44%  (p=0.008 n=5+5)
MutexWrite/goroutines-4080-8   366ns ± 4%   413ns ± 3%   +13.02%  (p=0.008 n=5+5)
MutexWrite/goroutines-4160-8   371ns ± 4%   411ns ± 3%   +10.61%  (p=0.008 n=5+5)
MutexWrite/goroutines-4240-8   370ns ± 3%   413ns ± 3%   +11.39%  (p=0.008 n=5+5)
MutexWrite/goroutines-4320-8   380ns ± 1%   411ns ± 3%    +8.27%  (p=0.008 n=5+5)
MutexWrite/goroutines-4400-8   392ns ± 2%   421ns ± 6%    +7.29%  (p=0.008 n=5+5)
MutexWrite/goroutines-4480-8   388ns ± 2%   420ns ± 4%    +8.19%  (p=0.008 n=5+5)
MutexWrite/goroutines-4560-8   392ns ± 2%   427ns ± 1%    +8.87%  (p=0.008 n=5+5)
MutexWrite/goroutines-4640-8   398ns ± 1%   431ns ± 2%    +8.24%  (p=0.008 n=5+5)
MutexWrite/goroutines-4720-8   411ns ± 3%   443ns ± 3%    +7.84%  (p=0.008 n=5+5)

PR #128, opened 2019-11-25
rely on umask to create files/dirs with the requested permissions

mostynb:
This allows the admin to decide the filesystem permissions used for the
cache, for example to allow administration by a group without root access.

Fixes issue #78.

I think this solution is a bit simpler than #79.
@ob, @bdittmer: does this work for you?


ob:
Yeah, functionally equivalent and simpler! Thank you!

buchgr:
Nice @mostynb !

Issue #15, opened 2018-03-06
Allow per project cache


buchgr:
@BenTheElder mentioned to me that k8s is using a different remote cache for different projects and they distinguish between projects via a path component in the URL:

```https://cache.com/<project name>/(ac|cas)/```

I think that's a good idea and I believe @nicolov already laid the foundations for that.

**Comment by gjasny** [comment]
I started a prove-of-concept patch for on-disk storage here: https://github.com/gjasny/bazel-remote/tree/project_name

I noticed that for `cas` objects the cache enforces that the key equals the hashed content, so adding project name support there does not make much sense.

**Comment by BenTheElder** [comment]
the CAS only needs to enforce this for the last segment in the URL, so it's pretty easy to do this. there's an implementation here that inspired this issue: https://github.com/kubernetes/test-infra/tree/master/greenhouse

**Comment by sitsofe** [comment]
Looking through some logs with this in place I see the following:
```
020/09/08 11:44:29 REMAP AC HASH ed3360a7998f204de2c2f3311a2d8563d4f87bdfb9b4b442daee115f9c0bdb8a : centos-v2-Amd => 4699610a00158646b15c49ae315da98c
658db3cd2b0f376cc48de0c65f692993
2020/09/08 11:44:29  PUT 200     X.X.X.X /centos-v2-Amd/ac/ed3360a7998f204de2c2f3311a2d8563d4f87bdfb9b4b442daee115f9c0bdb8a
2020/09/08 11:44:29 S3 UPLOAD bazel-remote-prod bazel/cas/929d6d967df38d7d70128171a34ea8794603e05f5689d7996d06639eb05375b7 OK
2020/09/08 11:44:29 REMAP AC HASH a9c38c68068a703d5a3bcbba3729d0f21d78f515ef91c88fb5ebf798389caeaa : centos-v2-Amd => 105a6b03847b7df71ba1f2c7f3769f88
69405aa859abf4965344ef1e81cb7902
2020/09/08 11:44:29 S3 UPLOAD bazel-remote-prod bazel/cas/ed3360a7998f204de2c2f3311a2d8563d4f87bdfb9b4b442daee115f9c0bdb8a OK
```

- 4699610a00158646b15c49ae315da98c was mangled to ed3360a7998f204de2c2f3311a2d8563d4f87bdfb9b4b442daee115f9c0bdb8a
[...]

PR #148, opened 2019-12-28
Differentiate Action Cache objects by instance name

gjasny:
Hello,

Every now and then a Clang update in our toolchain breaks the remote caching (usually fiffreneces in system headers like clang version numbers jumping from `7.0.0` to `7.1.0`). A suggested work-around is to separate the caches.

For http this is possible with `--remote_cache=http://localhost:8080/foo` and for grpc by setting `--remote_instance_name=bar`.

For now the instance name only affects the action cache because in my use case the CAS content cannot differ.

I'd appreciate some feedback. (I'm also a golang beginner - please bear with me)

Thanks,
Gregor

Fixes: #15

### DIFF
--- a/server/grpc_ac.go
+++ b/server/grpc_ac.go
@@ -43,15 +43,15 @@ func (s *grpcServer) GetActionResult(ctx context.Context,
 		return nil, err
 	}
 
-	result, _, err := cache.GetValidatedActionResult(s.cache,
+	result, _, err := cache.GetValidatedActionResult(s.cache, req.GetInstanceName(),
 		req.ActionDigest.Hash)
 	if err != nil {
-		s.accessLogger.Printf("%s %s %s", errorPrefix, req.ActionDigest.Hash, err)
+		s.accessLogger.Printf("%s [%s] %s %s", errorPrefix, req.GetInstanceName(), req.ActionDigest.Hash, err)
 		return nil, status.Error(codes.Unknown, err.Error())
 	}
 
 	if result == nil {
-		s.accessLogger.Printf("%s %s NOT FOUND", errorPrefix, req.ActionDigest.Hash)
+		s.accessLogger.Printf("%s [%s] %s NOT FOUND", errorPrefix, req.GetInstanceName(), req.ActionDigest.Hash)
 		return nil, status.Error(codes.NotFound,
 			fmt.Sprintf("%s not found in AC", req.ActionDigest.Hash))
 	}
@@ -60,17 +60,17 @@ func (s *grpcServer) GetActionResult(ctx context.Context,
 
 	var inlinedSoFar int64
 
-	err = s.maybeInline(req.InlineStdout,
+	err = s.maybeInline(req.GetInstanceName(), req.InlineStdout,
 		&result.StdoutRaw, &result.StdoutDigest, &inlinedSoFar)
 	if err != nil {
-		s.accessLogger.Printf("%s %s %s", errorPrefix, req.ActionDigest.Hash, err)
+		s.accessLogger.Printf("%s [%s] %s %s", errorPrefix, req.GetInstanceName(), req.ActionDigest.Hash, err)
 		return nil, status.Error(codes.Unknown, err.Error())
 	}
 
-	err = s.maybeInline(req.InlineStderr,
+	err = s.maybeInline(req.GetInstanceName(), req.InlineStderr,
 		&result.StderrRaw, &result.StderrDigest, &inlinedSoFar)
 	if err != nil {
-		s.accessLogger.Printf("%s %s %s", errorPrefix, req.ActionDigest.Hash, err)
+		s.accessLogger.Printf("%s [%s] %s %s", errorPrefix, req.GetInstanceName(), req.ActionDigest.Hash, err)
 		return nil, status.Error(codes.Unknown, err.Error())
 	}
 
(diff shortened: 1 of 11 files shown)


server/grpc_ac.go
```diff
@@ -80,19 +80,19 @@ func (s *grpcServer) GetActionResult(ctx context.Context,
...
 	}
 
-	s.accessLogger.Printf("GRPC AC GET %s OK", req.ActionDigest.Hash)
+	s.accessLogger.Printf("GRPC AC GET [%s] %s OK", req.GetInstanceName(), req.ActionDigest.Hash)
 
 	return result, nil
 }
 
-func (s *grpcServer) maybeInline(inline bool, slice *[]byte, digest **pb.Digest, inlinedSoFar *int64) error {
+func (s *grpcServer) maybeInline(instanceName string, inline bool, slice *[]byte, digest **pb.Digest, inlinedSoFar *int64) error {
```
mostynb: I guess the instanceName arg is useless, since these items are stored in the CAS and this patch doesn't separate CAS entries by instance.

server/grpc_ac.go
```diff
@@ -114,8 +114,8 @@ func (s *grpcServer) maybeInline(inline bool, slice *[]byte, digest **pb.Digest,
 			}
 		}
 
-		if !s.cache.Contains(cache.CAS, (*digest).Hash) {
-			err := s.cache.Put(cache.CAS, (*digest).Hash, (*digest).SizeBytes,
+		if !s.cache.Contains(cache.CAS, instanceName, (*digest).Hash) {
+			err := s.cache.Put(cache.CAS, instanceName, (*digest).Hash, (*digest).SizeBytes,
```
mostynb: May as well pass an empty string for the instance, I guess.

server/grpc_cas.go
```diff
@@ -90,8 +90,8 @@ func (s *grpcServer) BatchUpdateBlobs(ctx context.Context,
 // Return the data for a blob, or an error.  If the blob was not
 // found, the returned error is errBlobNotFound. Only use this
 // function when it's OK to buffer the entire blob in memory.
-func (s *grpcServer) getBlobData(hash string, size int64) ([]byte, error) {
-	rdr, sizeBytes, err := s.cache.Get(cache.CAS, hash)
+func (s *grpcServer) getBlobData(instanceName string, hash string, size int64) ([]byte, error) {
```
mostynb: The instanceName arg seems to be useless here too, maybe just skip it and use the empty string?

server/grpc_cas.go
```diff
@@ -115,26 +115,26 @@ func (s *grpcServer) getBlobData(hash string, size int64) ([]byte, error) {
 	return data, rdr.Close()
 }
 
-func (s *grpcServer) getBlobResponse(digest *pb.Digest) *pb.BatchReadBlobsResponse_Response {
+func (s *grpcServer) getBlobResponse(instanceName string, digest *pb.Digest) *pb.BatchReadBlobsResponse_Response {
```
mostynb: Ditto.

server/grpc_cas.go
```diff
@@ -198,13 +198,13 @@ func (s *grpcServer) GetTree(in *pb.GetTreeRequest,
...
 
-	s.accessLogger.Printf("GRPC GETTREEREQUEST %s OK", in.RootDigest.Hash)
+	s.accessLogger.Printf("GRPC GETTREEREQUEST [%s] %s OK", in.GetInstanceName(), in.RootDigest.Hash)
 	return nil
 }
 
 // Attempt to populate `resp`. Return errors for invalid requests, but
 // otherwise attempt to return as many blobs as possible.
-func (s *grpcServer) fillDirectories(resp *pb.GetTreeResponse, dir *pb.Directory, errorPrefix string) error {
+func (s *grpcServer) fillDirectories(instanceName string, resp *pb.GetTreeResponse, dir *pb.Directory, errorPrefix string) error {
```
mostynb: Ditto.

server/http.go
```diff
@@ -67,35 +68,36 @@ func NewHTTPCache(cache cache.Cache, accessLogger cache.Logger, errorLogger cach
 }
 
 // Parse cache artifact information from the request URL
```
mostynb: Since we'd be returning two different strings, I think we should document what they are here now.

mostynb:
Thanks for the PR! I will try to look through it properly in the next few days.

> Every now and then a Clang update in our toolchain breaks the remote caching (usually fiffreneces in system headers like clang version numbers jumping from 7.0.0 to 7.1.0). A suggested work-around is to separate the caches.

Can you clarify what you mean by a toolchain update that "breaks the remote caching", and describe the suggested workaround a bit?

If you're referring to the fact that you get a bunch of cache misses after changing the toolchain, that is normal and expected.

mostynb:
I don't have plans to work on this myself because I don't understand the use case well enough, and it could potentially be a breaking change for some users.

If a toolchain is updated, shouldn't that change the corresponding Action messages, and therefore use different ActionCache lookup hashes?

kragniz:
@mostynb the use case is pretty much just to work around https://github.com/bazelbuild/bazel/issues/4558 without having to deploy multiple instances of bazel-remote. I imagine the behaviour change would be opt-in via a flag.

mostynb:
So to summarise:
* In some bazel client setups, actions can use unspecified input files, which means two different actions that produce different results can use the same ActionCache hash lookup key.
* As a workaround, the client configuration can be made to use a cache instance name that is generated from the unspecified inputs (eg by hashing relevant toolchain names/versions), if the cache server can be configured to give different action cache hits based on the instance name.

Is that right? If so, I think I can come up with a smaller opt-in feature to allow this.

mostynb:
@kragniz: please take a look at #339

Issue #154, opened 2020-01-07
add a profiling endpoint


mostynb:
I would like to have a profiling endpoint available, but disabled by default.

eg: if `--profile_port` is specified (with a unique+available port number), then start a new localhost http server with DefaultServeMux and the net/http/pprof URLs available. Otherwise, we do not start this server.

For this to work with docker, we can't listen on 127.0.0.1 (since we can't reach 127.0.0.1 inside the container from outside), so we could use another new flag, `--profile_host` which defaults to 127.0.0.1.

PR #155, opened 2020-01-08
add an optional http profiling endpoint

mostynb:
If --profile_port is specified (as a positive integer), then start a
http server on that port for runtime profiling data.

The host part of the address defaults to 127.0.0.1, and can be set
with the --profile_host flag. This is needed if you want to access
the profiling URLs from bazel-remote running inside docker.

Implements #154.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -56,6 +56,12 @@ container_image(
         "/app/bazel-remote-base.binary",
         "--port=8080",
         "--dir=/data",
+
+        # Listen on all addresses, not just 127.0.0.1, so this can
+        # be reached from outside the container (with a -p mapping).
+        "--profile_host=",
+        # Specify a port to enable the profiling http server.
+        "--profile_port=6060",
     ],
     ports = ["8080"],
     visibility = ["//visibility:public"],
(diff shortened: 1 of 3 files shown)


BUILD.bazel
```diff
@@ -56,6 +56,11 @@ container_image(
         "/app/bazel-remote-base.binary",
         "--port=8080",
         "--dir=/data",
+
+        # This defaults to 127.0.0.1 which is inaccessible outside the container.
+        "--profile_host=",
+        # This is disabled by default.
+        "--profile_port=6060",
```
buchgr: I am not sure I understand the comment "This is disabled by default". It looks enabled to me?

mostynb: Updated the comments.

The (separate) http server with pprof endpoints is only started if --profile_port is specified.

buchgr: Thanks! Do you mind updating the README.md to also mention the necessity to add a -p mapping if one wants to access the profiling endpoint?

mostynb: Thanks- I forgot that. Pushed a fix. There are some other help text changes from the urfave/cli upgrade, added the non-profiling changes in a second commit.

PR #162, opened 2020-01-16
refactor the cache.Cache interface

### LINKED ISSUE #161: refactor the cache.Cache interface
I think we should refactor the cache.Cache interface. At the moment we have a single cache.Cache interface that is used both by the disk backend and the proxy backends (s3, http/gcs). The proxy backends contain the same mostly copy+pasted logic for using the local/disk backend, and just forward some calls (like Stats) directly to the local/disk backend.

Instead, we could move the duplicated logic into the disk backend, and have the disk backend make calls to an optional proxy backend. The proxy backends would then implement a smaller interface, and their code would be simpler. This in turn would make it simpler to implement action cache data integrity checks in one place (the disk backend).

mostynb:
Since all the proxy backends use the same copy+pasted logic to use
the disk cache, refactor to put that functionality in the disk cache.
This simplifies the proxy backends.

Fixes #161.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -2,10 +2,6 @@ package cache
 
 import (
 	"io"
-	"io/ioutil"
-
-	pb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
-	"github.com/golang/protobuf/proto"
 )
 
 // EntryKind describes the kind of cache entry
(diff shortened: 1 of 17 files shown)


cache/cache.go
```diff
@@ -45,85 +42,20 @@ type Error struct {
 	Text string
 }
 
+var ErrNotFound = errors.New("Item not found")
```
mostynb: I should use the `cache.Error` type here.

cache/disk/disk.go
```diff
@@ -41,16 +45,18 @@ func (i *lruItem) Size() int64 {
 	return i.size
 }
 
-// diskCache is an implementation of the cache backed by files on a filesystem.
-type diskCache struct {
-	dir string
+// DiskCache is filesystem-based cache, with an optional backend proxy.
```
mostynb: Note that it is safe for concurrent use.

cache/cache.go
```diff
@@ -45,85 +42,20 @@ type Error struct {
...
-			return nil, nil, nil // aka "not found"
-		}
-	}
-
-	if result.StderrDigest != nil && result.StderrDigest.SizeBytes > 0 {
-		if !c.Contains(CAS, result.StderrDigest.Hash) {
-			return nil, nil, nil // aka "not found"
-		}
-	}
+// Cache backends implement this interface, and are optionally used by DiskCache.
```
mostynb: Add a note about concurrency assumptions?

cache/http/http_test.go
```diff
@@ -2,229 +2,162 @@ package http
...
 	"testing"
 
-	"github.com/buchgr/bazel-remote/cache"
-	"github.com/buchgr/bazel-remote/cache/disk"
-	"github.com/buchgr/bazel-remote/utils"
+	testutils "github.com/buchgr/bazel-remote/utils"
 )
 
-func TestProxyReadWorks(t *testing.T) {
-	// Test that reading a blob from a proxy works and also populates the local
```
mostynb: This functionality has moved to DiskCache, we should add a test there. And similarly for some of the other tests in this file.

cache/disk/disk.go
```diff
@@ -270,12 +280,32 @@ func (c *diskCache) Put(kind cache.EntryKind, hash string, expectedSize int64, r
 		shouldCommit = true
 	}
 
+	if c.proxy != nil {
+		// TODO: buffer in memory, avoid a filesystem round-trip!
+		fr, err := os.Open(filePath)
+		if err != nil {
```
mostynb: Muscle-memory tripped me up here. This should be if err == nil.

buchgr:
Please let me know when it's ready for review.

mostynb:
@buchgr: I think this is ready for review. I will try to stress test it some more today.

mostynb:
Seems to WFM building a large project like kubernetes.

buchgr:
Did you get a chance to test the S3 and GCS implementations? If you haven't I can test the GCS implementation but I don't have an AWS account.

mostynb:
I nave neither S3 nor GCS accounts to test- if you could test GCS that would be great.

I started working on some s3 tests with github.com/johannesboyne/gofakes3, but it seems to be broken with the minio-go client (if I Put and then Get an object, the bytes comes back wrapped with some sort of signature-chunk wrapper). I'll try setting up a minio server and do a full system test with it.

buchgr:
I ll give GCS a try and report back.

mostynb:
Pushed a couple of fixes and some log format changes, after testing with minio.

Building bazel-remote itself with a bazel-remote + minio works pretty well with gRPC, but it's quite a bit slower using HTTP. I need to investigate the difference.

lukedirtwalker:
Tests this quickly with gRPC and S3 and works fine.

mostynb:
> Tested this quickly with gRPC and S3 and works fine.

Background: Lukas was expecting cache hits from the proxy backend after removing the bazel-remote cache and restarting. This was fixed in this PR since it also proxies HEAD/Contains operations.

buchgr:
I just tested it with GCS. Seems to work great too!

mostynb:
Re the performance difference between using http and grpc between bazel and the cache, it seems to be unrelated to this change. I think it's just that uploading go stdlib (lots of small files) is much less efficient with http than with grpc:
https://github.com/bazelbuild/rules_go/issues/1531
https://github.com/bazelbuild/bazel/issues/6091

So I will rebase and land this if there are no objections.

buchgr:
> So I will rebase and land this if there are no objections.

I am giving it a review right now. 1sec please :-).

> Re the performance difference between using http and grpc between bazel and the cache, it seems to be unrelated to this change. 

I believe the reason for that is different. The gRPC implementation in Bazel uploads all outputs of an action concurrently while the HTTP implementation does so serially. This can be fixed in Bazel.

mostynb:
> I am giving it a review right now. 1sec please :-).

Yes, of course.

> I believe the reason for that is different. The gRPC implementation in Bazel uploads all outputs of an action concurrently while the HTTP implementation does so serially. This can be fixed in Bazel.

I figured that must be the case.

cache/disk/disk.go
```diff
@@ -264,55 +281,181 @@ func (c *diskCache) Put(kind cache.EntryKind, hash string, expectedSize int64, r
...
+	filePath := cacheFilePath(kind, c.dir, hash)
+	tmpFilePath := filePath + ".tmp"
+	shouldCommit := false
+	tmpFileCreated := false
+	foundSize := int64(-1)
+	var f *os.File
+
+	// We're allowed to try downloading this blob from the proxy.
+	// Before returning, we have to either commit the item and set
+	// its size, remove the item from the LRU.
```
buchgr: s/,/or/

cache/disk/disk.go
```diff
@@ -264,55 +281,181 @@ func (c *diskCache) Put(kind cache.EntryKind, hash string, expectedSize int64, r
...
+		}
+
+		f.Close() // No need to check the error.
+	}()
+
+	r, foundSize, err := c.proxy.Get(kind, hash)
+	if err != nil || r == nil {
+		return nil, -1, err
+	}
+	defer r.Close()
```
buchgr: Could this possibly leak a file descriptor if there is an err but r is not nil?

Consider

```
r, foundSize, err := ...
if r != nil {
  defer r.Close()
}
```

buchgr:
LGTM. Thanks @mostynb

cache/disk/disk.go
mostynb: Thanks, fixed.

cache/disk/disk.go
mostynb: Yes, good call. Fixed.

cache/http/http_test.go
```diff
@@ -2,229 +2,168 @@ package http
...
 
-	data := []byte("hello world")
-	hashBytes := sha256.Sum256(data)
-	hash := hex.EncodeToString(hashBytes[:])
+	return &ts
+}
 
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		defer r.Body.Close()
+func TestEverything(t *testing.T) {
```
uri-canva: This test is only testing the functionality of `newTestServer`, which is only used in this test. I think it can be removed.

mostynb: Hmm, I intended to update the tests to use this backend, but somehow the actual tests got lost in the shuffle. I'll try adding them back.

mostynb: https://github.com/buchgr/bazel-remote/pull/201

PR #177, opened 2020-02-05
disk cache: avoid leaking cache files in onEvict function

### LINKED ISSUE #176: uncommitted cache files leak when evicted from the LRU index
The disk cache's `onEvict` function is called on the least-recently used cache item after it has been removed from the index (map + list), to remove the file from disk. But if that cache item was uncommitted (ie, there was an active upload to a temp file) then `onEvict` does nothing:
https://github.com/buchgr/bazel-remote/blob/f99b78caaa2f0483cb838eda1c68cb9fc3d90b56/cache/disk/disk.go#L88

When such an upload finishes, the temp file is renamed, but it is no longer in the index and not counted as using up disk space.

I think this is an old bug that predates the `cache.Cache` refactoring- earlier, random temp files whose names were known only to the in-progress `Put` call goroutine were used for uploading blobs, which is probably why `onEvict` did not attempt the remove them. But since we now use a predictable temp file name (the blob name with `.tmp` appended), `onEvict` could remove the file.

mostynb:
If there's an ongoing blob upload when that blob is evicted, then the
item was deleted from the LRU index, but the files on disk were left
behind and no longer included in the cache size accounting.

When we evict a blob that's being uploaded, we might need to remove:
1) Both the temp file and the regular cache file.
2) Just the regular cache file.

Note that if you're encountering this situation, it might be a sign
that your cache size is too small- when blob uploads start the item
is moved to the most-recently used end of the index, and the item
has moved all the way to the least-recently used end of the index
before the upload has completed.

Fixes #176.

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -360,13 +360,16 @@ func (c *DiskCache) Put(kind cache.EntryKind, hash string, expectedSize int64, r
 	return err
 }
 
-// Return true if the item is in the cache and available.
-// Otherwise if it's ok to proxy, return a non-nil uncommitted *lruItem.
-// The caller must either set the size and commit the item, or remove it
-// from the LRU.
-func (c *DiskCache) availableOrTryProxy(key string) (bool, *lruItem) {
+// Return two bools, `available` is true if the item is in the local
+// cache and ready to use.
+//
+// `tryProxy` is true if the item is not in the local cache but can
+// be requested from the proxy, in which case, a placeholder entry
+// has been added to the index and the caller must either replace
+// the entry with the actual size, or remove it from the LRU.
+func (c *DiskCache) availableOrTryProxy(key string) (available bool, tryProxy bool) {
 	inProgress := false
-	var newItem *lruItem
+	tryProxy = false
 
 	c.mu.Lock()
 
@@ -376,18 +379,19 @@ func (c *DiskCache) availableOrTryProxy(key string) (bool, *lruItem) {
 			inProgress = true
 		}
 	} else if c.proxy != nil {
-		newItem = &lruItem{
-			size:      -1, // Caller must fill this in later!
+		// Reserve a place in the LRU.
+		// The caller must replace or remove this!
+		tryProxy = c.lru.Add(key, &lruItem{
+			size:      0,
 			committed: false,
-		}
-		c.lru.Add(key, newItem)
+		})
 	}
 
 	c.mu.Unlock()
 
-	available := found && !inProgress
+	available = found && !inProgress
 
-	return available, newItem
+	return available, tryProxy
 }
 
 // Get returns an io.ReadCloser with the content of the cache item stored under `hash`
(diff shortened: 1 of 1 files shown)


cache/disk/disk.go
```diff
@@ -85,13 +85,55 @@ func New(dir string, maxSizeBytes int64, proxy cache.CacheProxy) *DiskCache {
...
+		
+		f := filepath.Join(dir, key.(string))
+
 		if value.(*lruItem).committed {
-			err := os.Remove(filepath.Join(dir, key.(string)))
+			// Common case. Just remove the cache file and we're done.
+			err := os.Remove(f)
 			if err != nil {
-				log.Println(err)
+				log.Printf("ERROR: failed to remove evicted cache file: %s", f)
```
buchgr: Shouldn't this be a fatal error that crashes the server? There seems something seriously wrong if we can't evict?

cache/disk/disk.go
```diff
@@ -85,13 +85,55 @@ func New(dir string, maxSizeBytes int64, proxy cache.CacheProxy) *DiskCache {
...
+		// too small (blobs are moved to the most-recently used end
+		// of the index when the upload begins, and these items are
+		// still uploading when they reach the least-recently used
+		// end of the index).
+
+		tf := f + ".tmp"
+		var fErr, tfErr error
+		removedCount := 0
+
+		tfErr = os.Remove(tf)
```
buchgr: Two thoughts:
1) Instead of only deleting the file shouldn't it rather also cancel the upload?
2) I believe one Windows one can't delete a file that's open. That being said I don't know if anyone runs bazel-remote on Windows but it should sure be taken into consideration.

buchgr:
Great catch!

cache/disk/disk.go
mostynb: I think fatal errors are OK during startup, but I'm not sure it's a great idea to kill the server so easily during normal operation- especially if it's a transient error. I wonder if there's another reasonable way to notify the admin somehow?

cache/disk/disk.go
mostynb: > 1. Instead of only deleting the file shouldn't it rather also cancel the upload?

Maybe, but that would require a bit of refactoring, eg storing a context in the lruItem.

Or maybe the item should be given some bias and periodically moved to the MRU end of the index while it's being uploaded (with some deadline)?

But IMO this is mostly a sign that the cache size is too small for this use case.

> 2. I believe one Windows one can't delete a file that's open. That being said I don't know if anyone runs bazel-remote on Windows but it should sure be taken into consideration.

Right. I think this would also be a problem in the regular eviction case on windows- there can be ongoing downloads of an item that is evicted, then we can't remove the file. To handle windows well would probably require a fancy file management layer, or some ugly retry loops.

mostynb:
These two fixes seem to be working (stress test includes a small disk cache, large blobs, and proxy backend):
https://github.com/buchgr/bazel-remote/issues/176#issuecomment-582826035

mostynb:
I think we should land these two fixes soon.

Filed #184 to figure out what the status of windows support is, that's probably a larger task.

PR #180, opened 2020-02-05
http: Add support for receiving and sending /ac/ protobuf in JSON encoding.

bentonj-vmw:
We are working on adding caching to SCons using the bazel-remote server, ideally we want to avoid adding the python protobuf dependencies into SCons and thus it is much better for us if we can use the JSON protobuf encoding. Here I add support for that to the http server by using the "Accept:" header for a GET request and a "Content-Type" header for a PUT.

### DIFF
--- a/server/BUILD.bazel
+++ b/server/BUILD.bazel
@@ -16,6 +16,7 @@ go_library(
         "//cache/disk:go_default_library",
         "@com_github_bazelbuild_remote_apis//build/bazel/remote/execution/v2:go_default_library",
         "@com_github_bazelbuild_remote_apis//build/bazel/semver:go_default_library",
+        "@com_github_golang_protobuf//jsonpb:go_default_library_gen",
         "@com_github_golang_protobuf//proto:go_default_library",
         "@go_googleapis//google/bytestream:bytestream_go_proto",
         "@go_googleapis//google/rpc:code_go_proto",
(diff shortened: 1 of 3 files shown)


server/http.go
```diff
@@ -127,18 +128,35 @@ func (h *httpCache) handleGetValidAC(w http.ResponseWriter, r *http.Request, has
...
-		h.logResponse(http.StatusInternalServerError, r)
-		return
-	}
+		w.Header().Set("Content-Type", "application/json")
+		marshaler := jsonpb.Marshaler{}
+		err = marshaler.Marshal(w, ar)
+		if err != nil {
+			h.logResponse(http.StatusInternalServerError, r)
+			return
+		}
```
mostynb: Maybe just return at the end of this block, then we don't need the "else" (handle the protobuf case at one indent level less).

server/BUILD.bazel
```diff
@@ -16,6 +16,7 @@ go_library(
         "//cache/disk:go_default_library",
         "@com_github_bazelbuild_remote_apis//build/bazel/remote/execution/v2:go_default_library",
         "@com_github_bazelbuild_remote_apis//build/bazel/semver:go_default_library",
+        "@com_github_golang_protobuf//jsonpb:go_default_library",
```
mostynb: `bazel run //:gazelle fix` on this commit produces a modified server/BUILD.bazel file for me- can you apply those changes?

mostynb:
Oh, and please mention this in the README.md.

PR #189, opened 2020-02-13
bazelci: mark bazel-remote binaries as build artifacts

mostynb:
These might be usable for binary releases.


mostynb:
This seems to be blocked on https://github.com/bazelbuild/continuous-integration/issues/563

mostynb:
This was fixed by #396.

PR #191, opened 2020-02-13
demonstrate bazelci bug

mostynb:


mostynb:
Fixed in the bazelbuild/continuous-integration repository.

PR #199, opened 2020-02-18
disk.go: Return the correct data for action result

philwo:
The code accidentally overwrote the contents of the outer `data` variable (containing a serialized ActionResult protobuf) with the bytes from the last Tree protobuf processed from the ActionResult's OutputDirectories.

The server then returned the Tree protobuf to Bazel via the `return` in the last line. However, Bazel was expecting the `ActionResult` protobuf of course. This caused the weirdest things to happen, most notably builds mysteriously failing due to output files not existing or Bazel complaining that some file names are not valid UTF-8.

This was an accidental regression introduced by https://github.com/buchgr/bazel-remote/commit/b2e2f705f43d20e12dd8139f1719e7bd9c40a9b2.

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -607,13 +607,13 @@ func (c *DiskCache) GetValidatedActionResult(hash string) (*pb.ActionResult, []b
 		return nil, nil, nil // aka "not found"
 	}
 
-	data, err := ioutil.ReadAll(rdr)
+	acdata, err := ioutil.ReadAll(rdr)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


buchgr:
nice catch. thanks!

philwo:
Sorry, there was a build error due to a missing `var` - I'll send a follow-up! (Shouldn't CI catch this? 🤔)

mostynb:
> Shouldn't CI catch this?

Looks like it did, but the change was merged before the CI failed.

PR #211, opened 2020-02-26
fix canceled writes remain in progress state

slofurno:
Let's say this is WIP, but I'm putting it here to get some feedback.

I've been running bazel-remote as a cas for buildfarm using the grpc endpoint, and I have noticed some uploads have been getting stuck in a loop.

What seems to be happening is the first attempt will die somehow, maybe someone's computer is turned off abruptly or there is some network issue. This will leave the disk cache in a state where the upload is "in progress", but never finishes or errors. At this point I can see a half uploaded .tmp file.

Since the disk cache thinks the upload is still in progress, subsequent attempts to upload the file are discarded, yet responded to with success. Bazel will then attempt to read the file it thinks it just uploaded, see its not in the cas, and the process repeats itself.

It doesn't seem like bazel-remote handles grpc deadlines anywhere, but I'm not familiar enough with grpc to know if Recv() handles that internally.

I haven't been able to test this yet, but I will try to reproduce this tomorrow with this change.

### DIFF
--- a/server/grpc_bytestream.go
+++ b/server/grpc_bytestream.go
@@ -337,6 +337,7 @@ func (s *grpcServer) Write(srv bytestream.ByteStream_WriteServer) error {
 			break
 		}
 		if err != nil {
+			pw.CloseWithError(err)
 			s.accessLogger.Printf("GRPC BYTESTREAM WRITE FAILED: %s",
 				err.Error())
 			return err
--- a/server/grpc_test.go
+++ b/server/grpc_test.go
@@ -23,6 +23,7 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/grpc/test/bufconn"
 
+	"github.com/buchgr/bazel-remote/cache"
 	"github.com/buchgr/bazel-remote/cache/disk"
 	"github.com/buchgr/bazel-remote/utils"
 )
@@ -41,6 +42,7 @@ var (
 	casClient pb.ContentAddressableStorageClient
 	bsClient  bytestream.ByteStreamClient
 	ctx       = context.Background()
+	diskCache *disk.DiskCache
 
 	badDigestTestCases = []badDigest{
 		{digest: pb.Digest{Hash: ""}, reason: "empty hash"},
@@ -65,7 +67,7 @@ func TestMain(m *testing.M) {
 	}
 	defer os.RemoveAll(dir)
 
-	diskCache := disk.New(dir, int64(10*maxChunkSize), nil)
+	diskCache = disk.New(dir, int64(10*maxChunkSize), nil)
 
 	accessLogger := testutils.NewSilentLogger()
 	errorLogger := testutils.NewSilentLogger()
@@ -407,6 +409,95 @@ func TestGrpcAcRequestInlinedBlobs(t *testing.T) {
 		if r.Status.GetCode() != int32(codes.OK) {
 			t.Fatal("missing blob:", r.Digest)
 		}
+	}
+}
+
+func TestGrpcByteStreamDeadline(t *testing.T) {
+	testBlobSize := int64(16)
+	testBlob, testBlobHash := testutils.RandomDataAndHash(testBlobSize)
+	testBlobDigest := pb.Digest{
+		Hash:      testBlobHash,
+		SizeBytes: int64(len(testBlob)),
+	}
+
+	instance := "deadlineExpired"
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*10)
+	defer cancel()
+
+	bswc, err := bsClient.Write(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	resourceName := fmt.Sprintf(
+		"%s/uploads/%s/blobs/%s/%d/deadline/metadata/here",
+		instance,
+		uuid.New().String(),
+		testBlobDigest.Hash,
+		len(testBlob),
+	)
+
+	for i := 0; i < len(testBlob); i++ {
+		bswReq := bytestream.WriteRequest{
+			ResourceName: resourceName,
+			FinishWrite:  false,
+			Data:         testBlob[i : i+1],
+			WriteOffset:  int64(i),
+		}
+
+		err := bswc.Send(&bswReq)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		time.Sleep(time.Millisecond)
+	}
+
+	_, err = bswc.CloseAndRecv()
+
+	statusError, ok := status.FromError(err)
+	if !ok {
+		t.Fatalf("expected rpc error code, got %v\n", err)
+	}
+
+	if code := statusError.Code(); code != codes.DeadlineExceeded {
+		t.Fatalf("expected codes.DeadlineExceeded, got %s\n", code.String())
+	}
+
+	ctx, cancel = context.WithTimeout(context.Background(), time.Millisecond*500)
+	defer cancel()
+
+	bswc, err = bsClient.Write(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	bswReq := bytestream.WriteRequest{
+		ResourceName: resourceName,
+		FinishWrite:  false,
+		Data:         testBlob,
+	}
+	err = bswc.Send(&bswReq)
+	if err != nil {
+		t.Fatalf("send error: %v\n", err)
+	}
+
+	_, err = bswc.CloseAndRecv()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, sz, err := diskCache.Get(cache.CAS, testBlobHash)
+	if err != nil {
+		t.Fatalf("get error: %v\n", err)
+	}
+
+	if sz != int64(len(testBlob)) {
+		t.Errorf("expected size: %d, got: %d\n", len(testBlob), sz)
 	}
 }
 


elianuber:
@slofurno So if we can detect the client cancelled the upload we can just delete the upload completely from the CAS so next client won't find it in progress or anywhere and re-upload, right ?

slofurno:
Improved the test to hopefully make it more clear what is going on

Recv does return an error when the deadline is exceeded, all that was missing before was a call to close the pipewriter. Just closing that with an error is enough to ensure the disk cache winds up in the correct state.

If you run the TestGrpcByteStreamDeadline test with CloseWithError commented out, you will see how the second write appears to be successful, but the cache still thinks the upload is in progress.

slofurno:
@mostynb let me know if there is something else i can improve on this pr

mostynb:
Sorry for the delay, I will try to look at this soon.

server/grpc_bytestream.go
```diff
@@ -326,28 +326,19 @@ func (s *grpcServer) Write(srv bytestream.ByteStream_WriteServer) error {
 	}()
 
 	select {
-	case err, ok := <-recvResult:
-		if !ok {
-			msg := "Receive loop closed unexpectedly."
-			s.accessLogger.Printf("GRPC BYTESTREAM WRITE FAILED: %s", msg)
-			return status.Error(codes.Internal, msg)
-		}
```
mostynb: If I understand correctly, the only change required to this file is an added `pw.CloseWithError(err)` ? If so, please undo the other changes.

server/grpc_test.go
```diff
@@ -41,6 +42,7 @@ var (
 	casClient pb.ContentAddressableStorageClient
 	bsClient  bytestream.ByteStreamClient
 	ctx       = context.Background()
+	diskCache *disk.DiskCache
```
mostynb: Would this cause trouble if we were to run tests concurrently?

server/grpc_test.go
```diff
@@ -410,6 +412,65 @@ func TestGrpcAcRequestInlinedBlobs(t *testing.T) {
...
+
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(10*time.Millisecond))
+	defer cancel()
+
+	bswc, err := bsClient.Write(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	resourceName := fmt.Sprintf("%s/uploads/%s/blobs/%s/%d/deadline/metadata/here", instance, uuid.New().String(), testBlobDigest.Hash, len(testBlob))
```
mostynb: Please wrap this long line.

server/grpc_test.go
```diff
@@ -410,6 +412,65 @@ func TestGrpcAcRequestInlinedBlobs(t *testing.T) {
...
+	fmt.Println(testBlobHash)
```
mostynb: This doesn't look particularly useful, should we remove it? If not, please use `t.Log()` instead.

server/grpc_test.go
```diff
@@ -422,7 +422,8 @@ func TestGrpcByteStreamDeadline(t *testing.T) {
 
 	instance := "deadlineExpired"
 
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(10*time.Millisecond))
+	//ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(10*time.Millisecond))
```
mostynb: Remove this commented-out line?

server/grpc_test.go
```diff
@@ -464,10 +465,36 @@ func TestGrpcByteStreamDeadline(t *testing.T) {
 	}
 
 	time.Sleep(time.Millisecond)
```
mostynb: Is this tiny sleep required?

server/grpc_test.go
```diff
@@ -464,10 +465,36 @@ func TestGrpcByteStreamDeadline(t *testing.T) {
...
+	ctx, cancel = context.WithTimeout(context.Background(), time.Millisecond*500)
```
mostynb: The code from this point onwards feels like it could be a separate test.

But maybe this is already covered by `TestGrpcByteStream`, which does a bytestream Write followed by Read- I guess that's effectively the same, just without a timeout. Maybe we should add a timeout there instead?

server/grpc_test.go
```diff
@@ -410,6 +412,65 @@ func TestGrpcAcRequestInlinedBlobs(t *testing.T) {
...
+			Data:         testBlob[i : i+1],
+			WriteOffset:  int64(i),
+		}
+
+		err := bswc.Send(&bswReq)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			t.Error(err)
```
mostynb: Should this be `t.Fatal(err)` ?

server/grpc_test.go
```diff
@@ -410,6 +412,65 @@ func TestGrpcAcRequestInlinedBlobs(t *testing.T) {
...
+
+	_, err = bswc.CloseAndRecv()
+
+	statusError, ok := status.FromError(err)
+	if !ok {
+		t.Fatalf("expected rpc error code, got %v\n", err)
+	}
+
+	if code := statusError.Code(); code != codes.DeadlineExceeded {
+		t.Errorf("expected codes.DeadlineExceeded, got %s\n", code.String())
```
mostynb: Should this be `t.Fatalf(...)` ?

server/grpc_test.go
```diff
@@ -410,6 +412,65 @@ func TestGrpcAcRequestInlinedBlobs(t *testing.T) {
 	}
 }
 
+func TestGrpcByteStreamDeadline(t *testing.T) {
+	testBlobSize := int64(maxChunkSize * 3 / 2)
```
mostynb: In curious about the size of the test blob here. If it's larger than `maxChunkSize`, maybe we'll hit a chunk size error which we're not trying to test here. Maybe we should pick a a small blobsize like 10, a deadline of 10ms, 2ms sleeps between sending the next byte?

server/grpc_test.go
slofurno: This test shows that a canceled upload doesn't leave the cache in a bad state and prevent future uploads from being discarded.

This first portion of this test actually passes without the fix. We won't fail any assertions until trying to read the file from the cache, after the second successful write (which was actually discarded) and finding it is not there.

server/grpc_test.go
slofurno: I don't know if it makes much of a difference, I usually reserve fatal for failed assertions which would later panic

server/grpc_test.go
slofurno: done

server/grpc_bytestream.go
slofurno: done

server/grpc_test.go
slofurno: It seems to me the disk cache is safe for concurrent use

server/grpc_test.go
slofurno: done

server/grpc_test.go
mostynb: Ah, I see.

server/grpc_test.go
mostynb: The disk cache is safe for concurrent use, but I think it's possible for two concurrent tests to confuse each other if they share a cache (stats will be different, items can be evicted etc). Instead of doing `diskCache.Get()` the new test could do a bytestream `Read()` call, and then we don't need a package level `diskCache` variable.

server/grpc_test.go
```diff
@@ -410,6 +412,95 @@ func TestGrpcAcRequestInlinedBlobs(t *testing.T) {
...
+	}
+
+	bswReq := bytestream.WriteRequest{
+		ResourceName: resourceName,
+		FinishWrite:  false,
+		Data:         testBlob,
+	}
+	err = bswc.Send(&bswReq)
+	if err != nil {
+		t.Errorf("send error: %v\n", err)
```
mostynb: I guess this should be `t.Fatalf(...)` too, since it means the test setup has failed and any subsequent errors are likely to be due to the test setup failure and not due to the code that we're trying to test.

server/grpc_test.go
```diff
@@ -410,6 +412,95 @@ func TestGrpcAcRequestInlinedBlobs(t *testing.T) {
...
+	}
+
+	_, err = bswc.CloseAndRecv()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, sz, err := diskCache.Get(cache.CAS, testBlobHash)
+	if err != nil {
+		t.Errorf("get error: %v\n", err)
```
mostynb: I would make this `t.Fatalf(...)` also, since it is usually expected that if a function in go returns multiple values including a non-nil error, that the other return values should not be used (and so we should not check `sz` below if we get an error here).

server/grpc_test.go
slofurno: Would it not still be the case that the disk cache is still shared across tests through the single grpc server instance?

It looks like we would have to move the entirety of TestMain into a setup function which is called before each individual test.

mostynb: Hmm, you're right. I don't use TestMain often, and for some reason I thought that it was run once per test. Sorry about the distraction.

server/grpc_bytestream.go
```diff
@@ -359,12 +361,7 @@ func (s *grpcServer) Write(srv bytestream.ByteStream_WriteServer) error {
 		return err
 	}
 
-	err, ok := <-putResult
-	if !ok {
-		msg := "Cache Put closed unexpectedly."
-		s.accessLogger.Printf("GRPC BYTESTREAM WRITE FAILED: %s", msg)
-		return status.Error(codes.Unknown, msg)
-	}
+	err := <-putResult
```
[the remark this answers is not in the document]
slofurno: done

server/grpc_test.go
slofurno: done

server/grpc_test.go
slofurno: done

mostynb:
Pushed to master with minor modifications to the commit message. Thanks for contributing this subtle but important fix!

PR #228, opened 2020-03-13
Add gRPC / HTTP basic metrics, throughput, latency and sizes

elianuber:
In this PR,

- add HTTP / gRPC middleware to record endpoint metrics
- update go modules


main.go
```diff
@@ -23,7 +23,10 @@ import (
 
 	"github.com/buchgr/bazel-remote/config"
 	"github.com/buchgr/bazel-remote/server"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	httpmetrics "github.com/slok/go-http-metrics/metrics/prometheus"
+	httpmiddleware "github.com/slok/go-http-metrics/middleware"
```
mostynb: This is a relatively unknown library with only a single author. Could we use the promhttp import one line above instead?

elianuber: It's already using prometheus. This library is adding basic metrics for each HTTP endpoint.
See https://github.com/slok/go-http-metrics/blob/master/metrics/prometheus/prometheus.go
I just don't want to rewrite more code

elianuber:
@mostynb 
> * Are the module upgrades required for this to work? If not, could you do that in a separate commit in such a way that each commit still builds OK?
You're right, module upgrades not required I just fixed it. (only is gRPC library to 1.28).
> * Could you flesh out the commit messages a bit? What are the particular "golden metrics"?
I have updated the commits and README to explain what metrics we add.

main.go
mostynb: Oh, these are more useful metrics than `promhttp.Handler()` adds itself. OK.

Issue #233, opened 2020-03-23
add the empty blob optimization to HTTP also


mostynb:
The gRPC API implementation has some special handling for empty blobs. IIRC bazel was uploading ActionResults which reference the empty blob without ever actually uploading the empty blob itself. Which is kind of OK, because the empty blob can be "regenerated" trivially.

Let's add the same feature for the HTTP API.

Issue #231, opened 2020-03-20
Can we have disable_ac_validation for gRPC?


brentleyjones:
I love that we have `--disable_http_ac_validation`. Can we have a command that applies to both, or a new one that also applies to gRPC? I'm using bazel-remote as a backend for BuildBarn. Each cache instance is one of many and BuildBarn does the AC validation. Without this setting I can only use the HTTP mode of `bazel-remote`.

mostynb:
I wasn't aware of that usecase- Interesting. I don't see why not, we just need to find a suitable name.

The reason I added the `--disable_http_ac_validation` flag was to support using bazel-remote as a network-accessible LRU cache for generic file blobs (unrelated to REAPI), so it does not check that the blobs are valid `ActionResult` protos, and as a result does not do any dependency checking.

For gRPC I think we can assume that all the blobs are valid `ActionResult` protos since that is handled by the gRPC framework- but you would like to simply skip checking for all the dependent blobs in the CAS if this new flag is specified? I'm torn between `--disable_grpc_ac_validation` (consistent with the other flag, but slightly inaccurate) and something like `--disable_grpc_ac_dependency_check` (a bit of a mouthful).

brentleyjones:
I would vote for checking that it’s valid, just not the dependency checking. As for naming, I’ll leave that for you. Maybe we make it apply to both gRPC and HTTP since in both cases we would want the AC to be valid but maybe not check the dependencies?

mostynb:
Here's a quick hack for you to test:
https://github.com/mostynb/bazel-remote/tree/disable_grpc_ac_deps

If this works, and if it's somehow better than using HTTP, then I can clean it up before landing. But I suspect it might not be much different to just using HTTP. Please let me know how it goes.

(The original purpose of `--disable_grpc_ac_validation` was to allow arbitrary data, so the behaviour needs to differ for gRPC.)

brentleyjones:
Thanks @mostynb. I’ll check it out Monday.

brentleyjones:
That is better, but something is still slightly off. When running in HTTP mode with validation disabled I can get 100% cache hits with two instances split. With this I still don't get all cache hits.

mostynb:
Do you have a small-enough testcase that you could capture the access logs for, starting with two empty bazel-remote instances?

brentleyjones:
I'll look into that, yes.

brentleyjones:
[...]

PR #234, opened 2020-03-23
don't require explicit empty blob uploads for HTTP either

mostynb:
The gRPC API implementation has a hack to handle clients (like bazel?) which in some cases don't upload the empty blob, but do reference it. It is kind of OK not to upload the empty blob, since it can be recognised by hash and trivially regenerated.

Let's imlpement the hack for the HTTP API also.

Implements #233.

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -193,6 +193,14 @@ func (h *httpCache) CacheHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		if kind == cache.CAS && hash == emptySha256 {
+			w.Header().Set("Content-Type", "application/octet-stream")
+			w.Header().Set("Content-Length", "0")
+			w.Write([]byte{})
+			h.logResponse(http.StatusOK, r)
+			return
+		}
+
 		rdr, sizeBytes, err := h.cache.Get(kind, hash)
 		if err != nil {
 			if e, ok := err.(*cache.Error); ok {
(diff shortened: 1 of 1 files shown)


mostynb:
@brentleyjones: would you like to cherry-pick + test this with your setup?

brentleyjones:
@mostynb Yes! I'll report back soon-ish. Thanks!

brentleyjones:
@mostynb This patch worked. BuildBarn is now able to validate AC's for HTTP mode with this patch.

Sadly infrastructure issues are preventing me from continuing to use HTTP mode, so I'll also have to followup on #231 testing for you.

brentleyjones:
[gRPC-instance1.txt](https://github.com/buchgr/bazel-remote/files/4370557/gRPC-instance1.txt)
[gRPC-instance2.txt](https://github.com/buchgr/bazel-remote/files/4370558/gRPC-instance2.txt)
[HTTP-instance1.txt](https://github.com/buchgr/bazel-remote/files/4370559/HTTP-instance1.txt)
[HTTP-instance2.txt](https://github.com/buchgr/bazel-remote/files/4370560/HTTP-instance2.txt)

I inserted a newline between the two builds. As you can see in instance 2 for gRPC, it can't find `ed11bb857931deee5257fae1ec0aeb51fc2e8be493c4af44315108c485613543`, but in instance 2 for HTTP it can.

mostynb:
Are the logs above from a test branch with both this change and the commit from #231? If so, then the logic for disabling the gRPC AC dependency check must be broken- I can't see any `GRPC NODEPS AC GET ... OK` log lines.

brentleyjones:
Yes, it has both commits from this and #231. This PR worked perfectly and can be merged, IMO. (I thought I was commenting in #231.)

PR #235, opened 2020-03-24
add a flag to disable gRPC GetActionResult dependency checks

mostynb:
It is possible for bazel-remote to be used as a sharded backing store for other bazel caches, using HTTP with the `--disable_http_ac_validation` flag. In this scenario, CAS blobs that the ActionResult refers to may only be stored on another shard.

This patch allows similar functionality for gRPC with the new `--disable_grpc_ac_deps_check` flag. Unlike with HTTP, gRPC GetActionResult requests are are still checked to confirm that the result is a valid ActionResult message.

Implements #231.


mostynb:
@brentleyjones: this adds some startup logging to confirm the validation status. Does it work for you?

brentleyjones:
When using `--disable_http_ac_validation` and `--disable_grpc_ac_deps_check` together, I had to pass them like `--disable_http_ac_validation --disable_grpc_ac_deps_check` and _not_ `--disable_http_ac_validation true --disable_grpc_ac_deps_check true`. With the `true`, the last once wouldn't be applied. That was probably the plumbing issue from last time.

Giving that, it worked! Thanks @mostynb!

mostynb:
Ahh, I meant to clarify what you meant by `--disable_grpc_ac_deps_check true`...

These are both boolean flags which take no argument, and it appears that the cli library bazel-remote uses still stop processing command line args when it sees the first argument not attached to a flag.
ie `./foo -one two -three` will only process `-one` and leave `two -three` unprocessed as positional arguments (and bazel-remote doesn't use positional arguments, so everything from the first `true` onwards in your command line was being ignored).

PR #253, opened 2020-04-21
Fix big cardinality issue on HTTP metrics

elianuber:
@buchgr We have noticed that Prometheus has issues to keep up with high cardinality load for HTTP metrics (the scraping times out). This PR keeps cardinality low by using the HTTP method as handler ID (instead of the blob SHAs).
For more, see https://github.com/slok/go-http-metrics#custom-handler-id

### DIFF
--- a/main.go
+++ b/main.go
@@ -283,16 +283,19 @@ func main() {
 
 		diskCache := disk.New(c.Dir, int64(c.MaxSize)*1024*1024*1024, proxyCache)
 
+		metricsMdlw := httpmiddleware.New(httpmiddleware.Config{
+			Recorder: httpmetrics.NewRecorder(httpmetrics.Config{}),
+		})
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -283,16 +283,19 @@ func main() {
 
 		diskCache := disk.New(c.Dir, int64(c.MaxSize)*1024*1024*1024, proxyCache)
 
+		metricsMdlw := httpmiddleware.New(httpmiddleware.Config{
+			Recorder: httpmetrics.NewRecorder(httpmetrics.Config{}),
+		})
```
mostynb, line 288: This doesn't seem to be used until a little further below. Could you move it to just before the `mux.Handle`  calls that first use `metricsMdlw`?

elianuber: done!

mostynb:
Thanks again.

PR #259, opened 2020-05-01
Add endpoint that returns HTTP 200 when the cache is ready

elianuber:
On large disks >4TB the LRU index takes a few minutes to load and also the endpoint `/status` suffers from lock contention on high-loads (sometimes `/status` takes 10 seconds to respond).
This new endpoint just returns OK when the cache is up / ready.

### DIFF
--- a/cache/disk/lru.go
+++ b/cache/disk/lru.go
@@ -8,6 +8,12 @@ type sizedItem interface {
 	Size() int64
 }
 
+// Key is the type used for identifying cache items. For non-test code,
+// this is a string of the form "<keyspace>/<hash>". Some test code uses
+// ints for simplicity.
+//
+// TODO: update the test code to use strings too, then drop all the
+// type assertions.
 type Key interface{}
 
 // EvictCallback is the type of callbacks that are invoked when items are evicted.


elianuber:
@mostynb Good, is this a constant endpoint ? can you show me a `curl` example ?

mostynb:
`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` is the null SHA256 hash, so if your bazel-remote instance is listening on port 8080 then you can do something like this:

`curl --head --fail http://localhost:8080/cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`

You can also redirect stdout and stderr to /dev/null if you only want to use curl's exit code and don't want the console output.

This is handled here: https://github.com/buchgr/bazel-remote/blob/337c0dbe1dbdc3f95551dfd462c62acebb4b028f/server/http.go#L306
You can see that it does not require a cache lock, and will return OK even if the cache is empty.

elianuber:
Thank you! I think this should be enough then :+1:

mostynb:
Added a note about this to README.md: https://github.com/buchgr/bazel-remote/commit/a7d442dace50102973bcd91e5e15b45b9d0e1fac

PR #264, opened 2020-05-06
Add flag for enabling metrics and use bigger duration buckets

elianuber:
The default histogram buckets in prometheus is from 0 to 10 seconds, so anything that takes >10 secs cannot be tracked. This commit uses duration buckets to include bigger values.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -52,6 +52,7 @@ type Config struct {
 	IdleTimeout             time.Duration             `yaml:"idle_timeout"`
 	DisableHTTPACValidation bool                      `yaml:"disable_http_ac_validation"`
 	DisableGRPCACDepsCheck  bool                      `yaml:"disable_grpc_ac_deps_check"`
+	EnableEndpointMetrics   bool                      `yaml:"enable_endpoint_metrics"`
 }
 
 // New returns a validated Config with the specified values, and an error
@@ -60,7 +61,7 @@ func New(dir string, maxSize int, host string, port int, grpcPort int,
 	profileHost string, profilePort int, htpasswdFile string,
 	tlsCertFile string, tlsKeyFile string, idleTimeout time.Duration,
 	s3 *S3CloudStorageConfig, disableHTTPACValidation bool,
-	disableGRPCACDepsCheck bool) (*Config, error) {
+	disableGRPCACDepsCheck bool, enableEndpointMetrics bool) (*Config, error) {
 	c := Config{
 		Host:                    host,
 		Port:                    port,
@@ -78,6 +79,7 @@ func New(dir string, maxSize int, host string, port int, grpcPort int,
 		IdleTimeout:             idleTimeout,
 		DisableHTTPACValidation: disableHTTPACValidation,
 		DisableGRPCACDepsCheck:  disableGRPCACDepsCheck,
+		EnableEndpointMetrics:   enableEndpointMetrics,
 	}
 
 	err := validateConfig(&c)
--- a/main.go
+++ b/main.go
@@ -40,6 +40,9 @@ const (
 // is set through linker options.
 var gitCommit string
 
+// durationBuckets is the buckets used for Prometheus histograms in seconds.
+var durationBuckets = []float64{.5, 1, 2.5, 5, 10, 20, 40, 80, 160, 320}
+
 func main() {
 
 	log.SetFlags(logFlags)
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -351,6 +356,8 @@ func main() {
 				opts = append(opts, grpc.ChainStreamInterceptor(streamInterceptors...))
 				opts = append(opts, grpc.ChainUnaryInterceptor(unaryInterceptors...))
 
+				grpc_prometheus.EnableHandlingTimeHistogram(grpc_prometheus.WithHistogramBuckets(durationBuckets))
```
mostynb, line 359: The go-grpc-prometheus docs mention that this can be expensive for prometheus servers. Is this likely to be problematic? Also, does this add significant load to bazel-remote?

main.go
```diff
@@ -40,6 +40,9 @@ const (
 // is set through linker options.
 var gitCommit string
 
+// durationBuckets is the buckets used for Prometheus histograms in seconds.
+var durationBuckets = []float64{.5, 1, 2.5, 5, 10, 20, 40, 80, 160, 320}
```
mostynb, line 44: Is this something worth mentioning in the README.md, or is it something obvious from the server side?

main.go
elianuber: It mentions it as a precaution but I don't believe this will cause significant load. Do you prefer to put the handler metrics behind a option flag ? `--enable-endpoint-metrics`

main.go
elianuber: I believe the default values 0-10secs is not very useful. Blobs are usually big and latency can easily go past 10 secs for big blobs >500MB, we have seen this in our dash on peak hours. This new default values is more realistic.

elianuber: I can of course mention on the README too.

main.go
brentleyjones: A flag would be nice for people that don't use the metrics.

elianuber: done !! :-)

main.go
```diff
@@ -315,9 +315,22 @@ func main() {
...
+					DurationBuckets: durationBuckets,
+				}),
+			})
+			mux.Handle("/metrics", metricsMdlw.Handler("metrics", promhttp.Handler()))
+			mux.Handle("/status", metricsMdlw.Handler("status", http.HandlerFunc(h.StatusPageHandler)))
+			mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+				metricsMdlw.Handler(r.Method, http.HandlerFunc(cacheHandler)).ServeHTTP(w, r)
+			})
+		} else {
+			mux.Handle("/metrics", promhttp.Handler())
```
mostynb, line 330: Should this be skipped if metrics are disabled? Or maybe replaced by a 404 handler?

config/config.go
```diff
@@ -52,6 +52,7 @@ type Config struct {
 	IdleTimeout             time.Duration             `yaml:"idle_timeout"`
 	DisableHTTPACValidation bool                      `yaml:"disable_http_ac_validation"`
 	DisableGRPCACDepsCheck  bool                      `yaml:"disable_grpc_ac_deps_check"`
+	EnableEndpointMetrics   bool                      `yaml:"enable_endpoint_metrics"`
```
mostynb, line 55: Please also update the example configuration file in README.md.

main.go
```diff
@@ -200,6 +200,12 @@ func main() {
 			DefaultText: "false, ie enable ActionCache dependency checks",
 			EnvVars:     []string{"BAZEL_REMOTE_DISABLE_GRPS_AC_DEPS_CHECK"},
 		},
+		&cli.BoolFlag{
+			Name:        "enable_endpoint_metrics",
```
mostynb, line 204: Please also update the `--help` text in README.md.

main.go
mostynb: I don't have much experience with metrics dashboards- if it's something that would be obvious to people using these systems, then I don't think we need to document it.

config/config.go
elianuber: done!!

main.go
elianuber: good point, removed !

main.go
elianuber: done !

elianuber:
> lgtm- thanks. Could you please squash this to a single commit?

done !

PR #266, opened 2020-05-07
Consistent handling of empty files in CAS

ulrfa:
Hi,

There were special handling of empty files in server/grpc_cas.go and
server/http.go. It assumed that empty files are not stored in CAS.
The reason was to handle clients (like bazel?) which in some cases
don't upload the empty blob, but do reference it. It is kind of OK
not to upload the empty blob, since it can be recongnised by hash
and trivally regenerated.

**Problem was** that server/grpc_bytestream.go,
GetValidatedActionResult in disk.go and perhaps more places, still
expected the empty blobs to be available on disk.

**This commit moves the logic into the disk.go to assure the same
logic is used consistenly everywhere.**

Would you accept this pull request, if I add and update unit tests?

BR
Ulrik Falklöf

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -1,6 +1,7 @@
 package disk
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
@@ -59,6 +60,7 @@ type nameAndInfo struct {
 }
 
 const sha256HashStrSize = sha256.Size * 2 // Two hex characters per byte.
+const emptySha256 = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 
 // New returns a new instance of a filesystem-based cache rooted at `dir`,
 // with a maximum size of `maxSizeBytes` bytes and an optional backend `proxy`.
@@ -257,6 +259,11 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
 			len(hash), sha256.Size)
 	}
 
+	if kind == cache.CAS && expectedSize == 0 && hash == emptySha256 {
+		io.Copy(ioutil.Discard, r)
+		return nil
+	}
+
 	key := cacheKey(kind, hash)
 
 	c.mu.Lock()
@@ -403,14 +410,20 @@ func (c *Cache) availableOrTryProxy(key string) (available bool, tryProxy bool) 
 // Get returns an io.ReadCloser with the content of the cache item stored under `hash`
 // and the number of bytes that can be read from it. If the item is not found, the
 // io.ReadCloser will be nil. If some error occurred when processing the request, then
-// it is returned.
-func (c *Cache) Get(kind cache.EntryKind, hash string) (io.ReadCloser, int64, error) {
+// it is returned. The 'size' of the content to be retreived shall be provided when
+// known, or as -1 when unknown.
+func (c *Cache) Get(kind cache.EntryKind, hash string, size int64) (io.ReadCloser, int64, error) {
 
 	// The hash format is checked properly in the http/grpc code.
 	// Just perform a simple/fast check here, to catch bad tests.
 	if len(hash) != sha256HashStrSize {
 		return nil, -1, fmt.Errorf("Invalid hash size: %d, expected: %d",
 			len(hash), sha256.Size)
+	}
+
+	if kind == cache.CAS && size <= 0 && hash == emptySha256 {
+		cacheHits.Inc()
+		return ioutil.NopCloser(bytes.NewReader([]byte{})), 0, nil
 	}
 
 	var err error
@@ -423,11 +436,16 @@ func (c *Cache) Get(kind cache.EntryKind, hash string) (io.ReadCloser, int64, er
 		var fileInfo os.FileInfo
 		fileInfo, err = os.Stat(blobPath)
 		if err == nil {
+			foundSize := fileInfo.Size()
+			if isSizeMismatch(size, foundSize) {
+				cacheMisses.Inc()
+				return nil, -1, nil
+			}
 			var f *os.File
 			f, err = os.Open(blobPath)
 			if err == nil {
 				cacheHits.Inc()
-				return f, fileInfo.Size(), nil
+				return f, foundSize, nil
 			}
 		}
 
@@ -483,6 +501,9 @@ func (c *Cache) Get(kind cache.EntryKind, hash string) (io.ReadCloser, int64, er
 	if err != nil || r == nil {
 		return nil, -1, err
 	}
+	if isSizeMismatch(size, foundSize) {
+		return nil, -1, nil
+	}
 
 	f, err = os.Create(tmpFilePath)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


ulrfa:
> Do you know of existing clients that are broken with the current setup?

I don't know  (see my comment in #267) 

> Maybe we could add a size argument to disk.Get and disk.Contains and have the HTTP code specify -1 if the size is unkown? ... Then we can use a simple less than or equal to zero check before doing a string comparison with the empty hash.

Yes, good idea!

> Re the bytestream API, it is only intended to be used for large files (which are necessarily non-empty). It wouldn't hurt us to have a sanity check here too, but if you have clients doing this it might be worth reporting issues against them to switch to the ContentAddressableStorage APIs for small files, which I think are more efficient.

Good point! I don't know if there are such clients. Do you prefer grpc_bytstream.go to reject empty file requests with a clear error message? Or accept them and let disk.go handle also empty ones transparently?

> Re GetValidatedActionResult, I thought I covered all the corner cases where empty CAS blobs are valid, but it would not surprise me if I missed something.

Could the following be empty?

https://github.com/buchgr/bazel-remote/blob/69bce621e7ab073e92c7ec479e0708eb315d5b6a/cache/disk/disk.go#L663

https://github.com/buchgr/bazel-remote/blob/69bce621e7ab073e92c7ec479e0708eb315d5b6a/cache/disk/disk.go#L674

mostynb:
I think you're right about the two extra cases in GetValidatedActionResult. Here's a quickfix for those, and for the bytestream API (let's just handle this corner case, even though clients shouldn't use it): https://github.com/buchgr/bazel-remote/pull/270

Would you like to redo this PR, adding a size parameter to Get and Contains as described above (to be completed after I tag the next release)?

ulrfa:
Yes, I would like to redo this PR, but I wait until after your release. I will then add a size parameter to Get and Contains.

ulrfa:
The PR is now updated. Is it as you suggested @mostynb, with the added size parameter to Get and Contains?

cache/disk/disk_test.go
```diff
@@ -183,6 +184,44 @@ func TestCachePutWrongSize(t *testing.T) {
...
+
+	found, _ = testCache.Contains(cache.CAS, contentsHash, correctSize+1)
+	if found {
+		t.Error("Expected not found, due to size being different")
+	}
+
+	_, _, err = testCache.Get(cache.CAS, contentsHash, correctSize+1)
+	if err == nil {
+		t.Error("Expected error, due to size being different")
+	}
```
mostynb, line 212: I think this should just be a cache miss, no error.

cache/s3proxy/s3proxy.go
```diff
@@ -186,7 +186,6 @@ func (c *s3Cache) Get(kind cache.EntryKind, hash string) (io.ReadCloser, int64,
 
 func (c *s3Cache) Contains(kind cache.EntryKind, hash string) (bool, int64) {
 	size := int64(-1)
-
```
mostynb, line 189: nitpick: please remove this stray change

server/grpc_ac.go
```diff
@@ -44,10 +44,14 @@ func (s *grpcServer) GetActionResult(ctx context.Context,
 		return nil, err
 	}
 
+	// Clients provides hash and size of the Action, but not size of the ActionResult
+	// checked by the the disk cache.
+	const sizeActionResult = -1
```
mostynb, line 49: Maybe rename this to `unknownActionResultSize`?

cache/disk/disk.go
```diff
@@ -413,6 +421,10 @@ func (c *Cache) Get(kind cache.EntryKind, hash string) (io.ReadCloser, int64, er
 			len(hash), sha256.Size)
 	}
 
+	if kind == cache.CAS && size <= 0 && hash == emptySha256 {
+		return ioutil.NopCloser(bytes.NewReader([]byte{})), 0, nil
```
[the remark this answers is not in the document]
ulrfa: Done

cache/disk/disk.go
```diff
@@ -426,8 +438,14 @@ func (c *Cache) Get(kind cache.EntryKind, hash string) (io.ReadCloser, int64, er
 			var f *os.File
 			f, err = os.Open(blobPath)
 			if err == nil {
+				foundSize := fileInfo.Size()
+				err := c.checkSize(kind, hash, size, foundSize)
```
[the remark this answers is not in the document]
ulrfa: Done

cache/disk/disk.go
```diff
@@ -577,6 +610,14 @@ func (c *Cache) Stats() (currentSize int64, numItems int) {
 	return c.lru.CurrentSize(), c.lru.Len()
 }
 
+func (c *Cache) checkSize(kind cache.EntryKind, hash string, requestedSize int64, foundSize int64) error {
```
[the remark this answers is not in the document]
ulrfa: Done

cache/disk/disk.go
```diff
@@ -577,6 +610,14 @@ func (c *Cache) Stats() (currentSize int64, numItems int) {
...
+	if requestedSize > -1 && foundSize > -1 && requestedSize != foundSize {
+		return fmt.Errorf("Invalid size found: %d for %s %s, requested size: %d",
+			foundSize, kind, hash, requestedSize)
```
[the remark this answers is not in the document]
ulrfa: Done

cache/disk/disk.go
```diff
@@ -529,37 +550,49 @@ func (c *Cache) Get(kind cache.EntryKind, hash string) (io.ReadCloser, int64, er
...
 		item := val.(*lruItem)
-		foundLocally = item.committed
-		size = item.size
+		found = item.committed
+		foundSize = item.size
 	}
 	c.mu.Unlock()
 
-	if foundLocally {
-		return true, size
```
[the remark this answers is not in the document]
ulrfa: The reason was to avoid duplicating the size check in both the local and proxy branches. But perhaps your suggestion is more readable and I updated according to that.

cache/disk/disk_test.go
ulrfa: Done

cache/s3proxy/s3proxy.go
ulrfa: Done

server/grpc_ac.go
ulrfa: Done

ulrfa:
Thanks for the review! Something more?

cache/disk/disk_test.go
```diff
@@ -185,43 +185,88 @@ func TestCachePutWrongSize(t *testing.T) {
...
 
+func TestCacheGetContainsWrongSizeWithProxy(t *testing.T) {
+
+	cacheDir := tempDir(t)
+	defer os.RemoveAll(cacheDir)
+	testCache := New(cacheDir, 100, new(proxyStub))
+
+	var found bool
+	var rdr io.ReadCloser
+
```
mostynb, line 230: Please add a comment here to point out that proxyStub contains the the digest {contentsHash, contentsLength}.

ulrfa: Done

ulrfa:
Thanks! I have now pushed a squashed commit, including the comment.

mostynb:
Thanks- landed.

PR #267, opened 2020-05-07
Read After Write Guarantee

ulrfa:
Four related issues/scenarios addressed with this commit:

1. Client A uploads file to CAS. It becomes commited in disk storage. Later client B uploads same file with same hash overwriting the one from client A. **During second upload the disk storage changes the entry into uncommited, resulting in that file that was previously available can not be retrieved** for a while. This behviour is problematic for Bazel Remote Execution and the bazel feature builds-without-the-bytes, which expect read after write guarantee.
2. If the same file is uploaded concurrently (either by the same client or different clients) the **second one is silently discarded and the upload is acknowledge even before the upload has been completed** by the first client. This behviour is problematic for Bazel Remote Execution and the bazel feature builds-without-the bytes, which expect read after write guarantee.
3. A deferred os.Remove(tmpFilePath) after an upload, could get delayed and interfer with another upload of the same key.
4. After Put adds an uncommited entry, but before it creates a temporary file, it can become evicted by another concurrent add. Then the download would continue and could result in a file stored in CAS without entry in LRU, and it could also interfer with following concurrent uploads of the same key.

All these issues are addressed by this commit, without introducing risk for a fast uploading clients having to wait for a slower uploadling clients that are concurrently uploading the same file. Instead it adds support for concurrent uploads of the same key.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -52,6 +52,8 @@ type Config struct {
 	IdleTimeout             time.Duration             `yaml:"idle_timeout"`
 	DisableHTTPACValidation bool                      `yaml:"disable_http_ac_validation"`
 	DisableGRPCACDepsCheck  bool                      `yaml:"disable_grpc_ac_deps_check"`
+        ReadAfterWriteGuarantee bool                      `yaml:"read_after_write_guarantee"`
+        OverwriteCommited       bool                      `yaml:"overwrite_commited"`
 }
 
 // New returns a validated Config with the specified values, and an error
@@ -78,6 +80,8 @@ func New(dir string, maxSize int, host string, port int, grpcPort int,
 		IdleTimeout:             idleTimeout,
 		DisableHTTPACValidation: disableHTTPACValidation,
 		DisableGRPCACDepsCheck:  disableGRPCACDepsCheck,
+                ReadAfterWriteGuarantee: false,
+                OverwriteCommited:       true,
 	}
 
 	err := validateConfig(&c)
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -31,6 +31,26 @@ var (
 		Name: "bazel_remote_disk_cache_misses",
 		Help: "The total number of disk backend cache misses",
 	})
+	evictedUncommited = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_evicted_uncommited",
+		Help: "The total number of uncommited entries evicted from disk backend cache.",
+	})
+	newAcceptedWhenExistingCommited = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_new_accepted_commited",
+		Help: "The total number of new disk backend put requests accepted when an already existing commited entry",
+	})
+	newDiscardedWhenExistingCommited = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_new_discarded_commited",
+		Help: "The total number of new disk backend put requests discarded when an already existing commited entry",
+	})
+	newAcceptedWhenExistingUncommited = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_new_accepted_uncommited",
+		Help: "The total number of new disk backend put requests accepted when an already existing uncommited entry",
+	})
+	newDiscardedWhenExistingUncommited = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_new_discarded_uncommited",
+		Help: "The total number of new disk backend put requests discarded when an already existing uncommited entry",
+	})
 )
 
 // lruItem is the type of the values stored in SizedLRU to keep track of items.
@@ -51,6 +71,8 @@ type Cache struct {
 
 	mu  *sync.Mutex
 	lru SizedLRU
+        readAfterWriteGuarantee bool
+        overwriteCommited bool
 }
 
 type nameAndInfo struct {
@@ -63,7 +85,7 @@ const sha256HashStrSize = sha256.Size * 2 // Two hex characters per byte.
 // New returns a new instance of a filesystem-based cache rooted at `dir`,
 // with a maximum size of `maxSizeBytes` bytes and an optional backend `proxy`.
 // Cache is safe for concurrent use.
-func New(dir string, maxSizeBytes int64, proxy cache.Proxy) *Cache {
+func New(dir string, maxSizeBytes int64, proxy cache.Proxy, readAfterWriteGuarantee bool, overwriteCommited bool) *Cache {
 	// Create the directory structure.
 	hexLetters := []byte("0123456789abcdef")
 	for _, c1 := range hexLetters {
@@ -101,47 +123,24 @@ func New(dir string, maxSizeBytes int64, proxy cache.Proxy) *Cache {
 			return
 		}
 
-		// There is an ongoing upload for the evicted item. The temp
-		// file may or may not exist at this point.
+		// If there are concurrent uploads ongoing for a key that is
+		// evicted, then those uploads are allowed to continue and will
+		// re-add themself when finished.
 		//
-		// We should either be able to remove both the temp file and
-		// the regular cache file, or to remove just the regular cache
-		// file. The temp file is renamed/moved to the regular cache
-		// file without holding the lock, so we must try removing the
-		// temp file first.
-
+		// Trying to delete those files in uploading state would be
+		// problematic since the eviction might occur when they
+		// released the lock after they added as uncommited in LRU,
+		// but before they created the temporary file, resulting in
+		// that they might continue creating both their temporary file
+		// and the real file after onEvict already finished.
+		//
 		// Note: if you hit this case, then your cache size might be
 		// too small (blobs are moved to the most-recently used end
 		// of the index when the upload begins, and these items are
 		// still uploading when they reach the least-recently used
 		// end of the index).
 
-		tf := f + ".tmp"
-		var fErr, tfErr error
-		removedCount := 0
-
-		tfErr = os.Remove(tf)
-		if tfErr == nil {
-			removedCount++
-		}
-
-		fErr = os.Remove(f)
-		if fErr == nil {
-			removedCount++
-		}
-
-		// We expect to have removed at least one file at this point.
-		if removedCount == 0 {
-			if !os.IsNotExist(tfErr) {
-				log.Printf("ERROR: failed to remove evicted item: %s / %v",
-					tf, tfErr)
-			}
-
-			if !os.IsNotExist(fErr) {
-				log.Printf("ERROR: failed to remove evicted item: %s / %v",
-					f, fErr)
-			}
-		}
+		evictedUncommited.Inc()
 	}
 
 	c := &Cache{
@@ -149,8 +148,12 @@ func New(dir string, maxSizeBytes int64, proxy cache.Proxy) *Cache {
 		proxy: proxy,
 		mu:    &sync.Mutex{},
 		lru:   NewSizedLRU(maxSizeBytes, onEvict),
+		readAfterWriteGuarantee: readAfterWriteGuarantee,
+		overwriteCommited: overwriteCommited,
 	}
 
+	log.Printf("Read after write guarantee: %t\n", c.readAfterWriteGuarantee)
+	log.Printf("Overwrite commited: %t\n", c.overwriteCommited)
 	err := c.migrateDirectories()
 	if err != nil {
 		log.Fatalf("Attempting to migrate the old directory structure to the new structure failed "+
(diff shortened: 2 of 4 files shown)


ulrfa:
What do you think about the issues addressed by this commit?

Would you accept this pull request if I add and update unit tests?

Should the behavour be configurable?

mostynb:
Thanks for the PR. This is complicated code, so it's great to have another pair of eyes on it.

I will try to take a proper look early next week, but here are some initial comments...

> Four related issues/scenarios addressed with this commit:
> 
> 1. Client A uploads file to CAS. It becomes commited in disk storage. Later client B uploads same file with same hash overwriting the one from client A. **During second upload the disk storage changes the entry into uncommited, resulting in that file that was previously available can not be retrieved** for a while. This behviour is problematic for Bazel Remote Execution and the bazel feature builds-without-the-bytes, which expect read after write guarantee.

I have a large refactoring work-in-progress that might avoid this problem in another way (#186) by removing the need to overwrite CAS blobs. I need to find time to finish it off.

> 2. If the same file is uploaded concurrently (either by the same client or different clients) the **second one is silently discarded and the upload is acknowledge even before the upload has been completed** by the first client. This behviour is problematic for Bazel Remote Execution and the bazel feature builds-without-the bytes, which expect read after write guarantee.

This is a known issue, I guess it was less likely to occur in practice before builds-without-the-bytes. Have you encountered it in practice?

> 3. A deferred os.Remove(tmpFilePath) after an upload, could get delayed and interfer with another upload of the same key.

This might be a simple fix in the current setup- should unlock the mutex after removing the temp file here:
https://github.com/buchgr/bazel-remote/blob/69bce621e7ab073e92c7ec479e0708eb315d5b6a/cache/disk/disk.go#L470

And maybe also merge this deferred function with the one a few lines above, which uses the mutex:
https://github.com/buchgr/bazel-remote/blob/69bce621e7ab073e92c7ec479e0708eb315d5b6a/cache/disk/disk.go#L319

> 4. After Put adds an uncommited entry, but before it creates a temporary file, it can become evicted by another concurrent add. Then the download would continue and could result in a file stored in CAS without entry in LRU, and it could also interfer with following concurrent uploads of the same key.

If you hit that case, it is probably a sign that your cache size is simply too small. I'm not sure if we need to "gracefully" handle this situation, since the real solution is to increase the cache size, which requires a restart.

> All these issues are addressed by this commit, without introducing risk for a fast uploading clients having to wait for a slower uploadling clients that are concurrently uploading the same file. Instead it adds support for concurrent uploads of the same key.

ulrfa:
Thanks for your instant response!

My scenario is bazel with -j 1000, using bazel-remote as CAS/AC for remote execution with Buildbarn. We use bazel-remote instead of Buildbarn’s bb-storage, due to higher troughput and lower cpu consumption.

I experienced failed builds, due to files missing in CAS, and tried to find potential causes, which resulted in https://github.com/buchgr/bazel-remote/pull/267 and https://github.com/buchgr/bazel-remote/pull/266. I have tested extensively with both those patches combined and after that never seen missing CAS entries.

I can’t say for sure which of the individual changes are most important. But I’ve seen the bazel client uploading the same file concurrently many times. So I suspect handling concurrent uncommited uploads of the same file is critical. And if some of the duplicated uploads gets delayed, it could reach bazel-remote also in commited state. Therefore handling is cricitcal both for uncommited and commited.

> This might be a simple fix in the current setup- should unlock the mutex after removing the temp file

Perhaps the temp file is removed outside the lock, in order to minimize file operations while holding the lock? For better parallellism?

I think the non-unique temporary file name is the root of this problem, and much of the other complexity. It the temporary file name would be unique, then it could be handled safely outside the lock and also using defer.

> If you hit that case, it is probably a sign that your cache size is simply too small. I'm not sure if we need to "gracefully" handle this situation, since the real solution is to increase the cache size, which requires a restart.

Yes, the cache size should be increased in that case, and failed builds and restart could be OK. But I’m a bit worried about if it could result in corrupt files in the cache. I have to think more about that.

/Ulrik

config/config.go
```diff
@@ -52,6 +52,8 @@ type Config struct {
 	IdleTimeout             time.Duration             `yaml:"idle_timeout"`
 	DisableHTTPACValidation bool                      `yaml:"disable_http_ac_validation"`
 	DisableGRPCACDepsCheck  bool                      `yaml:"disable_grpc_ac_deps_check"`
+        ReadAfterWriteGuarantee bool                      `yaml:"read_after_write_guarantee"`
+        OverwriteCommited       bool                      `yaml:"overwrite_commited"`
```
mostynb, line 56: I think I would rather keep things simple and have only one behaviour, ie allow concurrent writes.

ulrfa: Agree. Then ReadAfterWriteGuarantee configuration can be removed.

**What about OverwriteCommited?** Personally I don’t have a use case for replacing committed entries, because:

1. Either the cache is huge and contains so many artifacts from so many different builds, that it becomes too hard to know which entries are poisoned and should be overwritten. If I discovered that such a cache contained poisened entries, I would clear the whole cache anyway.
2. Or the cache is tiny and could be re-populated quickly, and then it would be OK to clear the whole cache.

mostynb: The only two reasons I can think why we would want to overwrite blobs are:
1) If the blob is somehow known to be corrupt, then we can make a client re-upload everything to replace the bad items.
2) For simplicity. We let the client upload the blob anyway, and saving it to disk is takes much less time than the upload so why not do it anyway, without bothering to check if the preexisting blob was corrupt?

mostynb:
> I can’t say for sure which of the individual changes are most important. But I’ve seen the bazel client uploading the same file concurrently many times. So I suspect handling concurrent uncommited uploads of the same file is critical. And if some of the duplicated uploads gets delayed, it could reach bazel-remote also in commited state. Therefore handling is cricitcal both for uncommited and commited.

This can be time-consuming, since you probably need to start with an empty cache for such a stress-test. But if you are able to, could you check what kind of blobs were causing trouble? I suspect the most likely candidates are the empty blob, or large blobs (which take some time to upload).

> > This might be a simple fix in the current setup- should unlock the mutex after removing the temp file
> 
> Perhaps the temp file is removed outside the lock, in order to minimize file operations while holding the lock? For better parallellism?

Yes, IIRC that's what I was aiming for. But there's a potential race condition, as you say.

ulrfa:
I think you are right about most likely candidates are empty blob or large blobs. And now with the special handling for empty blob, the empty blob should not trigger this race condition.

Without #267, I have seen it once also for a non-empty blob. It was a 59 MB large file (a tool) that a lot of concurrent actions had as input file:

`Remote connection/protocol failed with: execution failed FAILED_PRECONDITION: Failed to obtain input file "external/xxx": GRPC BYTESTREAM READ BLOB NOT FOUND: 7f580eb4ec42890f924420f552a828f4362d83e3e4104587cdea71d0472c2c51
`
(”xxx” is obfuscated)

I have now tried to reproduce it by clearing the caches between each build, but I have not seen it again. However I suspect the issue is still there. Because sometimes up to 25% of all the PUT operations reaching bazel-remote, are for keys already in commited state in the LRU, and with unfortunate scheduling that could trigger this issue.

Example with many overwritten keys in commited state. All of the numbers below are a single build, after clearing the caches, using Buildbarn with bazel-remote:
 
bazel -j 1000 results in:
- PUT with the key not in LRU: 50 429
- PUT with the key in LRU in commited state:  varying between 10 000 - 16 000.

bazel -j 250 results in:
- PUT with the key not in LRU: 50 429
- PUT with the key in LRU in commited state:  varying between 4 000 - 8 000.

16 000 / (50 429 + 16 000) = ~25%

config/config.go
ulrfa: Good point about simplicity! And I have not noticed any obvious difference in performance. So the conclusion is to remove both configuration parameters and only keep the code for the cases ReadAfterWriteGuarantee=true and OverwriteCommited=true.

ulrfa:
Do you have more comments @mostynb ? Or should I do the changes discussed here and add/update unit tests?

config/config.go
mostynb: Yes- let's not introduce configuration options, we just need to be careful to use atomic move/rename operations to replace existing files.

mostynb:
OK, how about we try this:
* Allow concurrent uploads, using a unique temp file inside the same dir as the eventual committed file.
* Attempt to keep the cache size accounting correct without stat'ing a file while holding the cache lock. Maybe we can add an ongoing uploads counter to the LRU items?

BTW in your testcase, is that a single bazel client? If so, the number of re-uploads looks unusually high. Maybe this behaviour is coming from buildbarn? Are the buildbarn workers uploading all their inputs to the CAS without checking if they're already there?

ulrfa:
> Attempt to keep the cache size accounting correct without stat'ing a file while holding the cache lock. Maybe we can add an ongoing uploads counter to the LRU items?

When there is a race between concurrent uploaders, then we don’t know in which order they rename their temporary files to the common commited destination. The first ones are overwritten and the last one wins. I can see two alternatives for keeping the cache size accounting correct:

A) Hold the cache lock during rename, **always**.

B) Stat the resulting file while holding the cache lock, **but only needed when there has been another concurrent uploader**.

How would an upload counter avoid holding the lock while doing a file operation?

> BTW in your testcase, is that a single bazel client? If so, the number of re-uploads looks unusually high. Maybe this behaviour is coming from buildbarn? Are the buildbarn workers uploading all their inputs to the CAS without checking if they're already there?

Yes, a single bazel client. But with -j 1000 on 72 core machine, with 396 GB RAM. I think buildbarn workers are uploading their output to CAS without checking if they are already there, but it can not explain this, because I now also tried buildbarn with no connected workers. Then bazel client uploads input files for the first parallel 1000 actions, but buildbarn won't process them nor upload any file. But I still got:

- PUT with the key not in LRU: 11 253
- PUT with the key in LRU in committed state: 2 944

2944 / (2944 + 11 253) = ~20%

mostynb:
> > Attempt to keep the cache size accounting correct without stat'ing a file while holding the cache lock. Maybe we can add an ongoing uploads counter to the LRU items?
> 
> When there is a race between concurrent uploaders, then we don’t know in which order they rename their temporary files to the common commited destination. The first ones are overwritten and the last one wins. I can see two alternatives for keeping the cache size accounting correct:
> 
> A) Hold the cache lock during rename, **always**.
> 
> B) Stat the resulting file while holding the cache lock, **but only needed when there has been another concurrent uploader**.
> 
> How would an upload counter avoid holding the lock while doing a file operation?

I think "last upload to finish wins" is OK, the server can't really promise anything more.

I was thinking of trying something like this: initialize the item's counter to 0 on the first upload, and increment it for each concurrent upload. Each incoming upload increases the cache size counter by the blob size. When an upload finishes:
1) Lock the mutex.
2) (Optional?) move the item to the front of the LRU.
3) If the item's counter is greater than 0, subtract the blob size from the cache size count and decrement the item's counter (no stat required).
4) Do an atomic rename to the final destination.
5) Release the lock.

We increase the size of the index a little, since every item has this counter value even if there are no concurrent uploads, but the code is simple.

I have an idea for a way to avoid this counter, at the cost of some complexity and some temporary miscounting of the cache size (temporary as long as each upload eventually ends), but I think we should attempt the simpler solution first.

> > BTW in your testcase, is that a single bazel client? If so, the number of re-uploads looks unusually high. Maybe this behaviour is coming from buildbarn? Are the buildbarn workers uploading all their inputs to the CAS without checking if they're already there?
> 
> Yes, a single bazel client. But with -j 1000 on 72 core machine, with 396 GB RAM. I think buildbarn workers are uploading their output to CAS without checking if they are already there, but it can not explain this, because I now also tried buildbarn with no connected workers. Then bazel client uploads input files for the first parallel 1000 actions, but buildbarn won't process them nor upload any file. But I still got:
> 
> * PUT with the key not in LRU: 11 253
> * PUT with the key in LRU in committed state: 2 944
> 
> 2944 / (2944 + 11 253) = ~20%

Interesting. Technically it is OK for the client to do this, just undesirable. But that percentage is higher than I would expect (especially given buchgr's comment about this being rare).

Still, it's good to make bazel-remote more robust.

mostynb:
> I have an idea for a way to avoid this counter, at the cost of some complexity and some temporary miscounting of the cache size (temporary as long as each upload eventually ends), but I think we should attempt the simpler solution first.

I thought this through a bit and realised we can avoid the counter by simply using the committed bool that we already have. When an upload finishes, if committed is true then we know we're replacing a blob, and if it is false then we know this is the first upload to add the blob, and it should be easy to adjust the cache size count accordingly.

On a related note, we might want to maintain a separate uncommitted size count, and include that when deciding if we need to purge items.

ulrfa:
> Interesting. Technically it is OK for the client to do this, just undesirable. But that percentage is higher than I would expect (especially given buchgr's comment about this being rare).

Perhaps buchgr’s comment were in the context of remote caching, and not remote execution. But I hope it is rare even for remote execution, except when I explicitly clear the cache to trigger it. ;-)

ulrfa:
I view this as two separate questions:

1: Should we minimize holding locks during os.Rename operations of temporary file?
2: What data structure for uncommitted entries?

For question 1, I see the three alternatives:

A) Always hold the global disk cache lock during os.Rename.
B) Introduce new fine granular locking per hash, to hold during os.Rename.
C) Don’t lock during os.Rename, instead **occasionally** stat after os.Rename, before adding as commited to LRU

Your posted pseudo code is using A. What do you think about high lock contention for that alternative? I guess os.Rename takes significant time and involves context switch to kernel space. A single writer holding the lock for a longer time, would reduce the parallelism for many readers.

I’m lean towards C, since the stat would only be required when multiple concurrent writers of same hash has been detected.

For question 2, I wonder if it is motivated to have uncommitted entries in the LRU’s list at all? And instead only count their sizes in LRU’s already existing currentSize field.

mostynb:
> I view this as two separate questions:
> 
> 1: Should we minimize holding locks during os.Rename operations of temporary file?
> 2: What data structure for uncommitted entries?
[snip]
> For question 2, I wonder if it is motivated to have uncommitted entries in the LRU’s list at all? And instead only count their sizes in LRU’s already existing currentSize field.

Let's take question 2 first...

I suspect that the uncommitted bool was originally added as a way to recognise concurrent uploads, so they could be easily ignored. Since we're going to change that behaviour, I think it would be a good idea to drop that field and only have committed entries in the LRU index.

> For question 1, I see the three alternatives:
> 
> A) Always hold the global disk cache lock during os.Rename.
> B) Introduce new fine granular locking per hash, to hold during os.Rename.
> C) Don’t lock during os.Rename, instead **occasionally** stat after os.Rename, before adding as commited to LRU
> 
> Your posted pseudo code is using A. What do you think about high lock contention for that alternative? I guess os.Rename takes significant time and involves context switch to kernel space. A single writer holding the lock for a longer time, would reduce the parallelism for many readers.
> 
> I’m lean towards C, since the stat would only be required when multiple concurrent writers of same hash has been detected.

(A) might be slow, if os.Rename is slow. We should ensure that the source and destination are on the same filesystem, which is typically a fast path, but I don't know precisely how fast.

(B) sounds complicated. I would prefer to skip this until we see that simpler options fail.

(C) sounds slow, given that your numbers show this happening ~20% of the time, and there are performance issues with os.Stat on mac.

So let's consider another option D: perform os.Rename without holding the lock, in such a way that we have eventual consistency in cache size accounting on the server side (assuming that all uploads eventually finish). How about this:

1. Lock the mutex.
2. Add the blob size to the cache size (but do not add the item to the LRU index yet).
3. Unlock the mutex.
4. Write the blob to a temp file in the same directory as the final file.
5. Rename the temp file to the final file. This is atomic unless there's a crash, according to posix IIRC. If the new blob is the same as the old blob or if there was no old blob, then clients will see no difference- this should be the vast majority of cases.

When the function returns, run a deferred function:
1. Lock the mutex.
2. If the upload failed (the temp file should have been removed), subtract the expected blob size from the cache size.
3. Else If the blob is already in the index, subtract the blob size from the cache size.
4. Unlock the mutex.

With this setup, clients who download a blob can only ever receive a complete blob. The blob should only ever change if there's a cache collision, and clients can choose to mitigate that possibility by only using a good hash function, which should be fine for SHA256, and if we later add support for weaker hashes that is a risk that clients can choose to take or not.

We would also need to modify Get when there's a proxy backend. If the client provides the size of the blob, then we can use a similar algorithm when downloading from the proxy. If the blob size is unknown then I guess we need a special case that only adds the blob size at commit-time (and subtracts the previous blob size if it was already in the index).

mostynb:
Here's my work-in-progress fix: https://github.com/mostynb/bazel-remote/commits/concurrent_inserts

PR #271, opened 2020-05-10
Add option for HTTP read/write timeouts.

elianuber:
Add option flags for server read/write timeouts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,8 @@ GLOBAL OPTIONS:
    --grpc_port value             The port the EXPERIMENTAL gRPC server listens on. Set to 0 to disable. (default: 9092) [$BAZEL_REMOTE_GRPC_PORT]
    --profile_host value          A host address to listen on for profiling, if enabled by a valid --profile_port setting. (default: "127.0.0.1") [$BAZEL_REMOTE_PROFILE_HOST]
    --profile_port value          If a positive integer, serve /debug/pprof/* URLs from http://profile_host:profile_port. (default: 0, ie profiling disabled) [$BAZEL_REMOTE_PROFILE_PORT]
-   --http_read_timeout value     The HTTP read timeout for a client request in seconds. (applies to the frontend not the proxy backend) (default: 0s, ie disabled) [$BAZEL_REMOTE_HTTP_READ_TIMEOUT]
-   --http_write_timeout value    The HTTP write timeout for a server response in seconds. (applies to the frontend not the proxy backend) (default: 0s, ie disabled) [$BAZEL_REMOTE_HTTP_WRITE_TIMEOUT]
+   --http_read_timeout value     The HTTP read timeout for a client request in seconds (does not apply to the proxy backends or the profiling endpoint) [$BAZEL_REMOTE_HTTP_READ_TIMEOUT]
+   --http_write_timeout value    The HTTP write timeout for a server response in seconds (does not apply to the proxy backends or the profiling endpoint) [$BAZEL_REMOTE_HTTP_WRITE_TIMEOUT]
    --htpasswd_file value         Path to a .htpasswd file. This flag is optional. Please read https://httpd.apache.org/docs/2.4/programs/htpasswd.html. [$BAZEL_REMOTE_HTPASSWD_FILE]
    --tls_enabled                 This flag has been deprecated. Specify tls_cert_file and tls_key_file instead. (default: false) [$BAZEL_REMOTE_TLS_ENABLED]
    --tls_cert_file value         Path to a pem encoded certificate file. [$BAZEL_REMOTE_TLS_CERT_FILE]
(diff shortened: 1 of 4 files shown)


config/config_test.go
```diff
@@ -19,6 +19,8 @@ tls_cert_file: /opt/tls.cert
 tls_key_file:  /opt/tls.key
 disable_http_ac_validation: true
 enable_endpoint_metrics: true
+read_timeout: 3
+write_timeout: 4
```
mostynb, line 23: These example values seem fairly low (at least compared to the current "no limit"). It might be worth adding a brief description here about how admins might choose suitable values.

README.md
```diff
@@ -103,6 +103,8 @@ GLOBAL OPTIONS:
    --grpc_port value             The port the EXPERIMENTAL gRPC server listens on. Set to 0 to disable. (default: 9092) [$BAZEL_REMOTE_GRPC_PORT]
    --profile_host value          A host address to listen on for profiling, if enabled by a valid --profile_port setting. (default: "127.0.0.1") [$BAZEL_REMOTE_PROFILE_HOST]
    --profile_port value          If a positive integer, serve /debug/pprof/* URLs from http://profile_host:profile_port. (default: 0, ie profiling disabled) [$BAZEL_REMOTE_PROFILE_PORT]
+   --read_timeout value          The HTTP read timeout for a client request in seconds. (default: 0s, ie disabled) [$BAZEL_REMOTE_READ_TIMEOUT]
+   --write_timeout value         The HTTP write timeout for a server response in seconds. (default: 0s, ie disabled) [$BAZEL_REMOTE_WRITE_TIMEOUT]
```
mostynb, line 107: It might be worth naming these flags in such a way that we can add other timeout values in the future (eg for proxy backends, and/or the profiling server) and still have consistent naming. And maybe explicitly mention in the description that these flags don't cover proxy backends.

main.go
```diff
@@ -110,6 +111,20 @@ func main() {
 			DefaultText: "0, ie profiling disabled",
 			EnvVars:     []string{"BAZEL_REMOTE_PROFILE_PORT"},
 		},
+		&cli.IntFlag{
```
mostynb, line 114: Should we use `DurationFlag`'s for these, like we do for `--idle_timeout`?

config/config_test.go
elianuber: Setting to some real world values :+1:

README.md
elianuber: Agreed, adding more description and fixing names.

main.go
elianuber: Good point, using Duration now.

README.md
```diff
@@ -139,6 +141,16 @@ host: localhost
...
 #profile_port: 7070
 
+# If specified, the HTTP read timeout for a client request in seconds.
+# (applies to the frontend not the proxy backend)
+# Time units can be one of: "s", "m", "h".
+#http_read_timeout: 5s
+
+# If specified, the HTTP write timeout for a server response in seconds.
+# (applies to the frontend not the proxy backend)
+# Time units can be one of: "s", "m", "h".
```
mostynb, line 151: I think we should skip this mostly duplicated text, and refer to "HTTP read/write timeouts" in the first sentence, with no newline between the two example lines. And reuse the same sentence in brackets as suggested for the flag text.

main.go
```diff
@@ -110,6 +110,20 @@ func main() {
 			DefaultText: "0, ie profiling disabled",
 			EnvVars:     []string{"BAZEL_REMOTE_PROFILE_PORT"},
 		},
+		&cli.DurationFlag{
+			Name:        "http_read_timeout",
+			Value:       0,
+			Usage:       "The HTTP read timeout for a client request in seconds. (applies to the frontend not the proxy backend)",
```
mostynb, line 116: How about this rewording:
`The HTTP write timeout for a server response in seconds (does not apply to the proxy backends or the profiling endpoint).`

README.md
```diff
@@ -139,6 +141,16 @@ host: localhost
 #profile_host: 127.0.0.1
 #profile_port: 7070
 
+# If specified, the HTTP read timeout for a client request in seconds.
+# (applies to the frontend not the proxy backend)
+# Time units can be one of: "s", "m", "h".
+#http_read_timeout: 5s
```
mostynb, line 147: How did you arrive at these example values (5 and 10s)? Is this an upper limit on the expected time for the server to receive an upload, and to write a response back?

main.go
elianuber: sounds good!

README.md
elianuber: fixed !

README.md
elianuber: These limits are just examples, It varies depending on blob sizes, etc. there is no one-fit-for-all :/

mostynb: I was thinking we could recommend something like double the maximum time expected for clients to upload/download items from the cache- ie a rough rule that can be used as a starting point. What do you think?

elianuber: sounds good ! @mostynb can you merge and update the docs then ?

mostynb: Sure.

mostynb:
Squashed, rebased and landed in #287. I will tweak the documentation tomorrow. Thanks!

Issue #302, opened 2020-07-02
Run as non-root user


mmikitka:
I want to run the bazel-remote container in a Kubernetes cluster constrained by a mandatory non-root Pod Security Policy rule. At present, bazel-remote is invoked as the root user.

A forthcoming PR will:
* Create the "bazelremote" user
* Create the "/data" directory
* Run the container as "bazelremote"

mmikitka:
Here is the PR: https://github.com/buchgr/bazel-remote/pull/303

mostynb:
Landed #303.

PR #303, opened 2020-07-02
Run as non-root user

mmikitka:
For issue https://github.com/buchgr/bazel-remote/issues/302

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -3,6 +3,7 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_image", "cont
 load("@io_bazel_rules_docker//contrib:passwd.bzl", "passwd_entry", "passwd_tar")
 load("@io_bazel_rules_docker//go:image.bzl", "go_image")
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+load("@rules_pkg//:pkg.bzl", "pkg_tar")
 
 gazelle(
     name = "gazelle",
@@ -44,16 +45,16 @@ go_binary(
     x_defs = {"main.gitCommit": "{STABLE_GIT_COMMIT}"},
 )
 
-# Create the non-root user, "bazelremote", so the container can run with reduced
+# Create the non-root user, "bazel-remote", so the container can run with reduced
 # privileges.
-BAZEL_REMOTE_USER_NAME = "bazelremote"
+BAZEL_REMOTE_USER_NAME = "bazel-remote"
 BAZEL_REMOTE_USER_ID = 1000
 passwd_entry(
     name = "bazelremote_passwd_entry",
     username = BAZEL_REMOTE_USER_NAME,
-    info = BAZEL_REMOTE_USER_NAME,
     uid = BAZEL_REMOTE_USER_ID,
-    home = "/data",
+    gid = BAZEL_REMOTE_USER_ID,
+    home = "/home/" + BAZEL_REMOTE_USER_NAME,
     create_home = True,
 )
 
@@ -85,6 +86,24 @@ go_image(
     x_defs = {"main.gitCommit": "{STABLE_GIT_COMMIT}"},
 )
 
+# Initialize the default /data storage directory since the bazel-remote
+# user does not have sufficient permissions at the root (/) directory.
+# Since an empty directory cannot be packaged by the pkg_tar rule, create
+# the "EMPTY" file.
+genrule(
+    name = "empty",
+    srcs = [],
+    outs = ["EMPTY"],
+    cmd = "touch $@",
+)
+
+pkg_tar(
+    name = "data_tar",
+    srcs = [":empty"],
+    package_dir = "/data",
+    owner = "{0}.{0}".format(BAZEL_REMOTE_USER_ID),
+)
+
 container_image(
     name = "bazel-remote-image",
     base = ":bazel-remote-base",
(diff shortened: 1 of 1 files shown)


mmikitka:
I found an issue: The Kubernetes PSP is checking the UID and not name (string). Patch is coming.

mmikitka:
Fixed with the most recent commit. Upon approval, this is ready to merge.

BUILD.bazel
```diff
@@ -46,13 +46,13 @@ go_binary(
 
 # Create the non-root user, "bazelremote", so the container can run with reduced
 # privileges.
-BAZEL_REMOTE_USER = "bazelremote"
+BAZEL_REMOTE_USER_NAME = "bazelremote"
```
mostynb, line 49: nitpick: please change this to bazel-remote.

BUILD.bazel
```diff
@@ -46,13 +46,13 @@ go_binary(
...
+BAZEL_REMOTE_USER_ID = 1000
 passwd_entry(
     name = "bazelremote_passwd_entry",
-    username = BAZEL_REMOTE_USER,
-    info = BAZEL_REMOTE_USER,
-    uid = 1000,
-    gid = 1000,
+    username = BAZEL_REMOTE_USER_NAME,
+    info = BAZEL_REMOTE_USER_NAME,
```
mostynb, line 54: Can we can skip this and just use the default value for "info"?

BUILD.bazel
```diff
@@ -43,6 +44,25 @@ go_binary(
...
+# Create the non-root user, "bazelremote", so the container can run with reduced
+# privileges.
+BAZEL_REMOTE_USER = "bazelremote"
+passwd_entry(
+    name = "bazelremote_passwd_entry",
+    username = BAZEL_REMOTE_USER,
+    info = BAZEL_REMOTE_USER,
+    uid = 1000,
+    gid = 1000,
+    home = "/data",
```
mostynb, line 56: I don't think we should reuse the cache directory as the home dir- bazel-remote might choose to delete unrecognised files in the cache dir, and this could become confusing if we forget and add some dot files or similar (eg a ~/.netrc file for the remote asset api).

BUILD.bazel
mmikitka: Done

BUILD.bazel
mmikitka: Done

BUILD.bazel
mmikitka: Done via a genrule that creates `/data/EMPTY` in a pkg_tar tarball (see comment in code for details)

mmikitka:
@mostynb I believe that I addressed your concerns

BUILD.bazel
```diff
@@ -85,6 +86,24 @@ go_image(
...
+# Initialize the default /data storage directory since the bazel-remote
+# user does not have sufficient permissions at the root (/) directory.
+# Since an empty directory cannot be packaged by the pkg_tar rule, create
+# the "EMPTY" file.
+genrule(
+    name = "empty",
+    srcs = [],
+    outs = ["EMPTY"],
+    cmd = "touch $@",
+)
```
mostynb, line 98: Perhaps we can use this instead?
https://github.com/bazelbuild/rules_docker/issues/37#issuecomment-544652904

mmikitka: Using `empty_dirs` builds without errors, but the directories are owned by the root user, and not the `user` attribute (in this case "bazel-remote").

Here is the error upon running the built image when using `empty_dirs`:
```12:45:41 ⌂91% [mmikitka:~/src/bazel-remote] docker-nonroot(+4/-26) 8s ± docker run --rm -it bazel:bazel-remote-image
2020/07/03 16:45:49 bazel-remote built with go1.14.3 from git commit 63bcf072dd5f197e3ee539d4ec32062c9e583d74.
2020/07/03 16:45:49 Initial RLIMIT_NOFILE cur: 1048576 max: 1048576
2020/07/03 16:45:49 Setting RLIMIT_NOFILE cur: 1048576 max: 1048576
2020/07/03 16:45:49 mkdir /data/cas: permission denied
```

mostynb:
Could you please `bazel run //:gazelle -- fix`, and then squash this all down to a single commit?

mmikitka:
@mostynb I ran `bazel run //:gazelle -- fix`, but I do not understand why I need to squash the changes, and I don't want to loose history on the source branch. Why not use squash commit as the PR merge strategy? That would give you a squash commit and preserve source branch history.

mostynb:
> @mostynb I ran `bazel run //:gazelle -- fix`, but I do not understand why I need to squash the changes, and I don't want to loose history on the source branch. Why not use squash commit as the PR merge strategy? That would give you a squash commit and preserve source branch history.

I prefer "squash and rebase", which github doesn't seem to support. So I did this locally and manually landed the change.

Thanks again.

mmikitka:
@mostynb Thanks for merging.

What is the release schedule for official Docker images? I'm currently using a pre-built image, and would like to use an official image, if possible.

mostynb:
I have a little cleanup to apply (eg #305), then I will ask Jakob to upload a new image to dockerhub later in the week.

BUILD.bazel
```diff
@@ -65,6 +88,24 @@ go_image(
...
+
+pkg_tar(
+    name = "data_tar",
+    srcs = [":empty"],
+    package_dir = "/data",
```
c4urself, line 105: I'm getting a permission denied issue because I had my volume mounted at /cache instead of /data -- does this change affect the dir-path?

mostynb: > I'm getting a permission denied issue because I had my volume mounted at /cache instead of /data -- does this change affect the dir-path?

When you start the docker container you need to map an existing cache dir in the host system that the docker user has read/write/execute permissions on to `/data` inside the container, in your case something like `-v /cache:/data`.

PR #309, opened 2020-07-09
docker: simplify non-root user setup

mostynb:
It turns out that our base container image already has a "nonroot" user, with uid/gid 65532.

The previous setup worked more of less out of the box when running as uid 1000 (the default uid for the initial user on most linux setups). Since we changed this default we need to document how to specify the required user.

Followup to #302.


mostynb:
@mmikitka: does this still work for you?

mmikitka:
@mostynb Sorry for not catching that in my PR, but yes this works for me. Thanks for asking.

mostynb:
No problem, I didn't catch it either. That [dive](https://github.com/wagoodman/dive) tool is pretty nice.

PR #313, opened 2020-07-16
Fall back to default IAM endpoint for S3 proxy

lavignes:
[s3proxy.go](https://github.com/buchgr/bazel-remote/blob/cc9030667416ab63d89b9fbf543f0243292009b4/cache/s3proxy/s3proxy.go#L59) only supports fetching IAM credentials if `s3.iam_role_endpoint` is defined.

In the README it is suggested that one uses the value `http://169.254.170.2` when running bazel-remote on ECS.
Unfortunately, this would configure [minio to make requests](https://github.com/minio/minio-go/blob/a57fec8037ec86c0f84116a7ccec82666b71dcd5/pkg/credentials/iam_aws.go#L110) to `http://169.254.170.2` rather than the correct value of `http://169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`.

Where `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is a task-specific environment variable set by ECS when running the image.

As far as I know, without running the docker image with a shell there is no easy way to expand this environment variable into the stock docker hub image's `ENTRYPOINT`.

To get around this issue I've made a new image that can expand this variable before starting bazel-remote. But it would be better if bazel-remote could just fall-back to using minio's default credentials provider logic.

The easiest fix I can see would be to flip the logic today to fallback to minio's defaults if `s3.access_key_id` and `s3.secret_access_key` are not defined.


mostynb:
Thanks for the PR.

Switching the order of these seems fine. I guess we should update help text and README.md if the information about ECS is wrong?

Should we also try to read `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` from the environment? Can't that be passed in to docker somehow?

lavignes:
I think if you are running this in AWS you are best-off just not setting this variable. Letting minio handle finding the credentials is probably better than trying to do it in bazel-remote since this also now enables it to work on EKS as well as ECS. There are other endpoints and environment variables used to find credentials and it seems unnecessary to try and replicate that non-trivial logic in this project.

What do you think about just suggesting not specifying the option / leaving it blank when running on AWS?

main.go
```diff
@@ -196,13 +196,13 @@ func main() {
 		&cli.StringFlag{
 			Name:    "s3.iam_role_endpoint",
 			Value:   "",
-			Usage:   "Endpoint for using IAM security credentials, eg http://169.254.169.254 for EC2, http://169.254.170.2 for ECS.",
+			Usage:   "Endpoint for using IAM security credentials. By default it will look for credentials in the standard locations for the AWS platform being ran on.",
```
mostynb, line 199: nitpick: let's just say "for the AWS platform."

lavignes: Sure :)

mostynb:
Thanks!

PR #321, opened 2020-07-24
grpc bytestream: read bytes with EOF

lizan:
Reader.Read can return io.EOF while some data read, skipping them would result BulkTransferException in Bazel.

### DIFF
--- a/server/grpc_bytestream.go
+++ b/server/grpc_bytestream.go
@@ -129,6 +129,24 @@ func (s *grpcServer) Read(req *bytestream.ReadRequest,
 	var chunkResp bytestream.ReadResponse
 	for {
 		n, err := rdr.Read(buf)
+
+		if n > 0 {
+			if limitedSend {
+				if (sendLimitRemaining - int64(n)) < 0 {
+					msg := "GRPC BYTESTREAM READ FAILED: READ LIMIT EXCEEDED"
+					return status.Error(codes.OutOfRange, msg)
+				}
+				sendLimitRemaining -= int64(n)
+			}
+
+			chunkResp.Data = buf[:n]
+			sendErr := resp.Send(&chunkResp)
+			if sendErr != nil {
+				s.accessLogger.Printf("GRPC BYTESTREAM READ FAILED TO SEND RESPONSE: %s", sendErr)
+				return status.Error(codes.Unknown, sendErr.Error())
+			}
+		}
+
 		if err == io.EOF {
 			s.accessLogger.Printf("GRPC BYTESTREAM READ COMPLETED %s",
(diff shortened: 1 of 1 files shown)


server/grpc_bytestream.go
```diff
@@ -129,10 +129,9 @@ func (s *grpcServer) Read(req *bytestream.ReadRequest,
...
-		if err == io.EOF {
-			s.accessLogger.Printf("GRPC BYTESTREAM READ COMPLETED %s",
-				req.ResourceName)
-			return nil
+		eof := err == io.EOF
+		if eof {
+			err = nil
 		}
 
 		if err != nil {
```
mostynb, line 137: https://golang.org/pkg/io/#Reader says "Callers should always process the n > 0 bytes returned before considering the error err."

So how about we put an `if n > 0` block immediately after the `Read` call, with the `limitedSend` check and the `Send` call inside, and then check `if err == io.EOF { <success> }` and finally `if err != nil { <failure> }`

lizan: Done.

server/grpc_bytestream.go
```diff
@@ -129,26 +129,16 @@ func (s *grpcServer) Read(req *bytestream.ReadRequest,
...
+					return status.Error(codes.OutOfRange, msg)
+				}
+				sendLimitRemaining -= int64(n)
 			}
-			sendLimitRemaining -= int64(n)
-		}
 
-		if n > 0 {
 			chunkResp.Data = buf[:n]
 			err = resp.Send(&chunkResp)
```
mostynb, line 143: I think we need to use a different variable here, otherwise this will overwrite the value from `Read`.

lizan: ah yes, fixed.

server/grpc_bytestream.go
```diff
@@ -129,6 +129,24 @@ func (s *grpcServer) Read(req *bytestream.ReadRequest,
...
+					msg := "GRPC BYTESTREAM READ FAILED: READ LIMIT EXCEEDED"
+					return status.Error(codes.OutOfRange, msg)
+				}
+				sendLimitRemaining -= int64(n)
+			}
+
+			chunkResp.Data = buf[:n]
+			sendErr := resp.Send(&chunkResp)
+			if sendErr != nil {
+				s.accessLogger.Printf("GRPC BYTESTREAM READ FAILED TO SEND RESPONSE: %s", err)
```
mostynb, line 145: There's one last `err` here that needs to be changed to `sendErr`.

lizan: Fixed

mostynb:
Thanks!

Issue #318, opened 2020-07-23
Inconsistent result when CAS exists from proxy but not locally.


lizan:
I have CI cache backed by S3 and here is an HTTP cache log:

```
2020/07/22 23:54:39 S3 DOWNLOAD envoy-ci-build-cache-us-east-2 public-x64/ac/a559cd7910cbbe821525f48e6efef10964c4243aeff3b6747baffd1e2863c096 OK
2020/07/22 23:54:39 S3 CONTAINS envoy-ci-build-cache-us-east-2 public-x64/cas/d35d521228f07e30b28984179d3435236ace2eddaadaa1526aaea58e39a389ba OK
2020/07/22 23:54:39 S3 CONTAINS envoy-ci-build-cache-us-east-2 public-x64/cas/615324b6d4553bf401f3e20e1d1e7fbb02f99f4d6a672f069157871547b0b4cb OK
2020/07/22 23:54:39 S3 DOWNLOAD envoy-ci-build-cache-us-east-2 public-x64/cas/0e5166cbbaee2f9bcb621398ab8f0c75deb169766d88b9f7b4047f39e2dcef31 OK
2020/07/22 23:54:39 S3 CONTAINS envoy-ci-build-cache-us-east-2 public-x64/cas/f8e5351bad6c3c961fbc247790047c031508b98ae083f8d72483505c8c039931 OK
2020/07/22 23:54:39 S3 CONTAINS envoy-ci-build-cache-us-east-2 public-x64/cas/6a36cd36e694976de00dd7aa55a8731a7fa9e06cde1a855d6409530bcfbd4e40 OK
2020/07/22 23:54:39 S3 DOWNLOAD envoy-ci-build-cache-us-east-2 public-x64/cas/f5ea44e96bcbec7c85633f0ebaecb16cc374ed1b1e2fb29513b76d0f1896289e OK
2020/07/22 23:54:39 S3 CONTAINS envoy-ci-build-cache-us-east-2 public-x64/cas/f8e5351bad6c3c961fbc247790047c031508b98ae083f8d72483505c8c039931 OK
2020/07/22 23:54:39 S3 CONTAINS envoy-ci-build-cache-us-east-2 public-x64/cas/6a36cd36e694976de00dd7aa55a8731a7fa9e06cde1a855d6409530bcfbd4e40 OK
[...]

PR #323, opened 2020-07-27
allow concurrent Puts and proxied Gets

mostynb:
At the moment, bazel-remote only allows a single concurrent Put for a given blob. If another Put starts, while one is ongoing, we discard the second copy and return quickly, but the blob is not yet available. This can cause trouble for the second client, which assumes that the blob is available as soon as its Put finishes. The same problem also exists for Get calls when using a proxy backend.

I guess this was not a problem when bazel-remote was first written and only supported the http cache protocol, because bazel serializes those requests more than it does with the newer gRPC protocol. With gRPC, a single bazel client is able to hit this problem.

This change refactors `disk.Cache` to allow concurrent Puts and proxied Gets, by reserving space in the LRU index and using pseudo-random tempfiles, While I was at it I extracted some common code into a couple of functions and reuse them in both Put and Get.

Hopefully this fixes #267 and #318.

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -14,8 +14,11 @@ import (
 	"regexp"
 	"sort"
 	"sync"
+	"syscall"
 
 	"github.com/buchgr/bazel-remote/cache"
+	"github.com/buchgr/bazel-remote/utils/tempfile"
+
 	"github.com/djherbis/atime"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -35,11 +38,12 @@ var (
 	})
 )
 
+var tfc = tempfile.NewCreator()
+
 // lruItem is the type of the values stored in SizedLRU to keep track of items.
 // It implements the sizedItem interface.
 type lruItem struct {
-	size      int64
-	committed bool
+	size int64
 }
 
 func (i *lruItem) Size() int64 {
@@ -91,59 +95,10 @@ func New(dir string, maxSizeBytes int64, proxy cache.Proxy) *Cache {
 	// This function is only called while the lock is held
 	// by the current goroutine.
 	onEvict := func(key Key, value sizedItem) {
-
 		f := filepath.Join(dir, key.(string))
-
-		if value.(*lruItem).committed {
-			// Common case. Just remove the cache file and we're done.
-			err := os.Remove(f)
-			if err != nil {
-				log.Printf("ERROR: failed to remove evicted cache file: %s", f)
-			}
-
-			return
-		}
-
-		// There is an ongoing upload for the evicted item. The temp
-		// file may or may not exist at this point.
-		//
-		// We should either be able to remove both the temp file and
-		// the regular cache file, or to remove just the regular cache
-		// file. The temp file is renamed/moved to the regular cache
-		// file without holding the lock, so we must try removing the
-		// temp file first.
-
-		// Note: if you hit this case, then your cache size might be
-		// too small (blobs are moved to the most-recently used end
-		// of the index when the upload begins, and these items are
-		// still uploading when they reach the least-recently used
-		// end of the index).
-
-		tf := f + ".tmp"
-		var fErr, tfErr error
-		removedCount := 0
-
-		tfErr = os.Remove(tf)
-		if tfErr == nil {
-			removedCount++
-		}
-
-		fErr = os.Remove(f)
-		if fErr == nil {
-			removedCount++
-		}
-
-		// We expect to have removed at least one file at this point.
-		if removedCount == 0 {
-			if !os.IsNotExist(tfErr) {
-				log.Printf("ERROR: failed to remove evicted item: %s / %v",
-					tf, tfErr)
-			}
-
-			if !os.IsNotExist(fErr) {
-				log.Printf("ERROR: failed to remove evicted item: %s / %v",
-					f, fErr)
-			}
+		err := os.Remove(f)
+		if err != nil {
+			log.Printf("ERROR: failed to remove evicted cache file: %s", f)
 		}
 	}
 
@@ -259,10 +214,7 @@ func (c *Cache) loadExistingFiles() error {
 	log.Println("Building LRU index.")
 	for _, f := range files {
 		relPath := f.name[len(c.dir)+1:]
-		ok := c.lru.Add(relPath, &lruItem{
-			size:      f.info.Size(),
-			committed: true,
-		})
+		ok := c.lru.Add(relPath, &lruItem{size: f.info.Size()})
 		if !ok {
 			err = os.Remove(filepath.Join(c.dir, relPath))
 			if err != nil {
@@ -275,10 +227,13 @@ func (c *Cache) loadExistingFiles() error {
 	return nil
 }
 
-// Put stores a stream of `expectedSize` bytes from `r` into the cache.
+// Put stores a stream of `size` bytes from `r` into the cache.
 // If `hash` is not the empty string, and the contents don't match it,
 // a non-nil error is returned.
-func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.Reader) error {
+func (c *Cache) Put(kind cache.EntryKind, hash string, size int64, r io.Reader) (rErr error) {
+	if size < 0 {
+		return fmt.Errorf("Invalid (negative) size: %d", size)
+	}
 
 	// The hash format is checked properly in the http/grpc code.
 	// Just perform a simple/fast check here, to catch bad tests.
@@ -287,95 +242,121 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
 			len(hash), sha256.Size)
 	}
 
-	if kind == cache.CAS && expectedSize == 0 && hash == emptySha256 {
+	if kind == cache.CAS && size == 0 && hash == emptySha256 {
 		io.Copy(ioutil.Discard, r)
 		return nil
 	}
 
 	key := cacheKey(kind, hash)
 
-	c.mu.Lock()
+	var tf *os.File // Tempfile.
 
-	// If there's an ongoing upload (i.e. cache key is present in uncommitted state),
-	// we drop the upload and discard the incoming stream. We do accept uploads
-	// of existing keys, as it should happen relatively rarely (e.g. race
-	// condition on the bazel side) but it's useful to overwrite poisoned items.
-	if existingItem, found := c.lru.Get(key); found {
-		if !existingItem.(*lruItem).committed {
-			c.mu.Unlock()
-			io.Copy(ioutil.Discard, r)
-			return nil
-		}
-	}
-
-	// Try to add the item to the LRU.
-	newItem := &lruItem{
-		size:      expectedSize,
-		committed: false,
-	}
-	ok := c.lru.Add(key, newItem)
-	c.mu.Unlock()
-	if !ok {
-		return &cache.Error{
-			Code: http.StatusInsufficientStorage,
-			Text: fmt.Sprintf("The item is larger (%d) than the cache's maximum size (%d).",
-				expectedSize, c.lru.MaxSize()),
-		}
-	}
-
-	// By the time this function exits, we should either mark the LRU item as committed
-	// (if the upload went well), or delete it. Capturing the flag variable is not very nice,
-	// but this stuff is really easy to get wrong without defer().
-	shouldCommit := false
-	filePath := cacheFilePath(kind, c.dir, hash)
+	// Cleanup intermediate state if something went wrong and we
+	// did not successfully commit.
+	unreserve := false
+	removeTempfile := false
 	defer func() {
-		c.mu.Lock()
-		if shouldCommit {
-			newItem.committed = true
-		} else {
-			c.lru.Remove(key)
+		// No lock required to remove stray tempfiles.
+		if removeTempfile {
+			os.Remove(tf.Name())
 		}
-		c.mu.Unlock()
 
-		if shouldCommit && c.proxy != nil {
-			// TODO: buffer in memory, avoid a filesystem round-trip?
-			fr, err := os.Open(filePath)
-			if err == nil {
-				c.proxy.Put(kind, hash, expectedSize, fr)
+		if unreserve {
+			c.mu.Lock()
+			err := c.lru.Unreserve(size)
+			if err != nil {
+				// Set named return value.
+				rErr = err
+				log.Printf(rErr.Error())
 			}
+			c.mu.Unlock()
 		}
 	}()
 
-	// Download to a temporary file
-	tmpFilePath := filePath + ".tmp"
-	f, err := os.Create(tmpFilePath)
+	if size > 0 {
+		c.mu.Lock()
+		ok, err := c.lru.Reserve(size)
+		if err != nil {
+			c.mu.Unlock()
+			return &cache.Error{
+				Code: http.StatusInternalServerError,
+				Text: err.Error(),
+			}
+		}
+		if !ok {
+			c.mu.Unlock()
+			return &cache.Error{
+				Code: http.StatusInsufficientStorage,
+				Text: fmt.Sprintf("The item (%d) + reserved space is larger than the cache's maximum size (%d).",
+					size, c.lru.MaxSize()),
+			}
+		}
+		c.mu.Unlock()
+		unreserve = true
+	}
+
+	// Final destination, if all goes well.
+	filePath := cacheFilePath(kind, c.dir, hash)
+
+	// We will download to this temporary file.
+	tf, err := tfc.Create(filePath)
 	if err != nil {
 		return err
 	}
-	defer func() {
-		if !shouldCommit {
-			// Only delete the temp file if moving it didn't succeed.
-			os.Remove(tmpFilePath)
+	removeTempfile = true
+
+	err = writeAndCloseFile(r, kind, hash, size, tf)
+	if err != nil {
+		return err
+	}
+
+	if c.proxy != nil {
+		rc, err := os.Open(tf.Name())
+		if err != nil {
+			log.Println("Failed to proxy Put:", err)
+		} else {
+			// Doesn't block, should be fast.
+			c.proxy.Put(kind, hash, size, rc)
 		}
-		// Just in case we didn't already close it.  No need to check errors.
-		f.Close()
+	}
+
+	unreserve, removeTempfile, err = c.commit(key, tf.Name(), filePath, size, size)
+
+	// Might be nil.
+	return err
+}
+
+func writeAndCloseFile(r io.Reader, kind cache.EntryKind, hash string, size int64, f *os.File) error {
+	closeFile := true
+	defer func() {
+		if closeFile {
+			f.Close()
+		}
 	}()
+
+	var err error
 
 	var bytesCopied int64
 	if kind == cache.CAS {
 		hasher := sha256.New()
-		if bytesCopied, err = io.Copy(io.MultiWriter(f, hasher), r); err != nil {
+		bytesCopied, err = io.Copy(io.MultiWriter(f, hasher), r)
+		if err != nil {
 			return err
 		}
 		actualHash := hex.EncodeToString(hasher.Sum(nil))
 		if actualHash != hash {
 			return fmt.Errorf(
-				"hashsums don't match. Expected %s, found %s", key, actualHash)
+				"checksums don't match. Expected %s, found %s", hash, actualHash)
 		}
 	} else {
 		if bytesCopied, err = io.Copy(f, r); err != nil {
 			return err
 		}
+	}
+
+	if isSizeMismatch(bytesCopied, size) {
+		return fmt.Errorf(
+			"sizes don't match. Expected %d, found %d", size, bytesCopied)
 	}
 
 	if err = f.Sync(); err != nil {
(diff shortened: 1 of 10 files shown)


mostynb:
@buchgr: any thoughts on this change?

@ulrfa, @lizan: please let me know if you're able to test this with your builds.

buchgr:
Thanks @mostynb . Taking a look!

mostynb:
> Thanks @mostynb . Taking a look!

The diff is a bit confusing, it might be easier to look at the entire file:
https://github.com/buchgr/bazel-remote/blob/1621686f98d07baefc117f5e57a8c068a3b160c9/cache/disk/disk.go

cache/disk/disk.go
```diff
@@ -288,153 +242,191 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
 
 	c.mu.Lock()
+	defer c.mu.Unlock()
 
-	existingItem, found := c.lru.Get(key)
-	if found {
-		if !existingItem.(*lruItem).committed {
-			inProgress = true
+	val, available := c.lru.Get(key)
+	if available {
```
buchgr, line 401: If you move computing `tryProxy` above the `if available` block you can flatten it

```go
if available {
  item := val.(*lruItem)
  if isSizeMismatch(size, item.size) {
    return nil, -1, tryPoxy
  }
  
  fileInfo, err = os.Stat(blobPath)
  if err != nil {
    return nil, -1, tryProxy
  }
  
  foundSize := fileInfo.Size()
  if isSizeMismatch(size, foundSize) {
    return nil, -1, tryProxy
  }
  
  f, err := os.Open(blobPath)
  if err != nil {
    return nil, -1, tryProxy
  }
  
  return f, foundSize, false
}
```

I think this is a lot easier to read but also makes it clear that `err` is not being handled. IIUC at least if `os.Open()` errs we should surface this no?

cache/disk/disk.go
```diff
@@ -288,153 +242,191 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
 	}
 
-	c.mu.Unlock()
-
-	available = found && !inProgress
+	if c.proxy != nil {
+		if size > 0 {
+			// If we know the size, attempt to reserve that much space.
+			tryProxy = c.lru.Reserve(size)
+		} else {
```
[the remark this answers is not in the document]
mostynb: The HTTP API allows Content-Length to be unset/-1.

cache/disk/lru.go
```diff
@@ -120,6 +124,44 @@ func (c *sizedLRU) MaxSize() int64 {
 	return c.maxSize
 }
 
+func (c *sizedLRU) Reserve(size int64) bool {
+	if size < 0 || size > c.maxSize || c.reservedSize+size > c.maxSize {
```
buchgr, line 128: `if size < 0 || c.reservedSize+size > c.MaxSize`

When would size be < 0?

cache/disk/lru.go
```diff
@@ -27,16 +27,20 @@ type SizedLRU interface {
 	Get(key Key) (value sizedItem, ok bool)
 	Remove(key Key)
 	Len() int
-	CurrentSize() int64
+	CurrentSize() int64 // Returns the current used + reserved size.
```
buchgr, line 30: Rename this method to CommittedAndReservedSize()? Might be interesting to export both as stats?

cache/disk/lru.go
```diff
@@ -27,16 +27,20 @@ type SizedLRU interface {
...
 }
 
 type sizedLRU struct {
 	// Eviction double-linked list. Most recently accessed elements are at the front.
 	ll *list.List
 	// Map to access the items in O(1) time
-	cache       map[interface{}]*list.Element
-	currentSize int64
+	cache        map[interface{}]*list.Element
+	currentSize  int64 // Includes reserved size.
```
buchgr, line 42: nit move comment one line up

cache/disk/lru.go
mostynb: This is a safety check for HTTP API request where Content-Length is unset/-1.

buchgr: Also shouldn't you be checking c.currentSize instead of c.reservedSize?

cache/disk/lru.go
```diff
@@ -120,6 +124,44 @@ func (c *sizedLRU) MaxSize() int64 {
...
+		return false
+	}
+
+	// Evict elements until we are able to reserve enough space.
+	for c.currentSize+size > c.maxSize {
+		ele := c.ll.Back()
+		if ele != nil {
+			c.removeElement(ele)
+		} else {
+			return false // This should have been caught at the start.
```
buchgr, line 138: If this happens it's a bug?

cache/disk/lru.go
```diff
@@ -120,6 +124,44 @@ func (c *sizedLRU) MaxSize() int64 {
...
+func (c *sizedLRU) Unreserve(size int64) bool {
+	if size < 0 {
+		return false
+	}
+
+	newC := c.currentSize - size
+	newR := c.reservedSize - size
+
+	if newC < 0 || newR < 0 {
+		return false
```
buchgr, line 156: It's a bug if this happens?

mostynb: Yes.

cache/disk/lru.go
mostynb: Yes.

cache/disk/disk.go
```diff
@@ -288,153 +242,191 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
 	if err != nil {
 		return err
 	}
+
+	if c.proxy != nil {
+		rc, err := os.Open(tf.Name())
+		if err != nil {
+			log.Println("Failed to proxy Put:", err)
+		} else {
+			c.proxy.Put(kind, hash, size, rc) // Async, should be fast.
```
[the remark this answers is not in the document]
mostynb: It feels wrong to let this proxy.Put block cause the entire operation to fail (proxy.Put is allowed to fail silently). But if `os.Open(tf.Name())` fails then the `commit` is very likely to fail too, I guess.

cache/disk/disk.go
mostynb: One issue with that setup is that calculating `tryProxy` will (usually) need to reserve space, but we do not want to reserve space if the blob available locally. I guess we could do that with a more complicated deferred function (maybe unreserve space, then unlock the mutex).

cache/disk/lru.go
mostynb: Checking with `reservedSize` instead of `currentSize` is definitely a bug.

But just checking `c.currentSize+size > c.maxSize` for going over the upper bound is problematic if the result overflows int64. Maybe unlikely in practice, but `size` comes directly from the client. If the client sends an unusually value, then `size > c.maxSize` is likely to catch this, but I think I can come up with a more correct check.

cache/disk/lru.go
mostynb: I wonder what is most useful form? This is only used by `Stats()` and tests. A pair of `<current used + reserved size>, <reserved size>` seems reasonable and won't confuse existing users who are collecting and graphing "CurrSize".

Callers could use `totalSize, reservedSize := c.CurrentSize()` which is pretty readable IMO.

cache/disk/disk.go
```diff
@@ -288,153 +242,191 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
+		}
+	}
+
+	unreserve, removeTempfile, err = c.commit(key, tf.Name(), filePath, size)
+	return err // Might be nil.
```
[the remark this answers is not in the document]
mostynb: Done.

cache/disk/disk.go
```diff
@@ -288,153 +242,191 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
+}
+
+func writeAndCloseFile(r io.Reader, kind cache.EntryKind, hash string, size int64, f *os.File) error {
+	closeFile := false
 	defer func() {
-		if !shouldCommit {
-			// Only delete the temp file if moving it didn't succeed.
-			os.Remove(tmpFilePath)
+		if closeFile {
```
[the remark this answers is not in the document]
mostynb: Oops, flipped the initialisation. Nice catch.

cache/disk/disk.go
```diff
@@ -288,153 +242,191 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
-		if bytesCopied, err = io.Copy(io.MultiWriter(f, hasher), r); err != nil {
+		bytesCopied, err = io.Copy(io.MultiWriter(f, hasher), r)
+		if err != nil {
 			return err
 		}
 		actualHash := hex.EncodeToString(hasher.Sum(nil))
 		if actualHash != hash {
 			return fmt.Errorf(
-				"hashsums don't match. Expected %s, found %s", key, actualHash)
+				"hashsums don't match. Expected %s, found %s", hash, actualHash)
```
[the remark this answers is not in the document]
mostynb: Done.

cache/disk/lru.go
mostynb: Done.

cache/disk/lru.go
mostynb: Hmm, wait... We shouldn't check `c.currentSize+size > c.maxSize` at the start, since we can evict items to make space.

This code was checking `c.reservedSize+size > c.maxSize` because if that fails, we can't evict items to reserve enough space. I will change that part back for now, while still considering how to handle overflow.

lizan:
@mostynb I was able to test at 7c1e9bd2305b521be0baecb86b0a4f696e26d44e and it fixed my issue.

cache/disk/lru.go
mostynb: Renamed to `TotalSize()` and added `ReservedSize()`. I left the label in the json output as `CurrSize`.

cache/disk/lru.go
mostynb: Refactored this a bit to improve the overflow checking, and added a comment about why we check size+reservedSize.

cache/disk/disk.go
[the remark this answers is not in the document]
buchgr: Do we need to support chunked encoding though? I believe both client and server always know the length of the content in advance?

cache/disk/lru.go
buchgr: Shouldn't we fail somehow then and not just return false? This will put the cache in an undefined state?

cache/disk/disk.go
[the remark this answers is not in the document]
mostynb: Sorry, I got mixed up. Content-Length is not relevant for http Get. I was trying to refer to the fact that incoming HTTP Get requests only specify the hash, without the size:

https://github.com/buchgr/bazel-remote/blob/fe90bd14fb2715599afb27ef3c554a5bc5e1d69e/server/http.go#L196

Also, in the gRPC API clients do provide the expected size for CAS items, but cannot do so for the action cache. So these also result in a -1 here.

lizan:
@mostynb @buchgr thank you for working on this, any update?

mostynb:
> @mostynb @buchgr thank you for working on this, any update?

I will do another pass on this and try to land it next week.

dstranz:
Hello @mostynb, 

We have similar issue when using bazel 3.4.1 and bazel-remote:master instance with s3 storage as remote-cache. 

I've just deploy bazel-remote from your branch on our server but it's not resolving this issue.

```
WARNING: Reading from Remote Cache:
BulkTransferException
WARNING: Writing to Remote Cache:
BulkTransferException
```
or
```
WARNING: Reading from Remote Cache:
ClosedChannelException
```

Let me know, if I could help you with testing your fix.

mostynb:
@dstranz: are there any errors in bazel-remote's logs when you encounter this error? And do the bazel client errors mention any specific blob hashes that we can look for in bazel-remote's logs?

dstranz:
@mostynb I will try to compare build output with bazel-remote logs tomorrow and let you know.

ulrfa:
> @ulrfa, ... please let me know if you're able to test this with your builds.

I had a long summer vacation and is now trying to catch up. I will try to find time to review and test next week.

dstranz:
@mostynb Looking into bazel-remote logs was a very good point. I figure out that our issue was related to s3 configuration - after fixing it everything works perfectly.

cache/disk/disk.go
```diff
@@ -288,161 +242,220 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
-	// Download to a temporary file
-	tmpFilePath := filePath + ".tmp"
-	f, err := os.Create(tmpFilePath)
+	err = writeAndCloseFile(r, kind, hash, size, tf)
 	if err != nil {
 		return err
 	}
+
+	if c.proxy != nil {
+		rc, err := os.Open(tf.Name())
```
ulrfa, line 314: Smart that you use the temporary file, instead of the final destination file, for the proxy, so that no lock is needed.

cache/disk/disk.go
```diff
@@ -288,161 +242,220 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
+	val, available := c.lru.Get(key)
+	if available {
+		item := val.(*lruItem)
+		if !isSizeMismatch(size, item.size) {
+			fileInfo, err = os.Stat(blobPath)
+			if err == nil {
+				foundSize := fileInfo.Size()
+				if !isSizeMismatch(size, foundSize) {
+					var f *os.File
+					f, err = os.Open(blobPath)
```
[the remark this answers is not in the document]
mostynb: I don't have access to a setup for high concurrency testing at the moment, So I have been mostly checking for correctness.

IIRC my reasoning here was that calling `os.Stat()` and then `os.Open()` without a lock is racy. But now I realise that we can open the file, release the lock, and then call `File.Stat()` safely. So then we would be down to one atomic filesystem operation while checking the availability, and we can decide if it's OK for there to be a race between checking the index (with the lock) and checking the filesystem (without the lock).

If we release the lock before opening the file, there's a potential race condition with the item being purged (or overwritten, but I think that case is OK). The purged-before-open is most likely when the disk cache is small, eg as some users have when using a proxy backend. But in this scenario, we would just fall back to retrieving the file from the proxy backend. We currently have this race on master, and it's probably OK- the cache is too small for the workload and should be increased. I will push a change to do this.

cache/disk/disk.go
```diff
@@ -288,161 +242,220 @@ func (c *Cache) Put(kind cache.EntryKind, hash string, expectedSize int64, r io.
...
-		// This flag is used by the defer() block above.
-		shouldCommit = true
+	if !c.lru.Add(key, &lruItem{size: foundSize}) {
+		err = fmt.Errorf("INTERNAL ERROR: failed to add: %s, size %d", key, foundSize)
+		log.Println(err.Error())
+		return unreserve, removeTempfile, err
 	}
 
-	return err
+	err = os.Rename(tempfile, finalPath)
```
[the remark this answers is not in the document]
mostynb: If the cache item filenames contain the size, then it would be difficult to service requests (like HTTP GET/HEAD) that do not specify a size.

I think we need the file renaming and insertion into the index to be atomic for a given key. Otherwise overlapping Put calls can leave the file on disk out of sync with the index.

Aside: I just notice that `os.Rename()` is not atomic, it calls `os.Lstat()` and then `syscall.Rename()`. We may as well just call `syscall.Rename()` directly.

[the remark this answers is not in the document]
mostynb: Hmm, another idea... what if we write to a tempfile in the dir, then rename it to something which indicates that we have finished writing to it, then obtain the lock, and lru.Add an item to the LRU which contains the final filename and the same key as now. lru.Add could then return the previous filename if there was one. Then we release the lock and remove the old file that was returned (if there was one).

On startup, if there are multiple "finished" files for a given key we keep the one with the most recent atime and delete the others.

But I think this change might be best done in a followup PR, if calling syscall.Rename turns out to be too slow.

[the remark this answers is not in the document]
ulrfa: > If the cache item filenames contain the size, then it would be difficult to service requests (like HTTP GET/HEAD) that do not specify a size.

For PUT we know the size after writeAndCloseFile finished for the temporary file. And for GET/HEAD with unspecified size, perhaps we could get the size from LRU, via the hash as key?

> I think we need the file renaming and insertion into the index to be atomic for a given key. Otherwise overlapping Put calls can leave the file on disk out of sync with the index.

I’m thinking about if embedding both hash and size in the final file name, could avoid the risk for out of sync between file on disk and index.

> Aside: I just noticed that os.Rename() is not atomic, it calls os.Lstat() and then syscall.Rename(). We may as well just call syscall.Rename() directly.

Good finding!

ulrfa:
> I don't have access to a setup for high concurrency testing at the moment, So I have been mostly checking for correctness.

At Ericsson we deploy bazel-remote on 72 core machines. I plan to allocate some of them during the coming weekend, and evaluate if holding the mutex during syscall.Rename and os.Open, affects build performance or not, when there are many concurrent GET and PUT.

The previous conversation is easily lost in the collapsed code review for outdated code above. @mostynb, do you have more reflections about embedding both hash and size in the file path, to potentially reduce the need for holding mutex during the syscalls?

mostynb:
> At Ericsson we deploy bazel-remote on 72 core machines. I plan to allocate some of them during the coming weekend, and evaluate if holding the mutex during syscall.Rename and os.Open, affects build performance or not, when there are many concurrent GET and PUT.

Thanks- that would be really useful.

> The previous conversation is easily lost in the collapsed code review for outdated code above. @mostynb, do you have more reflections about embedding both hash and size in the file path, to potentially reduce the need for holding mutex during the syscalls?

Unfortunately I think it might be racy, eg:

* abc-111 is in the index and on disk
* request 1: add abc-222 to disk
* request 1: add abc-222 to index
* request 2: add abc-111 to disk
* request 1: remove abc-111 from disk (no lock)
* request 2: add abc-111 to the index
* request 2: remove abc-222 from disk (no lock)

A more complicated version might work, where we write to files of the form <hash>.tmp<unique> then move them to <hash>.done<unique> and then insert them into the index (including a field with the <unique> string). If we're replacing an item in the index then we remove the old <hash>.done<otherunique> file from disk after releasing the lock. That way there is much lower chance of collision between racing requests.

But I think this change is large enough that we should land it in the more safe version (removing the files while holding the lock), and then follow up with an optimisation if it's too slow. Changing the on-disk format in this way probably requires a major version bump.

ulrfa:
Perhaps a {hash}-{size} named file would not have to be removed by a later PUT of same hash but different size. Instead a new entry could be added to the LRU’s double linked list, and the original entry could still be in the double linked list and be evicted the normal way. That might involve having hash+size as key in the LRU, just as the gRPC protocol seems to do conceptually. And for HTTP with unspecified size, complement with an additional map in LRU with only hash as key, towards the most recent regardless of size. I think that could avoid the problematic race and still not require holding mutex during syscall.Rename.

However, I hope to get a chance to benchmark during the weekend. If no significant performance penalty, then we don’t need to bother about this and can keep it simple. :-)

ulrfa:
I benchmarked three alternative implementations using mutexes in different ways:
 
- **mutex rename** = Almost identical to 983d785 from mostynb:concurrent_inserts
- **mutex minimal** = As mutex rename, but releases the mutex during syscall.Rename in commit.
- **mutex open + rename** = As mutex rename, but keeps the mutex locked also during os.Open in availableOrTryProxy.

Each implementation is benchmarked in two different scenarios. Each scenario is executed using up to 87 bazel clients started in parallell, for generating the load. Each client on separate machine. The bazel-remote server is deployed on a machine with 72 logical processors (2 x Xeon 6154) and 2 x 10 Gbit network interface (link aggregation).

### Scenario A: 0% Cache Hits

-	All clients doing same build, but with unique dummy –action_env to trigger cache misses.
-	Each bazel client send:
             ~7000 HTTP GET requests for non existing AC entries,
             ~30 000 HTTP PUT requests for uploading AC and CAS entries. 
             No HTTP HEAD. No gRPC.
-	Nothing evicted from cache during benchmark.
-	**Bazel-remote scales really well if not holding mutexes during system calls.**
-	**The graph indicates holding mutexes during syscall.Rename, starts to cause lock contention issues from 50 concurrent bazel clients. Increases build times 70% at 87 clients.**
-	The SSD seems to handle the load. Messured Write IOPS peek at 65 000 and sustained 35 000. (But it could be interesting to see how much time f.sync() takes...)

bazelremote_mutex_graph_A

### Scenario B: 100% Cache Hits

-	Each bazel client send:
	     ~36 000 HTTP GET requests downloading AC and CAS entries.
	     No HTTP PUT, No HTTP HEAD. No gRPC.
-	Almost no disk accesses, all data already in kernel’s RAM file system cache.
-	The 2 x 10 Gbit network interfaces become saturated from ~30 concurrent clients.


bazelremote_mutex_graph_B

mostynb:
Thanks for running these benchmarks. Here are some initial questions/notes:
* Did you hit eny errors during testing?
* In the hot-cache/100% cache hit scenario, it looks like "mutex rename" performs pretty well. And since the network connection is saturated we might need to wait for something like https://github.com/bazelbuild/remote-apis/issues/147 to improve on this significantly.
* There's clearly room for improvement over "mutex rename" in the cold-cache/0% cache hit scenario, but hopefully this isn't a common case to be in.
* Re `f.Sync()`, there's an old issue about this being slow on Mac hardware (#67). I had a prototype for working around this, but it has gone stale.

ulrfa:
Thanks!

I believe the logic is correct, and did not notice any errors (related to this commit) during the load test.

If you decide to merge 983d785 now, despite the performance degradation, then it could perhaps be considered to embed both hash and file size in file path strings, the next time the on-disk format is changed for other reasons? (or address the degradation in other ways)

For the hot-cache/100% cache hit scenario, with saturated network interface, I agree that something like bazelbuild/remote-apis#147 and/or https://github.com/bazelbuild/bazel/issues/6862 would be needed.

mostynb:
> Thanks!
> 
> I believe the logic is correct, and did not notice any errors (related to this commit) during the load test.
> 
> If you decide to merge [983d785](https://github.com/buchgr/bazel-remote/commit/983d785883a05e177f413aa93413f7c3b8e8a486) now, despite the performance degradation, then it could perhaps be considered to embed both hash and file size in file path strings, the next time the on-disk format is changed for other reasons? (or address the degradation in other ways)

This PR is pretty large already, I think it's best to land it with the potential performance hit but continue working on improvements to see if we can win some of that back. This is surely faster than a failed build.

We can change the on-disk format later if needed (and probably bump the version number to 2.0.0 in that case).

> For the hot-cache/100% cache hit scenario, with saturated network interface, I agree that something like [bazelbuild/remote-apis#147](https://github.com/bazelbuild/remote-apis/issues/147) and/or [bazelbuild/bazel#6862](https://github.com/bazelbuild/bazel/issues/6862) would be needed.

We can continue optimising after we get compression support (builds-without-the-bytes are already supported of course).

ulrfa:
Thanks @mostynb! Sounds good!

PR #325, opened 2020-07-31
cache.Proxy.Put should take an io.ReadCloser instead of io.Reader

mostynb:
We pass an *io.File to the cache.Proxy.Put, which should be closed.

### DIFF
--- a/cache/s3proxy/s3proxy.go
+++ b/cache/s3proxy/s3proxy.go
@@ -22,7 +22,7 @@ type uploadReq struct {
 	hash string
 	size int64
 	kind cache.EntryKind
-	rdr  io.Reader
+	rc   io.ReadCloser
 }
 
 type s3Cache struct {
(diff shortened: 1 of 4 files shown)


cache/s3proxy/s3proxy.go
```diff
@@ -151,18 +151,21 @@ func (c *s3Cache) uploadFile(item uploadReq) {
...
 	case c.uploadQueue <- uploadReq{
 		hash: hash,
 		size: size,
 		kind: kind,
-		rdr:  rdr,
+		rc:   rc,
 	}:
 	default:
 		c.errorLogger.Printf("too many uploads queued\n")
+		rc.Close()
```
mostynb, line 168: I suspect we should drain the reader first. I wonder if that can be slow?

mostynb: We don't need to drain the reader, since it's an *os.File.

PR #337, opened 2020-08-22
Unify entry path creation directory structure.

jruhland:
The intent of this change is to expand the rate capacity of the s3 proxy, while at the same time unifying the path creation for the different caches.  S3 rate limiting is done by prefix (roughly translated to directory) at 3500/5500 qps.  Adding the additional depth allows for a large increase in throughput before hitting rate limits.

* Add cache.Key for common usage
* Refactor disk cache to utilize cache.Key
* Refactor s3proxy to utilize cache.Key

**DISCLAIMER**
This invalidates all existing S3 caches, if thats untenable, I'm happy to put in a flag that toggles this feature on/off for s3.  I don't want to pollute the flag space unless it's desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ commodity hardware and AWS servers. Outgoing bandwidth can exceed 15 Gbit/s on t
 ## HTTP/1.1 REST API
 
 Cache entries are set and retrieved by key, and there are two types of keys that can be used:
+
 1. Content addressed storage (CAS), where the key is the lowercase SHA256 hash of the stored value.
    The REST API for these entries is: `/cas/<key>` or with an optional but ignored cache pool name: `/<pool>/cas/<key>`.
 2. Action cache, where the key is an arbitrary 64 character lowercase hexadecimal string.
@@ -34,6 +35,7 @@ for GET requests and `Content-type: application/json` for PUT requests.
 **/status**
 
 Returns the cache status/info.
+
 ```
 $ curl http://localhost:8080/status
 {
@@ -49,6 +51,7 @@ $ curl http://localhost:8080/status
 
 The empty CAS blob is always available, even if the cache is empty. This can be used to test that
 a bazel-remote instance is running and accepting requests.
+
 ```
 $ curl --head --fail http://localhost:8080/cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
 HTTP/1.1 200 OK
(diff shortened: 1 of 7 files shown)


cache/cache.go
```diff
@@ -64,3 +65,8 @@ type Proxy interface {
 	// unknown).
 	Contains(kind EntryKind, hash string) (bool, int64)
 }
+
+// Key returns the proper cache key for an entry kind and hash
```
mostynb, line 69: nitpick: please add `.` at the end of the sentence.

cache/s3proxy/s3proxy.go
```diff
@@ -114,9 +114,10 @@ func New(s3Config *config.S3CloudStorageConfig, accessLogger cache.Logger,
 
 func (c *s3Cache) objectKey(hash string, kind cache.EntryKind) string {
 	if c.prefix == "" {
-		return fmt.Sprintf("%s/%s", kind, hash)
+		return cache.Key(kind, hash)
 	}
-	return fmt.Sprintf("%s/%s/%s", c.prefix, kind, hash)
+
+	return fmt.Sprintf("%s/%s", c.prefix, cache.Key(kind, hash))
```
mostynb, line 120: I think adding a flag to enable the new key format is better than the alternative. I am unsure how (or if) people are removing old s3 blobs, and that might setup might also require an update when using the new key format.

Do you have any ideas for a good name? https://cloud.google.com/storage/quotas suggests that GCS doesn't rate limit any operations we need to scale, so keeping this as an s3-specific setting seems wise. Maybe `--s3.new_key_format` or `--s3.three_level_key`?

mostynb (changes requested):
Are you hitting this limit in practice, and have you measured the improvement after testing this patch?

jruhland:
> Are you hitting this limit in practice, and have you measured the improvement after testing this patch?

Yes, we ran up to this limit aggressively.  In our forked version we actually added a third character to the structure, and have not run up against it at all.  I felt that was too aggressive in 1 PR for upstream.

cache/s3proxy/s3proxy.go
jruhland: Yeah, I think since GCS is a natively supported cache, rate limiting should not be a concern there.  Im happy with either of those, first makes more sense personally.

jruhland:
@mostynb I think I've addressed all the comments from review, and added the flag and corresponding unit test for enabling this feature.

README.md
```diff
@@ -310,3 +314,7 @@ To avoid leaking your password in log files, you can place this flag in a
 For more details, see Bazel's [remote
 caching](https://docs.bazel.build/versions/master/remote-caching.html#run-bazel-using-the-remote-cache)
 documentation.
+
+## AWS S3 Notes
```
mostynb, line 318: nitpick: s/Notes/note/ for consistency with the "Kubernetes note" section

mostynb (changes requested):
> > Are you hitting this limit in practice, and have you measured the improvement after testing this patch?
> 
> Yes, we ran up to this limit aggressively. In our forked version we actually added a third character to the structure, and have not run up against it at all. I felt that was too aggressive in 1 PR for upstream.

Thinking about the flag name some more, maybe `--s3.key_version_2` is more flexible in case we decide to tweak it again. That way we wouldn't need to resort to `--s3.new_new_key_format`

Also, could you add a go_test target to cache/s3proxy/BUILD.bazel ? `bazel run //:gazelle -- fix` might be able to create it for you, otherwise you can see the example in cache/disk/BUILD.bazel: https://github.com/buchgr/bazel-remote/blob/47774b5db7ac141bc832f71a440c5c507e0574d9/cache/disk/BUILD.bazel#L21

Otherwise, this looks good to land.

kragniz:
I'm not sure if this has been done for other flags, but could this aim to become the default in a few releases? At that point the flag can be removed and the `new_new_key_format` style naming doesn't become a problem (and there's less accumulation of flags that really should be enabled for any new users).

mostynb:
> I'm not sure if this has been done for other flags, but could this aim to become the default in a few releases? At that point the flag can be removed and the `new_new_key_format` style naming doesn't become a problem (and there's less accumulation of flags that really should be enabled for any new users).

We haven't deprecated any flags before, mostly because I don't have a way to know how people are using bazel-remote and how much breakage this would cause.

Maybe it's worth moving towards making a configuration file as the primary setup. eg:
* If there's no existing configuration file on startup, generate one with the current defaults.
* If there is an existing configuration file, use it. If it's missing a field, insert it with a non-breaking value and a comment that the admin should make a decision here.
* Print warnings at startup about configuration values that still have the "please make a decision" comment.

mostynb:
I can clean this up and land it, if you like?

I have been thinking about the flag name some more, and came up with `--s3.key_format=N` which is less likely to require new flags for this feature.

jruhland:
@mostynb hey sorry got pulled into other stuff and lost time to work on it, by all means have at it

mostynb:
Landed this: https://github.com/buchgr/bazel-remote/pull/345

PR #339, opened 2020-08-25
add option to mangle AC keys with (non-empty) instance names

### LINKED ISSUE #15: Allow per project cache

mostynb:
In some client setups, untracked local files can be used by an action without being included in the Action message, which causes action cache collisions: https://github.com/bazelbuild/bazel/issues/4558

Ideally this should be fixed on the client side (either in the client, or in the build configuration), but it is not always easy to do in practice.

As a workaround, this patch adds a setting to mangle ActionCache keys with the instance name provided by the client (if it is not empty), to produce a new ActionCache key. Clients are then able to specify a different instance name whenever a change is made that could affect these untracked inputs. The instance name value could be something like the hash of the compiler version. This allows multiple ActionCache items to exist in the cache, without requiring a change to the on-disk storage format.

This feature is disabled by default, since it would cause cache invalidations for existing users.

Fixes #15.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,21 @@ commodity hardware and AWS servers. Outgoing bandwidth can exceed 15 Gbit/s on t
 
 Cache entries are set and retrieved by key, and there are two types of keys that can be used:
 1. Content addressed storage (CAS), where the key is the lowercase SHA256 hash of the stored value.
-   The REST API for these entries is: `/cas/<key>` or with an optional but ignored cache pool name: `/<pool>/cas/<key>`.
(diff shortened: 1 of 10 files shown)


kragniz:
From a brief read, this looks great! I'll exercise it locally over the next day or two.

kragniz (approved):
This has been running in a dev environment without problems

PR #340, opened 2020-08-26
Don't add to cache if everything reserved

glukasiknuro:
Edge case, but looks like previous behavior was unexpected in this case:
the Add method would:
  - call removeElement on the newly added element;
  - not add the new entry to the LRU;
  - return true.


mostynb (commented):
Nice catch- thanks!

PR #348, opened 2020-09-10
Add disk cache size metrics

ulrfa:
Add Prometheus metrics for monitoring:

- How large is the disk cache now?

- How fast has the disk cache been growing?

- At what rate are cache data evicted?

- How common is it with multiple put requests of the same key?

### DIFF
--- a/cache/disk/lru.go
+++ b/cache/disk/lru.go
@@ -4,6 +4,26 @@ import (
 	"container/list"
 	"errors"
 	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	gaugeCacheSizeBytes = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "bazel_remote_disk_cache_size_bytes",
+		Help: "The current number of bytes in the disk backend",
+	})
+
+	counterEvictedBytes = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_evicted_bytes_total",
+		Help: "The total number of bytes evicted from disk backend, due to full cache",
+	})
+
+	counterOverwrittenBytes = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_overwritten_bytes_total",
+		Help: "The total number of bytes removed from disk backend, due to put of already existing key",
+	})
 )
 
 type sizedItem interface {
@@ -88,6 +108,7 @@ func (c *sizedLRU) Add(key Key, value sizedItem) (ok bool) {
 			return false
 		}
 		c.ll.MoveToFront(ee)
+		counterOverwrittenBytes.Add(float64(ee.Value.(*entry).value.Size()))
 		ee.Value.(*entry).value = value
 	} else {
 		sizeDelta = value.Size()
@@ -109,6 +130,8 @@ func (c *sizedLRU) Add(key Key, value sizedItem) (ok bool) {
 
 	c.currentSize += sizeDelta
 
+	gaugeCacheSizeBytes.Set(float64(c.currentSize))
+
 	return true
 }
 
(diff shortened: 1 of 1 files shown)


cache/disk/lru.go
```diff
@@ -4,6 +4,26 @@ import (
...
+
+	counterEvictedBytes = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_evicted_bytes_total",
+		Help: "The total number of bytes evicted from disk backend, due to full cache",
+	})
+
+	counterOverwrittenBytes = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_overwritten_bytes_total",
+		Help: "The total number of bytes removed from disk backend, due to put of already existing key",
+	})
```
mostynb, line 26: What are we aiming to count here- attempted duplicate Puts, or duplicate Puts that we didn't short-circuit?

cache/disk/lru.go
```diff
@@ -4,6 +4,26 @@ import (
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	gaugeCacheSizeBytes = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "bazel_remote_disk_cache_size_bytes",
+		Help: "The current number of bytes in the disk backend",
+	})
```
mostynb, line 16: We expect the cache size to increase until it reaches the limit, and then hover around there. What else would we look out for in this graph?

cache/disk/lru.go
```diff
@@ -4,6 +4,26 @@ import (
...
+
+	counterEvictedBytes = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "bazel_remote_disk_cache_evicted_bytes_total",
+		Help: "The total number of bytes evicted from disk backend, due to full cache",
+	})
```
mostynb, line 21: We expect this graph to stay at 0 until the cache is full, then increase with every Put that isn't short-circuited. What else should we be looking for in this graph?

mostynb (commented):
Thanks for the PR- just a few questions to make sure I understand why it might be useful to monitor these things.

cache/disk/lru.go
ulrfa: Yes. In some deployments it can take several weeks to reach the limit, and during that time it shows how fast the cache is growing and gives early indication about if the limit is high enough to keep entries for a reasonable amount of time. And also if growing with constant rate or flattern out after all products have been built at least once. Or if there are sudden jumps, either expected or unexpected, that could indicate if artifacts are not shared as desired.

After limit has been reached, it becomes a convenient way to show current configured limit for several caches, on same dashboard that shows other cache metrics. Observing that it hovers around the limit makes me feel that bazel-remote is working correctly. ;-)

I had this patch for a long time, and my users seems to have an intuitive understanding about what it means.

cache/disk/lru.go
ulrfa: Yes. After reaching the limit, it provides similar information as gaugeCacheSizeBytes does before reaching the limit.

This value also seems to be intuitive to understand for my users. (they don't know about short-circuited PUT, etc)

cache/disk/lru.go
ulrfa: My users does not understand this value on the dashboard, but for me it is interesting to see how common it is for the bazel client to re-upload same data several times. Either in the same session by concurrently executed actions, or by later sessions not checking if file exists before uploading. I had this patch since long time ago, before short-circuit functionality in 2d9211c3. Perhaps it could be interesting with a counter for skipped ones, to get the whole picture? Or do you have other suggestion?

cache/disk/lru.go
ulrfa: On a dashboard I show this value in a graph, with date on x axis, not only what the value is right now.

cache/disk/lru.go
mostynb: I think it's ok not to count skipped uploads, they're a part of normal operation from the point of view of the server.

cache/disk/lru.go
mostynb: Sounds reasonable.

cache/disk/lru.go
mostynb: I suppose it might be interesting to observe changes that invalidate large parts of the cache.

mostynb:
Thanks!

ulrfa:
Thanks for accepting the patch!

PR #350, opened 2020-09-16
Add AC hit rate metrics with prometheus labels

ulrfa:
This is a draft, not ready to be merged. I wait with test cases until getting feedback about the main functionality. 

Summary of what is added by this commit:

 - Prometheus counter for cache hit ratio of only AC requests.
 - Support for prometheus labels based on custom HTTP and gRPC headers.

Cache hit ratio for CAS entries is easily misinterpreted. Example:

  A typical action cache hit often involves 3 or more HTTP requests:

    GET AC 200
    GET CAS 200 (.o file)
    GET CAS 200 (.d file)
    ...

  But a cache miss for the same action is typically a single HTTP request:

    GET AC 404

The ratio between all HTTP GET 200 vs HTTP GET 404 above does not represent
the cache hit ratio experienced by the user for actions. The ratio of only
AC requests is easier to reason about, especially when AC requests checks
existence of CAS dependencies.

The number of AC hits and misses can be directly compared against numbers
printed in the end of each build by bazel client. And against other
prometheus counters produced by remote execution systems for executed actions.

An understanding about the reason for cache misses is necessary to improve the
cache hit ratio. It could be that the system has been configured in a way
that prevent artifacts from being reused between different OS. Or that the
cache is only populated by CI jobs on master, potentially resulting in cache
misses for other users, etc. It becomes easier to notice such patterns, if
cache hit ratio could be calculated for different categories of builds.
Such categories can be set as custom headers via bazel flags
--remote_header=branch=master and applied as prometheus labels. Mapping of
headers to prometheus labels are controlled in bazel-remote's config file.

The ratio between cache uploads and cache misses is also relevant, as an
view about which categories are not uploading their result. The ratio of cache
uploads can also indicate if much is uploaded but seldom requested. E.g. does it
make sense to populate central caches from interactive builds or only from CI?

Categories and custom headers, could also be set for an overview about:
 - Bazel versions using a cache instance?
 - How much separate organizations are using a cache instance?
 - From which network traffic originates?
 - Which products are built using the cache?
 - If the traffic comes via proxy adding its own headers?
 - Distinguish dummy requests for monitoring the cache, from real requests?
 - ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,23 @@ host: localhost
 
 # If true, enable experimental remote asset API support:
 #experimental_remote_asset_api: true
+
+# Allows mapping HTTP and gRPC headers to prometheus
+# labels. Headers can be set by bazel client as:
+# --remote_header=os=ubuntu18-04. Not all counters are
+# affected.
+#metrics:
+#  categories:
+#    os:
+#      - rhel7
+#      - rhel8
+#      - ubuntu16-04
+#      - ubuntu18-04
+#    branch:
+#      - master
+#    user:
+#      - ci
+
 ```
 
 ## Docker
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -21,6 +21,8 @@ const (
 	// used for HTTP when running with the --disable_http_ac_validation
 	// commandline flag.
 	RAW
+
+	UNKNOWN
 )
 
 func (e EntryKind) String() string {
@@ -30,7 +32,10 @@ func (e EntryKind) String() string {
 	if e == CAS {
 		return "cas"
 	}
-	return "raw"
+	if e == RAW {
+		return "raw"
+	}
+	return "unknown"
 }
 
 // Logger is designed to be satisfied by log.Logger.
--- a/utils/metrics/metrics.go
+++ b/utils/metrics/metrics.go
@@ -1,0 +1,187 @@
+package metrics
+
+import (
+	"github.com/buchgr/bazel-remote/config"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"strings"
+)
+
+// TODO Add test cases for this file
+
+type Method int
+type Status int
+type Kind int
+type Protocol int
+
+const (
+	METHOD_GET Method = iota
+	METHOD_HEAD
+	METHOD_PUT
+	METHOD_OTHER
+)
+
+const (
+	OK Status = iota
+	NOT_FOUND
+	OTHER_STATUS
+)
+
+const (
+	AC Kind = iota
+	CAS
+)
+
+const (
+	HTTP Protocol = iota
+	GRPC
+)
+
+func (e Method) String() string {
+	// Actually HTTP names, but can be conceptually mapped also to GRPC protocol.
+	if e == METHOD_GET {
+		return "GET"
+	}
+	if e == METHOD_HEAD {
+		return "HEAD"
+	}
+	if e == METHOD_PUT {
+		return "PUT"
+	}
+	return "other"
+}
+
+func (e Status) String() string {
+	// Names that works for both HTTP and GRPC, instead of HTTP or GRPC specific codes.
+	if e == OK {
+		return "OK"
+	}
+	if e == NOT_FOUND {
+		return "NotFound"
+	}
+	return "other"
+}
+
+func (e Kind) String() string {
+	if e == AC {
+		return "AC"
+	}
+	if e == CAS {
+		return "CAS"
+	}
+	return "other"
+}
+
+type Metrics interface {
+	// TODO Document interface
+	IncomingRequestCompleted(kind Kind, method Method, status Status, headers map[string][]string, protocol Protocol)
+}
+
+type metrics struct {
+	categoryValues               map[string]map[string]struct{}
+	counterIncomingCompletedReqs *prometheus.CounterVec
+}
+
+func NewMetrics(config *config.Metrics) Metrics {
+
+	labels := []string{"method", "status", "kind"}
+	categoryValues := make(map[string]map[string]struct{})
+
+	if config != nil && config.Categories != nil {
+		for categoryName, whiteListedValues := range config.Categories {
+			// Normalize to lower case since canonical for gRPC headers
+			// and convention for prometheus.
+			categoryName := strings.ToLower(categoryName)
+
+			// Store white listed category values as set for efficient access
+			whiteListedSet := make(map[string]struct{})
+			for _, categoryValue := range whiteListedValues {
+				whiteListedSet[categoryValue] = struct{}{}
+			}
+			categoryValues[categoryName] = whiteListedSet
+
+			// Construct a prometheus label for each category.
+			// Prometheus does not allow changing set of
+			// labels until next time bazel-remote is
+			// restarted.
+			labels = append(labels, categoryName)
+		}
+	}
+
+	// For now we only count AC requests, and only the most common status codes,
+	// becuse:
+	//
+	//  - No identified use case for others.
+	//  - Limit number of prometheus time series (if many configured categories).
+	//  - Reduce performance overhead of counters (if many configured categories).
+	//  - Would otherwise require injecting invocations in more places.
+	//
+	// But the naming, and the labels, of the counter, are generic to allow
+	// counting additional requests types or status codes in the future. Without
+	// having to rename the counter and get issues with non continous history of
+	// metrics.
+
+	counterIncomingCompletedReqs := promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "bazel_remote_incoming_requests_completed_total",
+			Help: "The number of incoming HTTP and gRPC request. Currently only AC requests",
+		},
+		labels)
+
+	m := &metrics{
+		categoryValues:               categoryValues,
+		counterIncomingCompletedReqs: counterIncomingCompletedReqs,
+	}
+	return m
+}
+
+func getLabelValueFromHeaderValues(headerValues []string, whiteListedValues map[string]struct{}) string {
+	for _, headerValue := range headerValues {
+		// Prometheus only allows one value per label.
+		// Pick the first white listed header value we find.
+		if _, ok := whiteListedValues[headerValue]; ok {
+			return headerValue
+		}
+	}
+
+	// The values found in the header has not been white listed in
+	// the configuration file. Represent them as "other".
+	//
+	// The white listening is an attempt to avoid polluting
+	// prometheus with too many different time series.
+	//
+	// https://prometheus.io/docs/practices/naming/ warns about:
+	//
+	//   "CAUTION: Remember that every unique combination of key-value
+	//   label pairs represents a new time series, which can dramatically
+	//   increase the amount of data stored. Do not use labels to store
+	//   dimensions with high cardinality (many different label values),
+	//   such as user IDs, email addresses, or other unbounded sets of
+	//   values."
+	//
+	// It would have been nice if bazel-remote could reload the set
+	// of white listed values from updated configuration file, by
+	// SIGHUP signal instead of having to restart bazel-remote.
+	return "other"
+}
+
+func (m *metrics) IncomingRequestCompleted(kind Kind, method Method, status Status, headers map[string][]string, protocol Protocol) {
+	labels := make(prometheus.Labels)
+	labels["method"] = method.String()
+	labels["status"] = status.String()
+	labels["kind"] = kind.String()
+	for labelName := range m.categoryValues {
+		// The canonical form of gRPC and HTTP/2 headers is lowercase "category"
+		headerName := labelName
+		if protocol == HTTP {
+			// but the golang http library is normalizing HTTP/1.1 headers as "Category".
+			headerName = strings.Title(headerName)
+		}
+		if headerValues, ok := headers[headerName]; ok {
+			labels[labelName] = getLabelValueFromHeaderValues(headerValues, m.categoryValues[labelName])
+		} else {
+			labels[labelName] = "" // no header for this label
+		}
+	}
+	m.counterIncomingCompletedReqs.With(labels).Inc()
+}
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -22,6 +22,7 @@ go_library(
         "//config:go_default_library",
         "//server:go_default_library",
         "//utils/idle:go_default_library",
+        "//utils/metrics:go_default_library",
         "//utils/rlimit:go_default_library",
         "@com_github_abbot_go_http_auth//:go_default_library",
         "@com_github_grpc_ecosystem_go_grpc_prometheus//:go_default_library",
--- a/config/config.go
+++ b/config/config.go
@@ -35,6 +35,11 @@ type HTTPBackendConfig struct {
 	BaseURL string `yaml:"url"`
 }
 
+// Metrics stores configuration for prometheus metrics.
+type Metrics struct {
+	Categories map[string][]string `yaml:"categories"`
+}
+
 // Config holds the top-level configuration for bazel-remote.
 type Config struct {
 	Host                        string                    `yaml:"host"`
@@ -55,6 +60,7 @@ type Config struct {
 	DisableGRPCACDepsCheck      bool                      `yaml:"disable_grpc_ac_deps_check"`
 	EnableACKeyInstanceMangling bool                      `yaml:"enable_ac_key_instance_mangling"`
 	EnableEndpointMetrics       bool                      `yaml:"enable_endpoint_metrics"`
+	Metrics                     *Metrics                  `yaml:"metrics"`
 	ExperimentalRemoteAssetAPI  bool                      `yaml:"experimental_remote_asset_api"`
 	HTTPReadTimeout             time.Duration             `yaml:"http_read_timeout"`
 	HTTPWriteTimeout            time.Duration             `yaml:"http_write_timeout"`
@@ -73,6 +79,7 @@ func New(dir string, maxSize int, host string, port int, grpcPort int,
 	disableGRPCACDepsCheck bool,
 	enableACKeyInstanceMangling bool,
 	enableEndpointMetrics bool,
+	metrics *Metrics,
 	experimentalRemoteAssetAPI bool,
 	httpReadTimeout time.Duration,
 	httpWriteTimeout time.Duration) (*Config, error) {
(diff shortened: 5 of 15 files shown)


cache/cache.go
```diff
@@ -21,6 +21,8 @@ const (
 	// used for HTTP when running with the --disable_http_ac_validation
 	// commandline flag.
 	RAW
+
+	UNKNOWN
```
mostynb, line 25: I'm not sure that we need this- if we can't figure out the blob type that's a probably a basic error that needs fixing on the client side, not something the server needs metrics for IMO.

README.md
```diff
@@ -223,6 +223,23 @@ host: localhost
...
+#  categories:
+#    os:
+#      - rhel7
+#      - rhel8
+#      - ubuntu16-04
+#      - ubuntu18-04
+#    branch:
+#      - master
+#    user:
+#      - ci
```
mostynb, line 241: I wonder if it makes more sense to track this on the client side. eg the client makes a build, extracts build statistics, then uploads the data somewhere for tracking/analysis. Fixes generally need to happen on the client side anyway.

mostynb (commented):
I will need to set aside some time to look through this properly (perhaps tomorrow, but more likely monday).

I agree that AC hit rates are more useful than CAS or combined AC+CAS hit rates- I wonder if it would be sufficient to modify the existing metrics in the disk cache layer?

ulrfa:
> I will need to set aside some time to look through this properly (perhaps tomorrow, but more likely monday).

Thanks Mostyn! No problem waiting to monday.

> I agree that AC hit rates are more useful than CAS or combined AC+CAS hit rates- I wonder if it would be sufficient to modify the existing metrics in the disk cache layer?

If the existing counters are modified (also or instead), then consider:

- If validating is enabled, then report AC hit only if all referenced CAS entries are found.
- If counter is incremented in disk.GetValidatedActionResult, then disk.Get should not also increment it for the same request. (One of the reasons to why I choosed to increment from http.go/grpc_ac.go instead of disk.go).
- Also count uploads.
- CAS or CAS+AC statistics might still also be interesting in some rare scenarios? (but I don't look at them myself)

cache/cache.go
ulrfa: I agree it is not relevant for server side metrics and it does not end up in the metrics since only AC is handled. The reason for adding it is only to have something to pass in here: https://github.com/buchgr/bazel-remote/blob/25e244e035a7364a4022187bb7a131e8c4b41c6f/server/http.go#L222-L225

README.md
ulrfa: My reasoning is that:

1. It is convenient to have this information in same prometheus databas as all other cache and remote execution related metrics, to correlate them and include in same dashboard. And since promethues actively fetches, instead of passively receive pushes, a more complex solution with proxy in between, would be needed, if client would push. 
2. Adding --remote_header flags is much easier than somehow extracting build statistics and send separately to a server. I’m not sure if BES could be used, but my company don’t have such infrastructure in place currently.
3. The client can’t provide statistics per cache instance, when there is an hierarcy of caches. E.g. bazel-remote instances on localhosts, as proxy for central bazel-remote instance. (However that use case might also require a proxy, a load balancer, or whatever is between, to add own headers or forward headers from client)

There is a limit on what information could be stored as prometheus labels, without causing too many time series. For some use cases it makes more sense to aggregate information on client side until after build finish, and then push to Elasticsearch, or similar. There are pros and cons with both ways, and I think they can complement each other.

utils/metrics/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+	//   label pairs represents a new time series, which can dramatically
+	//   increase the amount of data stored. Do not use labels to store
+	//   dimensions with high cardinality (many different label values),
+	//   such as user IDs, email addresses, or other unbounded sets of
+	//   values."
+	//
+	// It would have been nice if bazel-remote could reload the set
+	// of white listed values from updated configuration file, by
+	// SIGHUP signal instead of having to restart bazel-remote.
+	return "other"
```
kragniz, line 165: I wonder if this should be a word that gives more of a hint that it's masked a real user defined header value. I'm not sure of a good one, though

cache/cache.go
mostynb: If we handle metrics separately from logging, then we should be able to drop this.

utils/metrics/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+	counterIncomingCompletedReqs *prometheus.CounterVec
+}
+
+func NewMetrics(config *config.Metrics) Metrics {
+
+	labels := []string{"method", "status", "kind"}
+	categoryValues := make(map[string]map[string]struct{})
+
+	if config != nil && config.Categories != nil {
+		for categoryName, whiteListedValues := range config.Categories {
```
mostynb, line 91: Let's try to use the term "allowedValues", it's more inclusive.

utils/metrics/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+			whiteListedSet := make(map[string]struct{})
+			for _, categoryValue := range whiteListedValues {
+				whiteListedSet[categoryValue] = struct{}{}
+			}
+			categoryValues[categoryName] = whiteListedSet
+
+			// Construct a prometheus label for each category.
+			// Prometheus does not allow changing set of
+			// labels until next time bazel-remote is
+			// restarted.
```
mostynb, line 106: Is this a limitation of prometheus, or of bazel-remote?

utils/metrics/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+	//  - Would otherwise require injecting invocations in more places.
+	//
+	// But the naming, and the labels, of the counter, are generic to allow
+	// counting additional requests types or status codes in the future. Without
+	// having to rename the counter and get issues with non continous history of
+	// metrics.
+
+	counterIncomingCompletedReqs := promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "bazel_remote_incoming_requests_completed_total",
```
mostynb, line 126: I guess this should have "ac" or "action_cache" or similar in the name?

utils/metrics/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+	return "other"
+}
+
+type Metrics interface {
+	// TODO Document interface
+	IncomingRequestCompleted(kind Kind, method Method, status Status, headers map[string][]string, protocol Protocol)
+}
+
+type metrics struct {
+	categoryValues               map[string]map[string]struct{}
```
mostynb, line 81: I wonder if we should consider simplifying this, and just use a list of allowed labels? eg `ubuntu1804_ci_trunk` ?

mostynb (commented):
I have two high-level comments that don't have an obvious line to be added:

1) I agree that the AC hit rate is the most interesting metric, especially if we're doing validation/completeness checking. If we're not doing those checks then maybe the AC+CAS hit rate is more interesting? Should we just track both AC and CAS separately for flexibility? That way it might be possible to extract useful data in more scenarios.

2) I think we should break this down into at least a few steps (no guarantees that we'll decide to go forward with each stage):

a) Replace the existing hit rate metrics with just AC or separate AC and CAS metrics as discussed above.
b) Add support for client-specified labels (while doing so in a way that will make the next step easier, eg keeping these in a new/separate config file).
c) Add support for updating the set of allowed labels without restarting bazel-remote. This could get complicated, we should spend some time thinking about this before attempting to implement anything.

ulrfa:
> If we're not doing those checks then maybe the AC+CAS hit rate is more interesting?

Personally I would be most interested in only AC, regardless of validation/completeness checking, since I think it is too complex to draw conclusions from AC+CAS or CAS hit rate. But others might see use cases that I don't realize.

> Should we just track both AC and CAS separately for flexibility? That way it might be possible to extract useful data in more scenarios.

I don't have anything against that. But I suggest not adding client specific labels when counting CAS requests, since I'm afraid of potential performance overhead of that. E.g. there is a huge amount of CAS HEAD requests for remote execution scenarios.

> a) Replace the existing hit rate metrics with just AC or separate AC and CAS metrics as discussed above.

In disk.go or http.go/grpc*.go?

> b) Add support for client-specified labels (while doing so in a way that will make the next step easier, eg keeping these in a new/separate config file).

In disk.go or http.go/grpc*.go?

> c) Add support for updating the set of allowed labels without restarting bazel-remote. This could get complicated, we should spend some time thinking about this before attempting to implement anything.

I agree that it could get complicated and we should wait with c. But I think it should be considered also for some more parmeters, see https://github.com/buchgr/bazel-remote/issues/352. Should all parameters allowed to be reloaded, be placed in a separate file?

utils/metrics/metrics.go
ulrfa: yes, good point!

utils/metrics/metrics.go
ulrfa: It is enforced by the prometheus client golang library, used by bazel-remote.

I’m not sure, but I guess that prometheus server itself might handle it in theory, but perhaps only if the golang library had functionality to force prometheus server to reconnect and reload counters from the client?

utils/metrics/metrics.go
ulrfa: The logic in metrics.go is generic and there is a field called ’kind’ for the prometheus counter. Since http.go only forwards AC, that field will in practice always get the value AC.

My idea was to keep it generic, if someone will se a need to use that counter also for CAS, or something else, in the future. And without then renaming the counter (which would make it harder to see historical trends in prometheus database).

I use an expression like the following, to count all AC requests of method GET or HEAD:
 
`bazel_remote_incoming_requests_completed_total{kind="AC",method=~"GET|HEAD"}`

But I can remove the kind field and instead add AC to the counter name, if you think that is better?

utils/metrics/metrics.go
ulrfa: Isn’t there a risk that would complicate it rather than simplifying?
I.e. introducing special cases like: ”ubuntu18” when only ”ubuntu1804” is allowed, handling underscore in values, etc?

However, I’m certainly not a fan of the golang syntax `map[string]struct{}` for a set of strings...

utils/metrics/metrics.go
ulrfa: Please tell if you find a good word!

I agree that ”other” can be confusing for someone trying to set a header.

But I hope the word ”other” would make sense for most people looking at a dashboard, without knowing that there is a concept of allowed values involved.

utils/metrics/metrics.go
mostynb: If we can't solve this, would reloading labels from the config file be pointless?

utils/metrics/metrics.go
mostynb: I don't quite understand this map, how would it help if a client says ”ubuntu18” instead of ”ubuntu1804” ?

I figure that a list is simpler to implement, and probably easier to understand, at the cost of needing to enumerate all possible labels. But since we don't want too many labels, that number shouldn't be large anyway.

utils/metrics/metrics.go
mostynb: Should this function return an error if the label isn't allowed?

utils/metrics/metrics.go
mostynb: This sounds ok, I didn't spot the "kind".

ulrfa:
I’m elaborating about use cases with examples:

Bazel clients configured to always add flags like:
 
```
--remote_header=user=$USER
--remote_header=host=$HOST
--remote_header=os=\`get_os_name.sh\`
```

And bazel clients invoked via CI can in addition add headers like:
 
```
--remote_header=branch=$BRANCH_NAME
--remote_header=product=$PRODUCT_NAME
--remote_header=pipeline=$CI_PIPELINE_NAME
```

The value set for user and branch is not bounded and could therefore not be stored as is in Prometheus. But in many cases it is good enough to distinguish between if it was ci user or non-ci user, or if it was master or non-master branch. That could be achieved by configuring bazel-remote with:

```
branch:
   - master
user:
   - ci
```

and any non allowed branches or users would get the Prometheus label value “other”.

The number of alternative operating systems (“os” header) is typically bounded, and allowed values could be listed in a bazel-remote configuration file like: 

```
os:
   - rhel7
   - rhel8
   - ubuntu16-04
   - ubuntu18-04
```

From time to time operating system comes and goes, and then the list in the configuration could be updated. It does not happen very frequently, so it might be acceptable having to restart bazel-remote, but it would be preferable if possible without restart. And if there are a few rare ones, it could be OK to have them reported by bazel-remote as “other”.

The set of labels and header names (branch, user, os, …) would typically be more static, than the set of values each of them might have (rhel7, rhel8, …) . Therefore I think it is OK having to restart bazel-remote if the set of header names changes, but preferably not when their allowed set of values changes.

Example of Prometheus queries on the metrics, using the labels.

```
# Total number of AC requests from user ci on rhel8
sum (bazel_remote_incoming_requests_completed_total{kind="AC",user="ci",os="rhel8"})

# Total number of AC requests grouped by user "ci" and "other"
sum by (user) (bazel_remote_incoming_requests_completed_total{kind="AC"})

# Total number of AC requests grouped by: "rhel7", "rhel8", "ubuntu16-04", "ubuntu18-04" and "other"
sum by (os) (bazel_remote_incoming_requests_completed_total{kind="AC"})

# Total number of AC requests grouped operating systems, but only including those performed by user ci.
sum by (os) (bazel_remote_incoming_requests_completed_total{kind="AC",user="ci"})

# Total number of AC cache misses
sum (bazel_remote_incoming_requests_completed_total{kind="AC",method=~"GET|HEAD",status="NotFound"})

# Total number of AC cache misses, grouped per OS
sum by (os) (bazel_remote_incoming_requests_completed_total{kind="AC",method=~"GET|HEAD",status="NotFound"})
```

utils/metrics/metrics.go
ulrfa: Do you mean reject the http request with an 4xx error? I think that would be too strict. See my long comment from today where I’m trying to elaborate on use cases.

utils/metrics/metrics.go
ulrfa: I’m thinking that the set of labels/headers (branch, user, os, …) are typically more static, so it could be OK if those can’t be changed without restarting. But the allowed set of values, that those labels/headers can have, (rhel7, rhel8, ubuntu18-04, …) typically changes more often and would benefit from re-loading without restart.

utils/metrics/metrics.go
ulrfa: I think I missinterpreted your example of "ubuntu1804_ci_trunk" as a string with underscore as separator. That's why I mentioned handling of underscore etc.

categoryValues is a map, with label names/header names (e.g. "os" and "user") as keys. The keys are mapped to values by the map, and each such value is a set of strings, containing allowed values for that label/header (e.g. {"ubuntu18-04", "ubuntu16-04"})

Perhaps categoryValues isn't a good name for such a map?

See also my long comment from today where I’m trying to elaborate on use cases.

Does it make sense?

utils/metrics/metrics.go
mostynb: Ah, OK.

mostynb:
> > If we're not doing those checks then maybe the AC+CAS hit rate is more interesting?
> 
> Personally I would be most interested in only AC, regardless of validation/completeness checking, since I think it is too complex to draw conclusions from AC+CAS or CAS hit rate. But others might see use cases that I don't realize.
> 
> > Should we just track both AC and CAS separately for flexibility? That way it might be possible to extract useful data in more scenarios.
> 
> I don't have anything against that. But I suggest not adding client specific labels when counting CAS requests, since I'm afraid of potential performance overhead of that. E.g. there is a huge amount of CAS HEAD requests for remote execution scenarios.

Let's just report hit rate metrics for AC lookups to begin with.

> > a) Replace the existing hit rate metrics with just AC or separate AC and CAS metrics as discussed above.
> 
> In disk.go or http.go/grpc*.go?

If we're not differentiating between http and grpc cache lookups, I think it probably makes sense to do this in disk.go - unless there's something I'm missing?

> > b) Add support for client-specified labels (while doing so in a way that will make the next step easier, eg keeping these in a new/separate config file).

See above. It might be worth merging this step with (a), depending on the size of the changes.

> > c) Add support for updating the set of allowed labels without restarting bazel-remote. This could get complicated, we should spend some time thinking about this before attempting to implement anything.
> 
> I agree that it could get complicated and we should wait with c. But I think it should be considered also for some more parmeters, see #352. Should all parameters allowed to be reloaded, be placed in a separate file?

Good question. If possible I would prefer an upgrade path that does not require manual updating of config files (unless you want to make use of new features).

Which other parameters would we want to be reloadable? One argument in favour of putting reloadable settings in a separate file is that we don't need to figure out what to do if non-reloadable settings change. I suppose we could just log warnings that those settings have not been reloaded.

ulrfa:
I become inspired by https://github.com/buchgr/bazel-remote/issues/355 and will play around with what happens if disk.go would implement an interface like:

```
type CasAcCache interface {
    Put(kind EntryKind, hash string, size int64, rdr io.Reader, context RequestContext) (error)  // TODO or io.ReadCloser ?
    Get(kind EntryKind, hash string, size int64, context RequestContext) (io.ReadCloser, int64, error)
    Contains(kind EntryKind, hash string, size int64, context RequestContext) (bool, int64)
    GetValidatedActionResult(hash string, context RequestContext) (*pb.ActionResult, []byte, error)
}
```

Then http.go and grpc*.go could use that interface instead of accessing disk.go directly.

And it would be possible to have additional implementations of CasAcCache interface, that can wrap disk.go and handle metrics.

That should avoid the problem about if counters should be placed in disk.go or in http.go/grpc*.go.

And it could allow alternative implementations for non-prometheus metrics also.

RequestContext would provide some way for accessing http/grpc headers.

ulrfa:
> > In disk.go or http.go/grpc*.go?
> 
> If we're not differentiating between http and grpc cache lookups, I think it probably makes sense to do this in disk.go - unless there's something I'm missing?

Issues if placing in disk.go:

- How to avoid incrementing a counter twice for the same request? (incrementing both in disk.Get and disk.GetValidatedActionResult for the same AC request)

- Needs access to gRPC/HTTP headers in disk.go

I tried an alternative implementation in #358, with metrics in separate decorator instead.

> Which other parameters would we want to be reloadable? One argument in favour of putting reloadable settings in a separate file is that we don't need to figure out what to do if non-reloadable settings change. I suppose we could just log warnings that those settings have not been reloaded.

Use cases listed in #352. I'm mainly interested in reloadable settings for:
- metrics configuration (add additional allowed values)
- http proxy settings (url and perhaps others)
- enable/disable access log (no such configuration parameter currently)

Issue #351, opened 2020-09-22
Configure http proxy uploaders


ulrfa:
Allow configuring the constants in httpproxy.go, from configuration file:
 
```
const numUploaders = 100
const maxQueuedUploads = 1000000
``` 
And specifically allow numUploaders to be set to 0. That would mean that bazel-remote will try fetching from parent cache, but never uploading anything to parent cache.

Optionally also allow disabling forwarding of HEAD requests via proxy.

Use cases:

- Set to zero if the local bazel-remote instance mostly contains artifacts irrelevant for the parent cache. But there is chance of cache hits in other direction.

- Set to zero if the executor using the local bazel-remote instance is not trusted enough to populate a central shared cache.

- Set to zero if the remote execution input files uploaded to a specific remote execution cache instance, should not be distributed futher to also central cache instance. But if the remote executor should still get build result cache hits from central cache. (For this use case it could also make sense to disable forwarding of HEAD requests, and only forward GET requests).

- Reduce number of uploaders based on what a parent central cache could handle, to avoid overload conditions. (Not sure if it would work i practice or if a single uploader would still consume almost as much resources)

mostynb:
Making these configurable sounds reasonable. (The other proxy backends have the same settings, it would be good if these could use the same configuration setting instead of duplicating configuration options.)

I'm not so sure about disabling HEAD requests, unless we also disable GET requests. What's the intended use case for this?

Making these modifiable without restarting bazel-remote might be too complex.

ulrfa:
Thanks Mostyn,

Use case for forwarding GET but not HEAD: If there is a separate cache instance for a remote execution cluster, that should fetch build results also from another central cache instances, but the input files (e.g. .c and .h files) uploaded to the remote execution cache, should not be uploaded to the other central cache instance. The FindMissingBlobs requests from bazel client then results in huge amount of HEAD requests that would be unnecessary to forward via the proxy.

I'm not working on a pull request for this now, instead I'm focusing on https://github.com/buchgr/bazel-remote/pull/350

PR #359, opened 2020-09-28
#351 Add support for configuring uploaders and upload queue size.

bdittmer:
I'm opening this to see if this satisfies the needs outlined in #351 and gather feedback. 

If this looks good I'll fix the tests and write new cases.

### DIFF
--- a/cache/httpproxy/httpproxy.go
+++ b/cache/httpproxy/httpproxy.go
@@ -17,9 +17,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
-const numUploaders = 100
-const maxQueuedUploads = 1000000
-
 type uploadReq struct {
 	hash string
 	size int64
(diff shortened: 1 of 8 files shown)


ulrfa (commented):
Thanks @bdittmer! I like this! It would allow me to get rid of one of my internal patches.

cache/httpproxy/httpproxy.go
```diff
@@ -87,26 +84,31 @@ func uploadFile(remote *http.Client, baseURL *url.URL, accessLogger cache.Logger
...
-
-	for i := 0; i < numUploaders; i++ {
-		go func(remote *http.Client, baseURL *url.URL, accessLogger cache.Logger,
-			errorLogger cache.Logger) {
-			for item := range uploadQueue {
-				uploadFile(remote, baseURL, accessLogger, errorLogger, item)
-			}
-		}(remote, baseURL, accessLogger, errorLogger)
-	}
+	errorLogger cache.Logger, numUploaders, maxQueuedUploads int) cache.Proxy {
```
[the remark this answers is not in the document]
bdittmer: I prefer to keep configuration code strictly in the main package and treat the rest of the code like a library. The main package gathers configs, chooses interface implementations and ties them all together. The idea is to keep dependencies between packages very, very light.

Have you considered creating your own main package for bazel-remote? This is our approach for hooking into our internal services (like logging, configuration management). It works pretty well, but I think there is a lot we can do to allow users to drop dependencies they don't actually need

config/config.go
```diff
@@ -12,27 +12,36 @@ import (
...
 // HTTPBackendConfig stores the configuration for a HTTP proxy backend.
 type HTTPBackendConfig struct {
-	BaseURL string `yaml:"url"`
+	BaseURL   string                `yaml:"url"`
+	Uploaders *ProxyUploadersConfig `yaml:"uploaders"`
+}
+
+// ProxyUploadersConfig stores the configuration for uploaders and queue size in a Proxy backend.
+type ProxyUploadersConfig struct {
+	Num       int `yaml:"num_uploaders"`
```
mostynb, line 43: I wonder if we should just make these top-level configuration settings since they're identical for each proxy backend? Something like `num_proxy_uploaders` and `max_proxy_queue_size` might be reasonable names.

cache/httpproxy/httpproxy.go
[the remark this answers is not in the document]
mostynb: It's very interesting to hear about your local forks. Please do open issues for anything that you think might be worth upstreaming.

I prefer to have individual arguments here- there's less chance of breakage due to changes to raw arguments, compared to a struct.

mostynb (commented):
@bdittmer: I think this is pretty good, please go ahead with the cleanup/tests/etc.

config/config.go
bdittmer: I thought about that, but the benefit of using a struct is we can check if it's nil and provide the default values (unless the yaml package provides something for that) that have been hard coded. Unfortunately the zero values work against us here and I didn't want users to unwittingly turn off uploads. 

Happy to change this though if there's another solution!

mostynb: In `newFromYaml` we pass an empty `Config` to `yaml.Unmarshal`, and values are only set in the `Config` if they exist in the yaml file. Instead of passing an empty `Config` I think we can pass in one with defaults set.

mostynb:
Squashed and landed on the master branch- thanks!

Issue #355, opened 2020-09-24
Metrics interface(s)


bdittmer:
Regarding metrics collection, will you consider a PR that replaces the hard dependency on prometheus with a dependency on one or more metrics interfaces? It will keep prometheus as the default/example implementation. Background:

We have a fork of bazel-remote to wire in our internal services (app config, logging, health checks, etc.). We also try (aspire?) to keep our fork as close to the upstream as possible. This is _somewhat_ manageable with a separate `main` package and the available interfaces.

Internally we don't officially support prometheus. With the existing metrics implementation I think there are two options to avoid modifying upstream code:

1. Enable prometheus and run a sidecar (or goroutine) to periodically collect and transform metrics to our format.
2. Wrap cache handlers and inspect request/responses for basic things like cache hits/misses. I don't know if this is possible with the gRPC server.

I have a general idea what this looks like but want to gauge interest before committing to a PR. Thanks!

mostynb:
Hi @bdittmer.

@ulrfa has proposed some changes to the metrics in #350 - as part of that effort maybe we can move to a metrics interface that would minimise the maintenance cost of your fork? It would be good to have your input on that (both on the interface/config and also the metrics that are being collected). That might give you another option: a local patch to implement your metrics implementation, but with a cleaner integration point.

ulrfa:
Hi @bdittmer,

Would https://github.com/buchgr/bazel-remote/pull/358 be useful for you? It is an alternative implementation of #350.

bdittmer:
Hi @ulrfa!

#358 is very interesting and I like the direction. Interfaces for the AC and CAS caches is very useful and makes layering functionality on top of `disk.Cache` a lot easier...for instance, creating an in-memory cache backed by a `disk.Cache`. I also think `RequestContext` is a solid introduction and can help with routing requests to different `cache.Proxy` backends.

But for metrics, I was thinking of something like this:

```
type MetricsCollector interface {
  Incr(name string, ctx cache.RequestContext)
  Guage(name string, value float64, ctx cache.RequestContext)
}
```

with concrete implementations in different packages (we shouldn't force users to pull in Prometheus dependencies if they don't use Prometheus). `main.go` can call then create a new Prometheus MetricsCollector and pass it around as needed. It's also useful for debugging (and testing) as you can create a logging metrics collector or a test implementation and assert against collected values.

Thoughts?

ulrfa:
[...]

PR #362, opened 2020-10-03
Introduce `metric.Collector` interface

bdittmer:
This commit wraps the standard Prometheus metrics in interfaces. The idea is to support different ways to collect metrics while keeping Prometheus the default. I haven't addressed the gRPC metrics collection yet, but it's on my radar. 

I'm opening this PR to gather feedback and address the ongoing conversation in #355. Happy to discuss anything and everything!


bdittmer:
Cleanup!

PR #363, opened 2020-10-04
Add flags and defaults for new proxy upload config settings

TyBrown:
When using bazel-build with only ENV variables or Flags to configure the app with s3proxy and a build of bazel-remote which includes #359, `NumUploaders` and `MaxQueuedUploads` ends up defaulting to 0, which effectively disables s3proxy uploads entirely.

I believe this is due to the only place where the defaults are configured is in `config.NewFromYamlFile()`, and when you configure using only ENV variables or Flags, the defaults never get set.
```
		if configFile != "" {
			c, err = config.NewFromYamlFile(configFile)
		} else {
```

So this PR adds Flags/ENV Variables, and sets the defaults when not using a configFile as well.


README.md
```diff
@@ -125,6 +125,8 @@ GLOBAL OPTIONS:
    --http_read_timeout value          The HTTP read timeout for a client request in seconds (does not apply to the proxy backends or the profiling endpoint) (default: 0s, ie disabled) [$BAZEL_REMOTE_HTTP_READ_TIMEOUT]
    --http_write_timeout value         The HTTP write timeout for a server response in seconds (does not apply to the proxy backends or the profiling endpoint) (default: 0s, ie disabled) [$BAZEL_REMOTE_HTTP_WRITE_TIMEOUT]
    --htpasswd_file value              Path to a .htpasswd file. This flag is optional. Please read https://httpd.apache.org/docs/2.4/programs/htpasswd.html. [$BAZEL_REMOTE_HTPASSWD_FILE]
+   --max_queued_uploads value         When using S3, sets the maximum number of objects in queue for upload. If the queue is full, uploads will be skipped until the queue has space again. (default: 1000000) [$BAZEL_REMOTE_MAX_QUEUED_UPLOADS]
+   --num_uploaders value              When using S3, sets the number of Goroutines to process parallel uploads to S3. (default: 100) [$BAZEL_REMOTE_NUM_UPLOADERS]
```
mostynb, line 129: s/S3/proxy backends/

main.go
```diff
@@ -134,6 +134,18 @@ func main() {
 			Usage:   "Path to a .htpasswd file. This flag is optional. Please read https://httpd.apache.org/docs/2.4/programs/htpasswd.html.",
 			EnvVars: []string{"BAZEL_REMOTE_HTPASSWD_FILE"},
 		},
+		&cli.IntFlag{
+			Name:    "max_queued_uploads",
+			Value:   1000000,
+			Usage:   "When using S3, sets the maximum number of objects in queue for upload. If the queue is full, uploads will be skipped until the queue has space again.",
```
mostynb, line 140: As mentioned earlier, refer to "proxy backends".

main.go
```diff
@@ -134,6 +134,18 @@ func main() {
...
+			EnvVars: []string{"BAZEL_REMOTE_MAX_QUEUED_UPLOADS"},
+		},
+		&cli.IntFlag{
+			Name:    "num_uploaders",
+			Value:   100,
+			Usage:   "When using S3, sets the number of Goroutines to process parallel uploads to S3.",
```
mostynb, line 146: How about: "When using a proxy backend, sets the number of upload threads." ?

mostynb (commented):
Thanks for the fix!

README.md
TyBrown: Done

mostynb (approved):
Looks good- thanks again.

PR #368, opened 2020-10-15
import REAPI generated bindings

mostynb:
This is in preparation for some prototyping work.

The bindings were generated from `c820ad5347a6cea95674676aeb74151b768440e7` in the https://github.com/bazelbuild/remote-apis.git repository using the `hooks/pre-commit` script. `github.com/bazelbuild/remote-apis` imports in the resulting files were modified to refer to this repository instead.


HeyiSun:
I'm curious why these generated bindings are committed. 
This brings the duplicate definition error when I import packages from buchgr/bazel-remote. Because my project import the remote-api/remote-execution.proto.

https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict

mostynb:
This was done to make it easier to work on REAPI protocol changes (and IIRC to try and avoid having bazel compile protobuf by not importing the .proto files).

bazel-remote is intended to be a standalone binary- it's not really designed to be imported as a module, though I have accepted some changes in the past to make it possible.

Which packages are you trying to use? If they are useful by themselves I might consider setting up a separate repository for them.

PR #388, opened 2021-01-21
Specify oauth scopes for service account keys

tomgr:
When using bazel-remote as a proxy for a GCS bucket, and using a json credentials file for a service account, requests to access the GCS bucket fail with errors such as:
```
Get "https://storage.googleapis.com/...":
oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_scope","error_description":"Invalid OAuth scope or ID token audience provided."}
```
These errors are fixed by specifying the desired oauth scopes when reading the JSON service key.

### DIFF
--- a/cache/gcsproxy/gcsproxy.go
+++ b/cache/gcsproxy/gcsproxy.go
@@ -33,7 +33,8 @@ func New(bucket string, useDefaultCredentials bool, jsonCredentialsFile string,
 			err = fmt.Errorf("Failed to read Google Credentials file '%s': %v", jsonCredentialsFile, err)
 			return nil, err
 		}
-		config, err := google.CredentialsFromJSON(oauth2.NoContext, jsonConfig)
+		config, err := google.CredentialsFromJSON(oauth2.NoContext, jsonConfig,
+			"https://www.googleapis.com/auth/devstorage.read_write")
 		if err != nil {
 			err = fmt.Errorf("The provided Google Credentials file '%s' couldn't be parsed: %v",
 				jsonCredentialsFile, err)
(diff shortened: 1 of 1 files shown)


cache/gcsproxy/gcsproxy.go
```diff
@@ -33,7 +33,8 @@ func New(bucket string, useDefaultCredentials bool, jsonCredentialsFile string,
 			err = fmt.Errorf("Failed to read Google Credentials file '%s': %v", jsonCredentialsFile, err)
 			return nil, err
 		}
-		config, err := google.CredentialsFromJSON(oauth2.NoContext, jsonConfig)
+		config, err := google.CredentialsFromJSON(oauth2.NoContext, jsonConfig,
+			"https://www.googleapis.com/auth/cloud-platform")
```
mostynb, line 37: Hi- thanks for the patch.

I'm not so familiar with GCS, so I wonder if you could help me to understand this a little.
* Could this break use of GCS with non-service accounts? If not, should we add a configuration setting instead of hardcoding this?
* It seems that there's a https://www.googleapis.com/auth/devstorage.read_write scope which might be more limited, would this be better?

tomgr: Hi. Thanks for the comments. You're quite right about using the restricted scope: I've modified the patch to do so, and I've also modified the scopes requested in the application default credentials which ought to just be the devstorage ones too, looking at the operations being performed. I've manually checked this works on my machine using application default credentials and service accounts.

This shouldn't break anyone's usage: the purpose of the scopes parameter is to restrict the scopes of any tokens created by the application. Providing bazel-remote only performs operations that require the devstorage.read_write scope there shouldn't be a problems.

Looking at some other applications it looks like they've made similar fixes. For example see https://github.com/nytimes/gcs-helper/commit/31da8df9ba6b734cb76705f9110b18b57bea62ff.

mostynb:
Looks good- thanks for the contribution.

PR #390, opened 2021-01-23
Add new endpoint_metrics_duration_buckets config option

### LINKED ISSUE #389: Configure http request duration buckets for lower latency caches
Some smaller bucket sizes should be added here to cover lower latency caches: https://github.com/buchgr/bazel-remote/blob/b42d315f3763f9d365c6fd4119e22e66ea252d20/main.go#L50

Here's some numbers from a cache deployment with 4 replicas of bazel-remote (all replicas restarted in the last 24h):

Taking a look at the bucket counters themselves:
```
sum(http_request_duration_seconds_bucket{service="bazel-remote", endpoint!="metrics"}) by (le)
```

```
{le="0.5"}	30528468 # 30528468
{le="1"}	30541059 # 12591
{le="2.5"}	30543574 # 2515
{le="5"}	30543596 # 22
{le="10"}	30543605 # 9
{le="20"}	30543605
{le="40"}	30543605
{le="80"}	30543605
{le="160"}	30543605
{le="320"}	30543605
{le="+Inf"}	30543605
```

Which shows that around 99.9% of request durations are 0.5 or lower (histogram buckets are cumulative counts).

Calculating the mean request duration:

```
sum(http_request_duration_seconds_sum{service="bazel-remote", endpoint!="metrics"})
  /
sum(http_request_duration_seconds_count{service="bazel-remote", endpoint!="metrics"})
```

Returns 0.00845, which is a couple of orders of magnitude lower than the lowest bucket duration.

---

We can do one of:

1. Add some extra smaller buckets at the lower end (increases the cardinality again)
2. Keep the number of buckets the same, but reduce the smallest few
3. Make the duration configurable

Making durations configurable is possibly the best option, since altering bucket sizes makes querying old data less useful.

fwiw the default prometheus go client range looks like it would be better for my deployment at least: https://github.com/prometheus/client_golang/blob/b7799362e0ac323f658fb8d52c2d6df001cf272c/prometheus/histogram.go#L63

kragniz:
This allows people to configure a more suitable histogram range for their deployment. If the option is not specified, the default remains the same as the current values.

Fixes #389.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"sort"
 	"time"
 
 	yaml "gopkg.in/yaml.v2"
@@ -58,10 +59,13 @@ type Config struct {
 	DisableGRPCACDepsCheck      bool                      `yaml:"disable_grpc_ac_deps_check"`
 	EnableACKeyInstanceMangling bool                      `yaml:"enable_ac_key_instance_mangling"`
 	EnableEndpointMetrics       bool                      `yaml:"enable_endpoint_metrics"`
+	MetricsDurationBuckets      []float64                 `yaml:"endpoint_metrics_duration_buckets"`
 	ExperimentalRemoteAssetAPI  bool                      `yaml:"experimental_remote_asset_api"`
 	HTTPReadTimeout             time.Duration             `yaml:"http_read_timeout"`
 	HTTPWriteTimeout            time.Duration             `yaml:"http_write_timeout"`
 }
+
+var defaultDurationBuckets = []float64{.5, 1, 2.5, 5, 10, 20, 40, 80, 160, 320}
 
 // New returns a validated Config with the specified values, and an error
 // if there were any problems with the validation.
@@ -104,6 +108,7 @@ func New(dir string, maxSize int, host string, port int, grpcPort int,
 		DisableGRPCACDepsCheck:      disableGRPCACDepsCheck,
 		EnableACKeyInstanceMangling: enableACKeyInstanceMangling,
 		EnableEndpointMetrics:       enableEndpointMetrics,
+		MetricsDurationBuckets:      defaultDurationBuckets,
 		ExperimentalRemoteAssetAPI:  experimentalRemoteAssetAPI,
 		HTTPReadTimeout:             httpReadTimeout,
 		HTTPWriteTimeout:            httpWriteTimeout,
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -58,6 +59,7 @@ type Config struct {
 	DisableGRPCACDepsCheck      bool                      `yaml:"disable_grpc_ac_deps_check"`
 	EnableACKeyInstanceMangling bool                      `yaml:"enable_ac_key_instance_mangling"`
 	EnableEndpointMetrics       bool                      `yaml:"enable_endpoint_metrics"`
+	MetricsDurationBuckets      []float64                 `yaml:"endpoint_metrics_duration_buckets"`
```
kragniz, line 62: I'm open for suggestions to give this another name

config/config.go
```diff
@@ -211,5 +214,20 @@ func validateConfig(c *Config) error {
 		}
 	}
 
+	if c.MetricsDurationBuckets != nil {
+		if !sort.Float64sAreSorted(c.MetricsDurationBuckets) {
+			return errors.New("'endpoint_metrics_duration_buckets' must be in increasing order")
```
mostynb, line 219: Why not just sort them?

main.go
```diff
@@ -432,6 +429,14 @@ func main() {
 		}
 		log.Println("Mangling non-empty instance names with AC keys:", acKeyManglingStatus)
 
+		// durationBuckets is the buckets used for Prometheus histograms in seconds.
+		var durationBuckets []float64
+		if c.MetricsDurationBuckets != nil {
+			durationBuckets = c.MetricsDurationBuckets
+		} else {
+			durationBuckets = []float64{.5, 1, 2.5, 5, 10, 20, 40, 80, 160, 320}
```
mostynb, line 437: Let's move the defaults into config.go

config/config.go
mostynb: Would `metrics_duration_buckets` be specific enough?

kragniz: I think that depends on the likeliness of other future histograms tracking non-endpoint processes that will have a different enough distribution of values that they can't sensibly share the same config.

Currently it's also following the naming of the `enable_endpoint_metrics` option

config/config.go
kragniz: Good idea, I've done this in config now

main.go
kragniz: Done

config/config.go
mostynb: OK, let's stick with `endpoint_metrics_duration_buckets`.

mostynb:
Merged- thanks for the contribution.

kragniz:
Thanks!

ulrfa:
Thanks @kragniz and @mostynb!

PR #398, opened 2021-03-03
[disk] improve disk usage estimates by rounding file sizes up to 4kb blocks

### LINKED ISSUE #397: use block-based disk usage accounting
We currently use a simple sum of the file sizes of each blob in the cache to estimate the total disk usage. This underestimates the actual disk usage, since most filesystems use 4Kb blocks. We can get a better estimate from the `syscall.Stat_t` struct's `Blocks` field, which is a count of the number of 512 byte blocks used by the file.

mostynb:
Before this change, we approximated disk usage by keeping a running total of the file lengths of every blob in the index. But most filesystems store files in sequences of fixed size blocks, typically with size 4kb, so our estimate of filesystem space usage has been significantly lower than the actual usage.

It we round file sizes up to the nearest 4kb boundary for the disk usage accounting, then we would have a better estimate. An even better estimate would check the number of 512 byte blocks reported by stat(2), but that would require an extra syscall after writing every file- so let's try this simple approximation first.

Resolves #397.

### DIFF
--- a/cache/disk/disk_test.go
+++ b/cache/disk/disk_test.go
@@ -46,7 +46,7 @@ func TestCacheBasics(t *testing.T) {
 	itemSize := int64(256)
 
 	// Add some overhead for likely CAS blob storage expansion.
-	cacheSize := int64(itemSize * 2)
+	cacheSize := int64(itemSize*2 + BlockSize)
 
 	testCache, err := New(cacheDir, cacheSize, "zstd", nil)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ulrfa:
Thanks @mostynb 

I always configure the max cache size to lower than 100% of the file system capacity, as an attempt to reduce file system fragmentation. So personally I have plenty of marginal for approximations or occasionally exceeding the threshold. In other words, I prefer this approximation rather than an extra stat syscall after each write.

mostynb:
Yeah, hopefully this approximation is good enough for most installs.

After this change lands your actual disk usage should drop, so you might want to update the cache size limit.

PR #405, opened 2021-03-13
Fix race: Consume entire reader in WriteAndClose

glukasiknuro:
Noticed error on bazel side:
```
WARNING: Some artifacts failed be uploaded to the remote cache.
WARNING: Writing to Remote Cache:
BulkTransferException
```
And multiple errors on remote-cache side:
```
2021/03/13 01:57:49 GRPC BYTESTREAM WRITE INTERNAL ERROR instance/uploads/0098b6ad-4ecb-40c0-9348-622afbd38e0d/blobs/12593581821b68a00762613129d0998450759fc36c64014d9644bd6f895aeaff/302
```

The error leads to: https://github.com/buchgr/bazel-remote/blob/f0b31efdc1c6f0aa0d6db63443d54abbafbcb7d8/server/grpc_bytestream.go#L530

which can happen if `nil` err to `putResult` is sent before `recvResult` is closed, which can happen if this goroutine:
https://github.com/buchgr/bazel-remote/blob/f0b31efdc1c6f0aa0d6db63443d54abbafbcb7d8/server/grpc_bytestream.go#L432
finishes before recv stream is closed.

Will try to add test case slightly later.

### DIFF
--- a/cache/disk/casblob/casblob.go
+++ b/cache/disk/casblob/casblob.go
@@ -600,7 +600,7 @@ func WriteAndClose(r io.Reader, f *os.File, t CompressionType, hash string, size
 	}
 	h.chunkOffsets[nextChunk] = fileOffset
 
-	// Wait for EOF from the reader.
+	// Confirm that there is no data left to be read.
 	bytesAfter, err := io.ReadFull(r, uncompressedChunk)
 	if err == nil {
 		return -1, fmt.Errorf("expected %d bytes but got at least %d more", size, bytesAfter)
--- a/cache/disk/casblob/casblob.go
+++ b/cache/disk/casblob/casblob.go
@@ -600,6 +600,14 @@ func WriteAndClose(r io.Reader, f *os.File, t CompressionType, hash string, size
 	}
 	h.chunkOffsets[nextChunk] = fileOffset
 
+    // Wait for EOF from the reader.
+    bytesAfter, err := io.ReadFull(r, uncompressedChunk)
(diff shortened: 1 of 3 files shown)


cache/disk/casblob/casblob.go
```diff
@@ -600,6 +600,14 @@ func WriteAndClose(r io.Reader, f *os.File, t CompressionType, hash string, size
 	}
 	h.chunkOffsets[nextChunk] = fileOffset
 
+	// Wait for EOF from the reader.
```
mostynb, line 603: I would say "confirm that there's no data left to be read".

mostynb (commented):
Thanks for the fix!

Is this caused by bad input, when the client writes too much data? Or can this happen with valid input too?

glukasiknuro:
> Thanks for the fix!
> 
> Is this caused by bad input, when the client writes too much data? Or can this happen with valid input too?

Note, that I have seen those errors consistently for some time, this fixed it.

This happens on valid input too. There are two scenarios at least:

1. client sends last request with no data and Finish:true. This one is valid and easily reproducible, however not sure whether clients are behaving this way. Added a test case for it.
2. client sends last chunk of data, the Put method in disk_cache returns immediately, and there is a race between an item being put to putResult and recvResult. This is slightly harder to reproduce in unit tests, but may happen in real world.

mostynb:
Looks good to me, squashed and pushed to master.

Thanks again!

Issue #409, opened 2021-03-20
Downgrading from HEAD to 1.3.1 and upgrading back causes issues


glukasiknuro:
Got this error:
```
Loading of existing cache entries failed due to error: Unrecognized file: "cas.v2/42/42cc92fe15523ebfd3019a553ba8b068e51f2c63c2a37aa37865db776ad44da9.v1"
```
Possibly some migration code artifact?

glukasiknuro:
Actually, in general migration to v2 maybe broken here?

https://github.com/buchgr/bazel-remote/blob/6aa66a2b6098fb38823990d16915f05b12d1bd18/cache/disk/disk.go#L333

while later requiring size to be present in file name here:
https://github.com/buchgr/bazel-remote/blob/6aa66a2b6098fb38823990d16915f05b12d1bd18/cache/disk/disk.go#L374

Will try to create a fix.

mostynb:
Thanks for the fix- merged.

Note that in general bazel-remote does not support downgrading between major or minor versions. HEAD is basically v2.0.0 at the moment, I'm just waiting a bit for bugs to appear before making the release (thanks again for your help with this).

glukasiknuro:
> Note that in general bazel-remote does not support downgrading between major or minor versions.

Yeah I would be very cautions before doing downgrades, although the crash was pretty unexpected. This happened while testing on staging instance. Thanks!

PR #410, opened 2021-03-20
Fix v1->v2 cas migration

glukasiknuro:
See https://github.com/buchgr/bazel-remote/issues/409


mostynb (approved):
Looks good to me- thanks for the fix!

Issue #420, opened 2021-04-06
Changing from v1.3.1 to v2.0.0 breaks builds with AWS S3 storage backend


lukedirtwalker:
After the switch to version 2.0.0 of the bazel-remote image our builds started failing.
Example error from bazel:
```
(03:06:37) ERROR: An error occurred during the fetch of repository 'io_bazel_rules_go':
   Traceback (most recent call last):
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/2c5758fdd18438639bdb6a4b4f0b5e90/external/bazel_tools/tools/build_defs/repo/http.bzl", line 111, column 45, in _http_archive_impl
		download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Invalid size (must be non-negative): -1
```
We use `--experimental_remote_asset_api`.

From the logs of the bazel-remote I don't get any more insights: (the log is
```
Using S3 backend.
2021/04/05 03:06:16 GRPC GETCAPABILITIES
2021/04/05 03:06:17 S3 CONTAINS buildkite-bazel-cache-anapaya cas.v2/ec/ece669d52998c7a0df2c2380f37edbf4ed8ebb1a03587ed1781dfbececef9b3d OK
2021/04/05 03:06:17 S3 DOWNLOAD buildkite-bazel-cache-anapaya cas.v2/ec/ece669d52998c7a0df2c2380f37edbf4ed8ebb1a03587ed1781dfbececef9b3d OK
2021/04/05 03:06:17 GRPC BYTESTREAM READ: Invalid size (must be non-negative): -1
2021/04/05 03:06:37 GRPC GETCAPABILITIES
2021/04/05 03:06:37 S3 DOWNLOAD buildkite-bazel-cache-anapaya cas.v2/ec/ece669d52998c7a0df2c2380f37edbf4ed8ebb1a03587ed1781dfbececef9b3d OK
2021/04/05 03:06:37 GRPC BYTESTREAM READ COMPLETED blobs/ece669d52998c7a0df2c2380f37edbf4ed8ebb1a03587ed1781dfbececef9b3d/9356
2021/04/05 03:06:37 S3 CONTAINS buildkite-bazel-cache-anapaya cas.v2/6f/6f111c57fd50baf5b8ee9d63024874dd2a014b069426156c55adbf6d3d22cb7b OK
[...]

PR #421, opened 2021-04-06
add more context in bytestream resource name parsing error logs

mostynb:
This might help to debug issues like #420.


mostynb:
Merged to the master branch, we want these changes even if they don't fix this particular issue.

mostynb:
Pushed a new "latest" image to docker hub with this change.

PR #423, opened 2021-04-08
[s3] remove key_version flag

lukedirtwalker:
With the recent introduction of the compression mode, the flag was no
longer used. This seems to have been a breaking change from version 1 to
2. There is no longer the possibility to have key version 2 for
uncompressed mode in S3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,10 +239,6 @@ OPTIONS:
    --s3.region value The AWS region. Required when not specifying S3/minio
       access keys. [$BAZEL_REMOTE_S3_REGION]
 
-   --s3.key_version value Set to 1 for the legacy flat key format, or 2 for
-      the newer format that reduces the impact of S3 rate limits. (default: 1)
-      [$BAZEL_REMOTE_S3_KEY_VERSION]
-
    --disable_http_ac_validation Whether to disable ActionResult validation
       for HTTP requests. (default: false, ie enable validation)
       [$BAZEL_REMOTE_DISABLE_HTTP_AC_VALIDATION]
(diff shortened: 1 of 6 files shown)


lukedirtwalker:
Just realized that this flag is no longer used. Not sure what the best way to deal with it is, but this is one way ;)

utils/flags/flags.go
```diff
@@ -199,7 +199,7 @@ func GetCliFlags() []cli.Flag {
 		},
 		&cli.IntFlag{
 			Name:        "s3.key_version",
-			Usage:       "Set to 1 for the legacy flat key format, or 2 for the newer format that reduces the impact of S3 rate limits.",
+			Usage:       "DEPRECATED. Starting from version 2 of bazel-remote enabling compression will automatically enable key version 2. The flag will be removed.",
```
mostynb, line 202: If we leave this flag, I think we should log a warning on startup if any values other than 2 are provided.

README.md
```diff
@@ -239,10 +239,6 @@ OPTIONS:
    --s3.region value The AWS region. Required when not specifying S3/minio
       access keys. [$BAZEL_REMOTE_S3_REGION]
 
-   --s3.key_version value Set to 1 for the legacy flat key format, or 2 for
-      the newer format that reduces the impact of S3 rate limits. (default: 1)
-      [$BAZEL_REMOTE_S3_KEY_VERSION]
-
```
mostynb, line 245: Could you regenerate the help text, so the "DEPRECATED" note appears here? I used `COLUMNS=80 ./bazel-remote -h` to do this last time.

utils/flags/flags.go
```diff
@@ -199,7 +199,7 @@ func GetCliFlags() []cli.Flag {
...
 			Value:       1,
 			DefaultText: "1",
```
mostynb, line 204: The defaults should be updated too.

mostynb:
Merged a slightly modified version of this. Thanks for the cleanup!

lukedirtwalker:
Thanks for merging, I was on Holidays ;)

Issue #408, opened 2021-03-19
Reduce excessive logging


glukasiknuro:
Bazel-remote is pretty verbose logging all access information.

In particular it becomes problematic for `findMissingBlobs` call, which may have thousands of entries, and each entry cases a single line to be logged here: https://github.com/buchgr/bazel-remote/blob/6aa66a2b6098fb38823990d16915f05b12d1bd18/server/grpc_cas.go#L45

Note, the bazel calls this for all action it processes, causing huge amount of logs - in our case it caused some issues on kubernetes when processing was stalled.

Workaround was to just remove the most problematic logs, but possibly logging could be improved - e.g. in this case just don't log all singular CAS HEAD requests but a summary of those for single findMissingBlobs.

In general maybe accessLogger could log into limited list of entries, and this list could be accessible through some web-point on demand instead of logging those to stdout.

mostynb:
Can you describe the issues you encountered? Was there a bottleneck in bazel-remote, or did it simply flood your log daemon?

glukasiknuro:
Symptom was that bazel-remote started to work very slowly after a few hours of working. Note, that it's used as a backed in build-barn setup.

The actual root cause is not very well understood by me - whether it was just that bazel-remote become very slow because of logging, or it triggered some threshold on docker / kubernetes side. One thing that was clear was that checking logs I was getting at least 5M logs for each second - actually because of log trimming was not able to get more than that, so it could be even more than 5M/sec. After disabling some logs it just came back to normal and had no issues for the following days.

glukasiknuro:
Missed one important detail: when investigating the issue, found out in bazel stack traces that bazel client was hanging on `findMissingBlobs`. Reproducing the RPC call manually, was working very long - like for reduced call from 2k to 100 entries it was still taking 20 sec. This lead to above investigation.

Vanilla 1.3 bazel remote was used (with no proxy).

While checking it, also considered changing locking around cache.Contains - currently mutex is hold for each call of contains, but this call should be fast, so having single lock hold for whole findMissingBlob could be also beneficial - although it's speculative at this point, need to do some measurements.

Issue #352, opened 2020-09-22
Reload config file without restart


ulrfa:
Sometimes I would like to change bazel-remote’s configuration. But restarting bazel-remote would interfer with ongoing builds.

It would be nice if a configuration file could be re-loaded by a SIGHUP signal.

Allowing any configuration parameter to change anytime, might be too complicated, but perhaps support re-loading a few specific parameters? And log a warning if other parameters also changed in configuration file?

Use cases:

- Enable/disable access log. Today I redirect stderr to file, and the access log on stdout to /dev/null, since the later produces too much data. But ocassionally I would like to temporarily enable also the access log for a short period of time, without restarting bazel-remote.

- Change prometheus label configuration without restarting bazel-remote (see discussion in https://github.com/buchgr/bazel-remote/pull/350).

- Set, unset and change http proxy url, without restarting bazel-remote. E.g. if the parent cache goes down or is moved to another URL. Or when migrating users from one bazel-remote instance to another, by temporarily configuring the instances as proxies for each other, for smoth migration. (Ongoing remote execution builds fails if input files from client to old instance, is not avilable for remote executor on new instance)

- Change configuration dynamically to reduce proxy load on parent cache instance. If parent cache becomes overloaded. (I'm not sure about if that make sense or not, the other use cases are more important. See https://github.com/buchgr/bazel-remote/issues/351)

mostynb:
We could implement this naively, but that would mean adding more mutexes around using the settings which may change at any time (or using sync/atomic to access a struct).

I wonder if there's a way to avoid that overhead at the cost of some slight downtime when reloading, for example by restarting the http/grpc servers? If that's too difficult then maybe we can implement a fast-reload option: stop accespting requests, dump the index to disk, restart bazel-remote and import the index.

ulrfa:
Thanks Mostyn,

I'm thinking for example allow creating and replacing `cache.Proxy` instances at runtime. And protect the reference to current proxy instance in disk.go with a mutex.

And perhaps in a similar way creating and replacing instances of metrics.Metrics interface: (https://github.com/buchgr/bazel-remote/blob/25e244e035a7364a4022187bb7a131e8c4b41c6f/utils/metrics/metrics.go#L75-L78)

As long as replaced parts have well defined interfaces, and not too much dependencies, I think they could be replaced in runtime, without too much added complexity.

[...]

PR #438, opened 2021-05-12
config: Flag to disable access logger

LINKIWI:
The access logger can be quite noisy when `bazel-remote` is serving thousands of HTTP requests. This PR proposes the addition of a configuration switch to disable the access logger.

The changes:

* Attach the access and error `*log.Logger` instances to the `Config` struct
* Move all logging instantiation logic to a new file `config/logger.go` [+]
* Add a new configuration param `disable_access_log` available both as a CLI flag and in the YAML configuration to disable the access logger

[+] This approach, I think, makes it easier to make further customizations to the logging behavior in the future. For example, this PR doesn't touch the behavior of the error logger, but I suppose there could be a use case for suppressing those messages as well.

[+] This is also required because the current implementation (on master) creates the loggers before the config is read. This change proposes creating the loggers *after* the config is read because we need to have a parsed config to decide how to create the loggers.

Test:

Help text:

```bash
$ bazelisk run //:bazel-remote -- --help
...
   --disable_access_log  Whether to disable the standard output access logger (default: false, ie enable access logging) [$BAZEL_REMOTE_DISABLE_ACCESS_LOG]
```

Default behavior (no change compared to current master):

```bash
$ bazelisk run //:bazel-remote -- --dir /tmp/bazel --max_size 1 --port 8080
...
2021/05/12 00:18:52  GET 200             ::1 /cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

```bash
$ curl -i localhost:8080/cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
HTTP/1.1 200 OK
...
```

Disabled access logger via CLI flag:

```bash
$ bazelisk run //:bazel-remote -- --dir /tmp/bazel --max_size 1 --port 8080 --disable_access_log
```

```bash
$ curl -i localhost:8080/cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
HTTP/1.1 200 OK
...
```

Disabled access logger via config file:

```bash
$ cat config.yaml
dir: /tmp/bazel
max_size: 1
port: 8080
disable_access_log: true
$ bazelisk run //:bazel-remote -- --config_file $(pwd)/config.yaml
```

```bash
$ curl -i localhost:8080/cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
HTTP/1.1 200 OK
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,9 +265,9 @@ OPTIONS:
       asset API implementation. (default: false, ie disable remote asset API)
       [$BAZEL_REMOTE_EXPERIMENTAL_REMOTE_ASSET_API]
 
-   --disable_access_log Whether to disable the standard output access logger.
-      (default: false, ie enable access logging)
-      [$BAZEL_REMOTE_DISABLE_ACCESS_LOG]
+   --access_log_level The access logger verbosity level. If supplied, must
+      be one of "none" or "all". (default: all, ie enable full access logging)
+      [$BAZEL_REMOTE_ACCESS_LOG_LEVEL]
 
    --help, -h  show help (default: false)
 ```
@@ -366,8 +366,8 @@ host: localhost
 # If true, enable experimental remote asset API support:
 #experimental_remote_asset_api: true
 
-# If true, does not print access logs to standard output:
-#disable_access_log: true
+# If supplied, controls the verbosity of the access logger ("none" or "all"):
+#access_log_level: none
 ```
 
 ## Docker
--- a/README.md
+++ b/README.md
@@ -265,6 +265,10 @@ OPTIONS:
       asset API implementation. (default: false, ie disable remote asset API)
       [$BAZEL_REMOTE_EXPERIMENTAL_REMOTE_ASSET_API]
 
+   --disable_access_log Whether to disable the standard output access logger.
+      (default: false, ie enable access logging)
+      [$BAZEL_REMOTE_DISABLE_ACCESS_LOG]
+
    --help, -h  show help (default: false)
 ```
 
@@ -361,6 +365,9 @@ host: localhost
 
 # If true, enable experimental remote asset API support:
 #experimental_remote_asset_api: true
+
+# If true, does not print access logs to standard output:
+#disable_access_log: true
 ```
 
 ## Docker
--- a/config/BUILD.bazel
+++ b/config/BUILD.bazel
@@ -4,6 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = [
         "config.go",
+        "logger.go",
         "proxy.go",
         "tls.go",
     ],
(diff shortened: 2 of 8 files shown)


ulrfa:
I typically redirect the access log on stdout to /dev/null, and log stderr to file.

For me the possibility to enable/disable access log from configuration file, could be useful if combined with support for reloading configuration file via SIGHUP, as discussed in https://github.com/buchgr/bazel-remote/issues/352

mostynb:
Thanks for the PR @LINKIWI.

This is related to #408 (FYI @glukasiknuro).

I like these changes, but I would like to take some time to think through the command line flag/config file setting before merging. In particular I wonder if it would be better to use a more flexible setting than a boolean? Let's discuss some options before implementing them.

One possibility I have been thinking about recently:
`--access-log-level <NONE | MOST | ALL>` - where `MOST` could mean something like "truncate noisy FindMissingBlobs logs but include everything else"

@ulrfa: I have been doing some experimental refactoring that would make #352 a little easier to achieve. But it's still early days.

LINKIWI:
I also like the idea of a leveled access logger. Happy to send a patch to this PR if that's the preferred direction.

In terms of implementation, there's a few ways I think is reasonable:

* Including a numerical level in `Config` and placing `if configured level > None/Most/All/etc { accesslogger.Printf() }` at callsites
* Creating multiple `*log.Logger` instances (one per level), overriding `Write` logic to respect the configured logging level, and modifying callsites to use the appropriately leveled access logger for each call
* Migrating to a third party logging library like zap et al. that handles most of these with built-in abstractions. This approach feels a bit heavyweight and probably outside of the scope of changes I'm comfortable making.

mostynb:
If we were to add `--access-log-level <"none" | "all" (default)>` now, then we would have the option to add a "most" value later. Apart from changing the flag and the config file setting this PR could mostly be kept as it is.

Or we could use integers, but then we should define three levels now: `--access-log-level <0 | 1 | 2 (default)>` (and make 1 and 2 both mean "all" for now).

I think I prefer the string values better (though we could store the value as an integer internally). What does everyone else think?

LINKIWI:
I'm +1 to `--access-log-level <"none" | "all" (default)>`. It leaves room for future extensibility without needing to make major changes to the proposed business logic.

mostynb:
OK- let's go ahead with this.

LINKIWI:
Changes:

* Change flag from `disable_access_logger` to `access_log_level`
* Add a simple check in `validateConfig` that a supported log level is supplied
* Change the logging instantiation business logic to instead discard the output when the level is `"none"`

I didn't create constants for `"none"` and `"all"` since there doesn't seem to be a convention for that (see e.g. `"zstd"`/`"uncompressed"`) but I can do that if needed.

Additional manual verification:

```bash
$ bazelisk run //:bazel-remote -- --dir /tmp/bazel --max_size 1 --port 8080 --access_log_level invalid
...
2021/05/14 18:37:07 bazel-remote built with go1.16.2 from git commit ec16031f6133df98481eb48df8137d1204ba2b63.
'access_log_level' must be set to either "none" or "all"
...
```

mostynb:
Thanks- squashed and added to the master branch.

PR #442, opened 2021-05-19
[grpc] optimise FindMissingBlobs

mostynb:
FindMissingBlobs calls can contain lots of blob hashes to lookup. Instead of locking and unlocking the index once per hash, we can perform the lookups in batches. And instead of querying proxy backends sequentially, we can use a pool of goroutines to make lots of concurrent queries.

For the first part to work, we need to do this in the disk package instead of the server package (where the rest of the gRPC code is).

Relates to #408.

### DIFF
--- a/cache/disk/findmissing.go
+++ b/cache/disk/findmissing.go
@@ -1,0 +1,141 @@
+package disk
+
+import (
+	"context"
+	"sync"
+
+	"github.com/buchgr/bazel-remote/cache"
+
+	pb "github.com/buchgr/bazel-remote/genproto/build/bazel/remote/execution/v2"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type proxyCheck struct {
+	wg     *sync.WaitGroup
+	digest **pb.Digest
+}
+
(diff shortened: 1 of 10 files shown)


mostynb:
This is still a work-in-progress, barely tested. It doesn't offer full logging yet.

cache/disk/findmissing.go
```diff
@@ -0,0 +1,129 @@
...
+	for i := 0; i < numWorkers; i++ {
+		go func() {
+			var ok bool
+			for req := range c.containsQueue {
+				ok, _ = c.proxy.Contains(cache.CAS, (*req.digest).Hash)
+				if ok {
+					c.accessLogger.Printf("GRPC CAS HEAD %s OK", (*req.digest).Hash)
+					// The blob exists on the proxy, remove it from the
+					// list of missing blobs.
+					req.digest = nil
```
mostynb, line 121: Does this set the `**pb.Digest` to nil, or the `*pb.Digest` that it points to, to nil?

mostynb: Fixed, and added a test.

mostynb:
@glukasiknuro: do you have a good stress test for this that you could share?

PR #443, opened 2021-05-24
Add --max_blob_size flag

### LINKED ISSUE #440: Reject/warn/monitor on large objects entering the cache
Our cache deployment has suffered badly from docker images being added by users as inputs to actions. Something like 600MB object then needs to get fetched by CI agents and it overloads the cache with requests.

I'm not exactly sure what the cache could do differently to help. At least showing the largest object, with some hint about the action that generated it.

The feedback loop would result in us making sure these targets are tagged `no-remote-cache` so the actions they produce don't upload.

alexeagle:
This causes both a gRPC and HTTP endpoint to reject blob writes that are larger than the configured size (in MB)

Fixes #440

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -71,7 +71,7 @@ type Config struct {
 	HTTPReadTimeout             time.Duration             `yaml:"http_read_timeout"`
 	HTTPWriteTimeout            time.Duration             `yaml:"http_write_timeout"`
 	AccessLogLevel              string                    `yaml:"access_log_level"`
-	MaxBlobSize					int	                      `yaml:"max_blob_size"`
+	MaxBlobSize                 int64                     `yaml:"max_blob_size"`
 
 	// Fields that are created by combinations of the flags above.
 	ProxyBackend cache.Proxy
@@ -106,7 +106,7 @@ func newFromArgs(dir string, maxSize int, storageMode string,
 	httpReadTimeout time.Duration,
 	httpWriteTimeout time.Duration,
 	accessLogLevel string,
-	maxBlobSize int) (*Config, error) {
+	maxBlobSize int64) (*Config, error) {
 
 	c := Config{
 		Host:                        host,
@@ -384,6 +384,6 @@ func get(ctx *cli.Context) (*Config, error) {
 		ctx.Duration("http_read_timeout"),
 		ctx.Duration("http_write_timeout"),
 		ctx.String("access_log_level"),
-		ctx.Int("max_blob_size"),
+		ctx.Int64("max_blob_size"),
 	)
 }
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math"
 	"os"
 	"sort"
 	"time"
@@ -170,6 +171,7 @@ func newFromYaml(data []byte) (*Config, error) {
 		StorageMode:            "zstd",
 		NumUploaders:           100,
 		MaxQueuedUploads:       1000000,
+		MaxBlobSize:            math.MaxInt64,
 		MetricsDurationBuckets: defaultDurationBuckets,
 		AccessLogLevel:         "all",
 	}
@@ -246,7 +248,7 @@ func validateConfig(c *Config) error {
 		return errors.New("AllowUnauthenticatedReads setting is only available when authentication is enabled")
 	}
 
-	if c.MaxBlobSize <=0 {
+	if c.MaxBlobSize <= 0 {
 		return errors.New("The 'max_blob_size' flag/key must be a positive integer")
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -246,6 +246,10 @@ func validateConfig(c *Config) error {
 		return errors.New("AllowUnauthenticatedReads setting is only available when authentication is enabled")
 	}
 
+	if c.MaxBlobSize <=0 {
+		return errors.New("The 'max_blob_size' flag/key must be a positive integer")
+	}
+
 	if c.GoogleCloudStorage != nil && c.HTTPBackend != nil && c.S3CloudStorage != nil {
 		return errors.New("One can specify at most one proxying backend")
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -71,6 +71,7 @@ type Config struct {
 	HTTPReadTimeout             time.Duration             `yaml:"http_read_timeout"`
 	HTTPWriteTimeout            time.Duration             `yaml:"http_write_timeout"`
 	AccessLogLevel              string                    `yaml:"access_log_level"`
+	MaxBlobSize					int	                      `yaml:"max_blob_size"`
 
 	// Fields that are created by combinations of the flags above.
 	ProxyBackend cache.Proxy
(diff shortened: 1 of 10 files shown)


alexeagle:
note, I didn't add any tests yet, looking for guidance about this being the right shape first (and maybe where you'd expect tests to go)

also I didn't try to address this yet, because I don't really understand the nuance:
> There's a special case to consider- what should we do if the client uploads an ActionCache blob with inlined CAS blobs, with total size greater than the limit, but each individual blob below the size limit? gRPC UpdateActionResult calls are limited to 4M in practice by convention (and that's probably lower than any max_blob_size value that would be set), but HTTP AC uploads can be arbritrarily large.

mostynb:
> also I didn't try to address this yet, because I don't really understand the nuance:
> 
> > There's a special case to consider- what should we do if the client uploads an ActionCache blob with inlined CAS blobs, with total size greater than the limit, but each individual blob below the size limit? gRPC UpdateActionResult calls are limited to 4M in practice by convention (and that's probably lower than any max_blob_size value that would be set), but HTTP AC uploads can be arbritrarily large.

AC blobs can include inlined CAS blob data. This is unlikely to cause problems for gRPC because the practical upper size of a gRPC message is 4M, which is much smaller than a reasonable "max_blob_size" value.

However, there is no practical upper size limit for HTTP uploads, so it is possible for clients to perform an AC upload with arbitrarily large include CAS blobs. So this leaves a question- what if we have `max_blob_size` set to 20M, and we get a HTTP upload with two inlined 15M CAS blobs?

mostynb (commented):
Thanks- this looks like a good start.

I have two main open questions: how to handle inlined blobs, and how to apply the limit to compressed blobs.

utils/flags/flags.go
```diff
@@ -121,6 +121,13 @@ func GetCliFlags() []cli.Flag {
 			Usage:   "When using proxy backends, sets the maximum number of objects in queue for upload. If the queue is full, uploads will be skipped until the queue has space again.",
 			EnvVars: []string{"BAZEL_REMOTE_MAX_QUEUED_UPLOADS"},
 		},
+		&cli.Int64Flag{
+			Name:    "max_blob_size",
+			Value:   0,
+			Usage:   "The maximum size of an individual blob stored in the cache in MiB. Upload requests will fail for any larger blobs.",
```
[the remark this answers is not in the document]
alexeagle: At least for the use case I'm solving for, the issue was the load put on the network by transferring the object, and that caused the remote cache to backlog requests. In that particular case, the wire size is more relevant than the stored size (and the storage limit is already configurable). So I would argue this is just a transfer size limit and would be computed on the compressed size.

server/http.go
```diff
@@ -293,6 +295,14 @@ func (h *httpCache) CacheHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		if (h.maxBlobSize > 0 && contentLength > int64(h.maxBlobSize) * 1024 * 1024) {
```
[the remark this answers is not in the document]
alexeagle: I don't understand the semantics enough about inlining or how the client behaves to have any informed input about what to do, I'm afraid. Should we get some input from someone else?

server/grpc_bytestream.go
```diff
@@ -325,6 +325,11 @@ func (s *grpcServer) parseWriteResource(r string) (string, int64, casblob.Compre
 			status.Errorf(codes.InvalidArgument, "Invalid size (must be non-negative): %d from %q", size, r)
 	}
```
mostynb, line 327: I think we should perform a similar check in `BatchUpdateBlobs` and `UpdateActionResult`, and possibly also return a value other than `0` in `CacheCapabilities.MaxBatchTotalSizeBytes` if `max_blob_size` is a positive number less than 4M.

config/config.go
```diff
@@ -135,6 +137,7 @@ func newFromArgs(dir string, maxSize int, storageMode string,
 		HTTPReadTimeout:             httpReadTimeout,
 		HTTPWriteTimeout:            httpWriteTimeout,
 		AccessLogLevel:              accessLogLevel,
+		MaxBlobSize:                 maxBlobSize,
 	}
 
 	err := validateConfig(&c)
```
mostynb, line 143: Could you please add a check in `validateConfig` that `MaxBlobSize >= 0` ?

utils/flags/flags.go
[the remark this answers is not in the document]
mostynb: This PR checks the logical blob size, not the transfer size (which can be compressed). Which is not ideal but I think would be OK as long as we document it clearly.

server/http.go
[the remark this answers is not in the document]
mostynb: Let's just make this a simple policy, and make the limit apply to:
* The logical size of CAS blobs.
* The total logical size of AC blobs (which may contain inlined CAS blobs).

mostynb (commented):
Thanks- I left some comments inline.

utils/flags/flags.go
```diff
@@ -121,6 +122,13 @@ func GetCliFlags() []cli.Flag {
 			Usage:   "When using proxy backends, sets the maximum number of objects in queue for upload. If the queue is full, uploads will be skipped until the queue has space again.",
 			EnvVars: []string{"BAZEL_REMOTE_MAX_QUEUED_UPLOADS"},
 		},
+		&cli.Int64Flag{
+			Name:        "max_blob_size",
+			Value:       math.MaxInt64,
+			Usage:       "The maximum size of an individual blob stored in the cache in bytes. Upload requests will fail for any larger blobs.",
```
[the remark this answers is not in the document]
alexeagle: improved the docstring to capture this (though no other ones are multi-line so I tried to make it short to be consistent)

server/grpc_test.go
```diff
@@ -1576,6 +1578,8 @@ func TestParseWriteResource(t *testing.T) {
 	s := &grpcServer{
 		accessLogger: testutils.NewSilentLogger(),
 		errorLogger:  testutils.NewSilentLogger(),
+		// max(int64)
+		maxBlobSize: 9223372036854775807,
```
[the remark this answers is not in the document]
alexeagle: I was being consistent with the testdata below, would you like me to change both?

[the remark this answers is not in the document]
mostynb: Yes please.

mostynb:
This looks pretty close- I think you need to add the `MaxBlobSize` default value in `config.newFromYaml` and add some default values to expected configs in config/config_test.go.

Could you also add a commented-out `max_blob_size` with some notes in README.md and add the newly generated --help text?

server/grpc_bytestream.go
mostynb: I think this might be the last part missing before this can land.

mostynb: Acrtually, maybe we can change this a little- if we move the check into `disk.Put` then it would apply to all blobs the same way, regardless of whether they came from the HTTP or gRPC API. The errors might reach the client slightly later, but that doesn't need to be fast since it probably requires human intervention to fix anyway.

This won't affect loading preexisting blobs from the cache (unless we choose to add a check to `disk.loadExistingFiles` too).

alexeagle: that seems like a substantially different PR? I'm not very familiar with the architecture here, but I can try putting a check in that spot and verify that has the behavior I need and send another PR.

I hope you're also considering whether it's economical to lead me through making this change, vs. just making it yourself? I have no preference.

mostynb: Sure, I can finish this off for you.

alexeagle: That's awesome, thank you! I know it's hard to hand-hold contributors who don't know their way around the codebase, and even after that they don't become regular contributors :(

mostynb: No problem.

Here's the reworked version: https://github.com/buchgr/bazel-remote/pull/446

alexeagle:
replaced by #446

PR #446, opened 2021-06-04
Add max blob size flag (second version)

mostynb:


alexeagle (approved):
Thanks! I fetched this PR and locally verified it still works in my test fixture.

mostynb:
> Thanks! I fetched this PR and locally verified it still works in my test fixture.

Thanks for checking.

I pushed another commit that tries to return more precise error codes to both the HTTP and gRPC frontends. If this looks good, I will merge it and make a new 2.1.0 release.

alexeagle:
Verified again at that latest commit. As expected, the bazel client still doesn't give any hint what the error was, but at least the blob size is included in the digest so it's possible to diagnose if you're aware the cache is configured this way.
Thanks again!

mostynb:
Merged, and included in the v2.1.0 release.

PR #456, opened 2021-07-28
Fix prometheus api for grafana.

lanmaoxinqing:
AC validator currently prevent the [prometheus api](https://prometheus.io/docs/prometheus/latest/querying/api/) due to invalid sha256.

### DIFF
--- a/main.go
+++ b/main.go
@@ -143,6 +143,7 @@ func run(ctx *cli.Context) error {
 		})
 		mux.Handle("/metrics", middlewarestd.Handler("metrics", metricsMdlw, promhttp.Handler()))
 		mux.Handle("/status", middlewarestd.Handler("status", metricsMdlw, http.HandlerFunc(h.StatusPageHandler)))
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -143,6 +143,7 @@ func run(ctx *cli.Context) error {
 		})
 		mux.Handle("/metrics", middlewarestd.Handler("metrics", metricsMdlw, promhttp.Handler()))
 		mux.Handle("/status", middlewarestd.Handler("status", metricsMdlw, http.HandlerFunc(h.StatusPageHandler)))
+		mux.Handle("/api/v1/", middlewarestd.Handler("/api/v1/", metricsMdlw, promhttp.Handler()))
```
kragniz, line 146: What's the intended behaviour for this endpoint? This is just serving the prometheus metrics endpoint, which is served under `/metrics` by convention

lanmaoxinqing: Oh sorry I mixed up the prometheus server and the client library, for we deployed `prometheus server`  and `bazel-remote` on the same server. 
We visit both services with the same port through nginx and the request made by grafana is mistoken sended to bazel-remote.
I'll close the PR and sorry again for wasting your time.

PR #472, opened 2021-09-14
[metrics] count hits/misses for each cache entry kind separately

mostynb:
We also count Contains as well as Get requests now.


mostynb:
@ulrfa: how does this look to you? Are you still interested in adding support for labels on top of this?

cache/disk/disk.go
```diff
@@ -75,6 +64,13 @@ type Cache struct {
 	accessLogger  *log.Logger
 	containsQueue chan proxyCheck
 
+	acHit   prometheus.Counter
```
mostynb, line 67: We only use `Inc()`, so we could define an interface with just that.

cache/disk/disk.go
```diff
@@ -115,9 +117,18 @@ func New(dir string, maxSizeBytes int64, opts ...Option) (*Cache, error) {
 		return nil, err
 	}
 
+	noop := &noopCounter{}
```
mostynb, line 120: this could be nil and still work

cache/disk/options.go
```diff
@@ -56,3 +59,36 @@ func WithAccessLogger(logger *log.Logger) Option {
 		return nil
 	}
 }
+
+func WithEndpointMetrics() Option {
+	return func(c *Cache) error {
+		c.acHit = promauto.NewCounter(prometheus.CounterOpts{
```
mostynb, line 65: These could be passed in as arguments, and the disk module would no longer depend on the prometheus library.

ulrfa:
> @ulrfa: how does this look to you? Are you still interested in adding support for labels on top of this?

😄 Thanks for bringing this to life again Mostyn! I will think and write a response tomorrow.

ulrfa:
For the labels, I would need information from headers in the incoming HTTP/gRPC requests, when incrementing counters. Do you have something in mind for that? I notice that you previously added context.Context parameters. Perhaps those could be used? Or perhaps those contexts.Context could simplify https://github.com/buchgr/bazel-remote/pull/358.

ulrfa:
Again, **thanks for working on this Mostyn!** I would love to get rid of my internal patches and their counters. But my counters focus on the result to client as experienced by the user. Similar to this:

```
bazel_remote_incoming_requests_total{kind="AC", method="GET",  status="OK"}       = 9999
bazel_remote_incoming_requests_total{kind="AC", method="GET",  status="NotFound"} = 222
bazel_remote_incoming_requests_total{kind="AC", method="HEAD", status="OK"}       = 333333333
bazel_remote_incoming_requests_total{kind="AC", method="HEAD", status="NotFound"} = 55555
bazel_remote_incoming_requests_total{kind="AC", method="PUT",  status="OK"}       = 44444
bazel_remote_incoming_requests_total{kind="AC", method="PUT",  status="Error"}    = 11
```
The counters above **allow plotting cache hit ratio**, but gives flexibility at same time **also** for:

- Detecting errors.
- Indicating if bazel clients are configured to upload cache result or not.
- Gives strong correlation to if bazel client had to rebuild actions or not. And correlation with action metrics from Buildbarn. 
- Plotting incomming request rate over time. Grouped by label categories such as "CI" vs "User" if adding additional labels.
- Observing bazel client internal behavior, e.g. see effect of bazel's internal algorithm for when to use HEAD requests.
- ...

Here is an example from our dashboard, showing incoming requests rate, with hits in green color and misses in red color. The red parts are a bit hard to see in this scaled down screenshot, but that is also a sign that the amount of cache misses is low... :wink: 
example_dashboard

Parts of the use cases above might be accomplished also with the low-level disk centric counters in this PR. But correlation to overall user experience (e.g. when bazel client has to rebuild or not) would be weaker or less obvious:
- Example 1: Client consider it a cache **hit**, also when bazel-remote has to download via proxy and increases disk **miss** counter.
- Example 2: An AC referencing a missing CAS file is reported to client as **AC miss**, but disk metrics increase **acHit**.
- Example 3: A CAS file downloaded by client **only once**, could increase casHit counter **twice**, since also increased when GetValidatedActionResult calls c.Get().

It could be possible to have similar kind of metrics also for **outgoing** proxy requests, Example:
`bazel_remote_outgoing_requests_total{protocol="s3", kind="AC", method="GET", status="OK"}`
And compare **incoming** and **outgoing** with each other to distinguish disk cache hits and proxy cache hits.

I guess different people have different use cases. What is your view about this?

mostynb:
> For the labels, I would need information from headers in the incoming HTTP/gRPC requests, when incrementing counters. Do you have something in mind for that? I notice that you previously added context.Context parameters. Perhaps those could be used? Or perhaps those contexts.Context could simplify #358.

Yes, I think we could try using the context arg to disk.Get and disk.Contains to pass through these optional labels.

mostynb:
> Again, **thanks for working on this Mostyn!** I would love to get rid of my internal patches and their counters. But my counters focus on the result to client as experienced by the user. Similar to this:
...
> It could be possible to have similar kind of metrics also for **outgoing** proxy requests, Example:
> `bazel_remote_outgoing_requests_total{protocol="s3", kind="AC", method="GET", status="OK"}`
> And compare **incoming** and **outgoing** with each other to distinguish disk cache hits and proxy cache hits.
> 
> I guess different people have different use cases. What is your view about this?

The current state of this PR just counts disk cache hits, mostly because that's how it worked previously and I haven't thought it all the way through yet (I don't use the proxy backends myself). Maybe it would be OK to count proxy hits/misses in the disk layer too (to give overall hit rate), and also count the proxy hits/misses separately.

Or would it make more sense to count disk hits, proxy hits, misses?

ulrfa:
>> * Example 2: An AC referencing a missing CAS file is reported to client as **AC miss**, but disk metrics increase **acHit**.
>> * Example 3: A CAS file downloaded by client **only once**, could increase casHit counter **twice**, since also increased when GetValidatedActionResult calls c.Get().

> The current state of this PR just counts disk cache hits, mostly because that's how it worked previously and I haven't thought it all the way through yet (I don't use the proxy backends myself). Maybe it would be OK to count proxy hits/misses in the disk layer too (to give overall hit rate), and also count the proxy hits/misses separately.

Even if the proxy hit/misses would be included, the example 2 & 3 with side effects related to GetValidatedActionResult remains. I think all those issues could be avoided by instead placing and incrementing counters in a decorator, that wraps around disk.Cache, based on what the public methods of disk.Cache returns, instead of as logic within disk.Cache.

cache/disk/disk.go
```diff
@@ -131,6 +153,16 @@ func New(dir string, maxSizeBytes int64, opts ...Option) (*Cache, error) {
 		}
 	}
 
+	if c.acHit != noop && c.validateAC {
```
mostynb, line 156: TODO: confirm that this interface equality check is correct and valid.

mostynb:
> > > * Example 2: An AC referencing a missing CAS file is reported to client as **AC miss**, but disk metrics increase **acHit**.
> > > * Example 3: A CAS file downloaded by client **only once**, could increase casHit counter **twice**, since also increased when GetValidatedActionResult calls c.Get().
> 
> > The current state of this PR just counts disk cache hits, mostly because that's how it worked previously and I haven't thought it all the way through yet (I don't use the proxy backends myself). Maybe it would be OK to count proxy hits/misses in the disk layer too (to give overall hit rate), and also count the proxy hits/misses separately.
> 
> Even if the proxy hit/misses would be included, the example 2 & 3 with side effects related to GetValidatedActionResult remains. I think all those issues could be avoided by instead placing and incrementing counters in a decorator, that wraps around disk.Cache, based on what the public methods of disk.Cache returns, instead of as logic within disk.Cache.

Right- now I remember. Pushed a fixup commit which might avoid this problem.

ulrfa:
Have you thought about if there are other ways, with less complexity?

Why do you put logic, for which counter to increment, inside disk.go?

mostynb:
> Have you thought about if there are other ways, with less complexity?
> 
> Why do you put logic, for which counter to increment, inside disk.go?

Good questions. I put these counters in disk.go because that was the single point of control that both the http and grpc frontends use, so I didn't need to duplicate logic in both frontends (IIRC, it was a while ago).

Doing this inside the disk package doesn't seem to add too much complexity (and if we put the incrementors in a slice we wouldn't need the getCounters function). I'm not sure if using a decorator instead would be an improvement- at least in the current state of this PR. If we add optional labels then that might tip the scales in favour of decorators.

ulrfa:
Thanks @mostynb for answering questions!

> Doing this inside the disk package doesn't seem to add too much complexity

From my point of view, it adds the following complexity:

- Not intuitive how the ac*Validated counters relates to the normal counters. I had to read the code in detail to understand the relation.
- Exposing the validateAC boolean into disk.go and making assumptions about how GetValidatedActionResult will be invoked from rest of the code base.
- Risk incrementing wrong counter. E.g. forgetting "inc = counters.hit" in some branch when making changhes to the code base. Especially if separating misses vs errors in the future, and even more branches needs consideration.
- Defer functions that becomes "reconfigured".

> > Example 3: A CAS file downloaded by client only once, could increase casHit counter twice, since also increased when GetValidatedActionResult calls c.Get().

As an example of too much complexity in the code: I had to read the code carefully until I realize that it is still not addressing Example 3. Right?

> I'm not sure if using a decorator instead would be an improvement

As I see it, all of the above complexity would be avoided by a decorator.

What issues do you see with a decorator?

mostynb:
> Thanks @mostynb for answering questions!
> 
> > Doing this inside the disk package doesn't seem to add too much complexity
> 
> From my point of view, it adds the following complexity:
> 
> * Not intuitive how the ac*Validated counters relates to the normal counters. I had to read the code in detail to understand the relation.
> * Exposing the validateAC boolean into disk.go and making assumptions about how GetValidatedActionResult will be invoked from rest of the code base.

These could stand to be clarified in comments.

> * Risk incrementing wrong counter. E.g. forgetting "inc = counters.hit" in some branch when making changhes to the code base. Especially if separating misses vs errors in the future, and even more branches needs consideration.
> * Defer functions that becomes "reconfigured".

That's an interesting point re how to (or if to) count errors. I have been putting off thinking about this, but maybe it's better to make a decision up front. From the client's point of view an error is the same as a cache miss, but from the server's point of view it might be worth counting these separately, or logging errors and not counting. What's your opinion?

> > > Example 3: A CAS file downloaded by client only once, could increase casHit counter twice, since also increased when GetValidatedActionResult calls c.Get().
> 
> As an example of too much complexity in the code: I had to read the code carefully until I realize that it is still not addressing Example 3. Right?

That's right- the current state of this PR assumes that CAS metrics are kind of useless unless there are clients who only use bazel-remote as a CAS (no AC). Is there an clear best thing to do here? Should we count external CAS lookups separately from those triggered by validated AC lookups?

> > I'm not sure if using a decorator instead would be an improvement
> 
> As I see it, all of the above complexity would be avoided by a decorator.
> 
> What issues do you see with a decorator?

Switching to an interface would add a small amount of complexity (need to define an interface, and a small wrapper without a lot of code), and adds the cost of an interface call for every request. I don't think that's a lot of complexity or runtime cost, but I'm also unsure if it adds much benefit. If these are cheap, should we just always keep metrics and remove the configuration option?

ulrfa:
> These could stand to be clarified in comments.

Yes, but I still think it is important that metrics are **intuitive** to understand and reason about. I show metrics also to people that knows nothing about the internals of bazel-remote.
 
> That's an interesting point re how to (or if to) count errors. I have been putting off thinking about this, but maybe it's better to make a decision up front. From the client's point of view an error is the same as a cache miss, but from the server's point of view it might be worth counting these separately, or logging errors and not counting. What's your opinion?

In my internal patches, I categorize each metric increment call as one of the following:
- OK (200 if HTTP)
- Cache Miss (404 if HTTP)
- Client Error (4xx if HTTP)
- Server Error (5xx if HTTP)

And I calculate cache hit ratio based only on cache hits and cache misses. The cache hit ratio is interesting for the people writing bazel BUILD files in particular. But those people can’t do anything about HTTP 5xx errors, and therefore I prefer excluding errors from cache hit ratio calculations. 

The **Server Errors** are interesting for people operating the infrastructure. Those metrics can trigger alarms notifying them to check the log files for details.

I have not seen much use of the **Client Error** category, but I guess those metrics could detect incompatibilities between client and server.

> That's right- the current state of this PR assumes that CAS metrics are kind of useless unless there are clients who only use bazel-remote as a CAS (no AC). Is there an clear best thing to do here? Should we count external CAS lookups separately from those triggered by validated AC lookups?

I exclude CAS accesses for cache hit ratio calculations. But I typically include both AC and CAS accesses when plotting number of requests per second to see load on the server (ignoring cache hit ratio). E.g. to see how the load changes over time, or after adding an additional cache instance. For such load use cases it is not desired to count CAS access twice (both when validating AC and for GET request) when client is only downloading once.

For such load use cases, I find it relevant to also distinguish between:
- Put
- Get 
- Contains

since they have different characteristics. E.g. Put requests are most challenging for our SSDs and Contains requests the least. Therefore I distinguish between GET/PUT/CONTAINS with a label.

I combine that with additional custom labels, such as  CI vs interactive user  and  main branch vs maintenance branches. That allows use cases such as:

- What is the cache hit ratio on main branch when building from CI?
- Are all these PUT requests coming from CI or interactive users?
- ...

Different people have different use cases. And I think we can achieve flexibility by having a **single** metric for accesses to disk.go and instead categorize with labels. Then users get freedom to configure dashboards and write prometheus queries that make sense for their use case. Example of prometheus queries:

Number of AC cache misses, excluding errors:
`bazel_remote_incoming_requests_total{kind="AC",method=~"GET|CONTAINS",status=~"MISS"}`

Number of AC cache misses, including errors:
`bazel_remote_incoming_requests_total{kind="AC",method=~"GET|CONTAINS",status=~"MISS|ERROR"}`

Number of errors, regardless of type of request:
`bazel_remote_incoming_requests_total{status="ERROR"}`

Number of incomming GET requests per second, regardless of AC/CAS or status:
`rate(bazel_remote_incoming_requests_total{method="GET"}[60s])`

Using labels this way is recommended in [prometheus documentation](https://prometheus.io/docs/practices/instrumentation/#use-labels): _“When you have multiple metrics that you want to add/average/sum, they should usually be one metric with labels rather than multiple metrics. For example, rather than http_responses_500_total and http_responses_403_total, create a single metric called http_responses_total with a code label for the HTTP response code. You can then process the entire metric as one in rules and graphs.”_

>  need to define an interface

A well defined interface could also be seen as a way to reduce complexity, in my opinion.

>and adds the cost of an interface call for every request. I don't think that's a lot of complexity or runtime cost, but I'm also unsure if it adds much benefit. If these are cheap, should we just always keep metrics and remove the configuration option?

I believe the runtime cost of a decorator interface call would not be noticeable. I believe the alternative, using defer to increment, would cost more (bust still negligible).

However, the prometheus go library might add some overhead each time increasing a counter, especially when using labels. Personally I would always enable them anyway, but I would not be surprised if some people would prefer them disabled.

mostynb:
Pushed an update to switch to the decorator style, and to not double count cache hits for CAS lookups as part of an AC request. I assume that context-based labels would be easy enough to add to this(?), so I'll defer that and look into classifying/counting client & server side errors next.

Any comments/complaints/feedback on the current design?

ulrfa:
Thanks @mostynb! I think this Cache interface will work well!

Later it could be interesting to consider a common interface for both disk storage and the proxies. E.g. to allow using same metrics decorator implementation also for the proxies. And perhaps even supporting bazel-remote configurations with HTTP/gRPC servers using a proxy directly, with no local disk storage in between. However, personally I have no use cases for that at the moment, and **I’m happy with the Cache interface you created now!**

The new metricsDecorator.GetValidatedActionResult is now easier, but what is the reason for doing `m.acHitValidated.Inc()` instead of `m.counters[cache.AC].hit.Inc()`? I think the later would allow removing all knowledge from disk.go about validateAc boolean and also about the metrics decorator.

I think it is good that Contains and FindMissingBlobs increments same metric, but I see use cases where desire to differentiate those from the more expensive Get. Using the labels “Contains”, “Get” and “Put” would allow that flexibility. (I would use "Contains" label also for "FindMissingBlobs" in that case.)

I would prefer using Prometheus directly from metrics.go, without the Incrementor interface. And a single counter with labels instead of the 3 x incPairs. But maybe the pros/cons of that becomes more clear when trying to add labels.

mostynb:
Thanks for the review.

> Thanks @mostynb! I think this Cache interface will work well!
> 
> Later it could be interesting to consider a common interface for both disk storage and the proxies. E.g. to allow using same metrics decorator implementation also for the proxies. And perhaps even supporting bazel-remote configurations with HTTP/gRPC servers using a proxy directly, with no local disk storage in between. However, personally I have no use cases for that at the moment, and **I’m happy with the Cache interface you created now!**

Let's skip that for now, the proxy interface is smaller than the disk.Cache interface. (And I don't understand S3 or GCS well enough to implement a good LRU-like solution.)

> The new metricsDecorator.GetValidatedActionResult is now easier, but what is the reason for doing `m.acHitValidated.Inc()` instead of `m.counters[cache.AC].hit.Inc()`? I think the later would allow removing all knowledge from disk.go about validateAc boolean and also about the metrics decorator.

You're right- this was unneeded after the decorator refactoring, now removed.

> I think it is good that Contains and FindMissingBlobs increments same metric, but I see use cases where desire to differentiate those from the more expensive Get. Using the labels “Contains”, “Get” and “Put” would allow that flexibility. (I would use "Contains" label also for "FindMissingBlobs" in that case.)
> 
> I would prefer using Prometheus directly from metrics.go, without the Incrementor interface. And a single counter with labels instead of the 3 x incPairs. But maybe the pros/cons of that becomes more clear when trying to add labels.

Time for me to do some background reading on prometheus while I consider these points...

ulrfa:
It is problematic to rename counters in the future, since that makes it harder for people that plot trends over long time. Therefore I'm using `bazel_remote_incoming_requests_total` since I anticipate other metrics for **outgoing** requests to proxies in the future.

But I'm OK also with `bazel_remote_requests_total` since it is shorter, and it does not prevent us from adding a new `bazel_remote_proxy_requests_total` or similar in the future.

mostynb:
@ulrfa: this seems reasonable to me, though I don't like having to add `RegisterMetrics`. Do you see any problems with this in its current form? We can try to remove the need for `RegisterMetrics` and add add custom labels in followup PRs.

Issue #482, opened 2021-10-08
Consider adding option to use cgo zstd implementation


glukasiknuro:
When checking performance of bazel-remote noticed that most of the time is spent in zstd compression.

Created a benchmark that shows that cgo implementation of zstd is more performant (2x-3x compression, 4x-6x decompression) : https://github.com/glukasiknuro/go-zstd-benchmarks

I have created a proof of concept implementation using go zstd implementation of bazel-remote in https://github.com/buchgr/bazel-remote/commit/591894a55266d3552cbd983c65115e24e252df8e, and created a load test that confirms the benefits of using cgo implementation are significant:
https://github.com/glukasiknuro/go-zstd-benchmarks/tree/master/bazel-remote-load-test

I will try to suggest a patch that will allow to select cgo implementation of zstd to bazel-remote if compiled in CGO mode.

Also, some more details about performance of cgo implementation in https://github.com/klauspost/compress/issues/444

mostynb:
Thanks for the excellent investigation work- I will try to take a look early next week.

glukasiknuro:
Marked #485 as ready for review, the overhead of compression is actually expected.

PR #485, opened 2021-10-15
Added -zstd_implementation=cgo option.

glukasiknuro:
Allows to select cgo implementation of zstd if compiled in.

The cgo implementation is more performant than pure go implementation,
see https://github.com/buchgr/bazel-remote/issues/482 for actual benchmarks and dicussion.

Stats show that with cgo implementation the overhead of zstd compression itself is much lower -  40% vs 160% for downloading. It increases possible bandwidth for downloading from bazel-remote from 100MB/s to 190MB/s per single cpu the bazel-remote runs on, so roughly 2x.

For uploading the difference is smaller - 310% vs 490%, the overhead is big mainly because a) compression is slower than decompression in zstd case, and b) bazel-cache needs to perform also sha256 computation, which from basic observations is similarly expensive as zstd compression at faster levels. It increases upload of whole bazel-remote from 45MB/s to 64MB/s.

 See https://github.com/glukasiknuro/go-zstd-benchmarks/tree/master/bazel-remote-load-test#results---download


mostynb:
This landed in #559.

glukasiknuro:
Uh, thanks Mostyn for taking care of it!

PR #488, opened 2021-10-16
Rerun go mod tidy on master

glukasiknuro:
Looks like upgrade to go 1.17 failed some post-merge checks, see https://buildkite.com/bazel/bazel-remote-cache/builds/2060#cfb138b2-9a8d-4791-875e-822cca762457 - probably
because my PR https://github.com/buchgr/bazel-remote/pull/486 was not
on the most recent commit but one before it.

Those command now succeed:
```
go vet ./...
go test -race ./...
```


mostynb:
No problem- thanks for the followup.

PR #493, opened 2021-11-02
[doc] Add example for docker + aws_credentials_file

justfalter:


mostynb (approved):
Thanks!

mambon2:
@mostynb I was reading this quote from aws https://github.com/aws/aws-cli/issues/4982#issuecomment-894491454
In it they state the following:

> As of now, all AWS SDKs except C++ support the credentials from SSO login.

Is this true for this goapp?  If so, which authentication method should we use to leverage sso login (instead of credentials file)

justfalter:
> @justfalter @mostynb I was reading this quote from aws [aws/aws-cli#4982 (comment)](https://github.com/aws/aws-cli/issues/4982#issuecomment-894491454) In it they state the following:
> 
> > As of now, all AWS SDKs except C++ support the credentials from SSO login.
> 
> Is this true for this goapp? If so, which authentication method should we use to leverage sso login (instead of credentials file)

bazel-remote uses the minio go client library, and not the go aws sdk.

FWIW, I was able to use the aws cli + [gimme-aws-creds](https://github.com/Nike-Inc/gimme-aws-creds) to authenticate authenticate using SSO (okta) for AWS. I then tell bazel-remote to use my AWS credentials file, and everything works.

I have a script that I use to wrap up the call to `gimme-aws-creds` so that the AWS cli is all hooked up:
```
#!/bin/bash -e
export AWS_PROFILE=name-of-profile-in-credentials-file
export AWS_REGION=us-east-1
export PS1_PREFIX="(dev-myrole) > "

gimme-aws-creds --roles "arn:aws:iam::1234567:role/MyRole"

# Spawn new bash shell
bash
```

Note: `gimme-aws-creds` will throw an error if you don't have section in your credentials file for your AWS_PROFILE (ex: `[name-of-profile-in-credentials-file]`).

After running this, the section (ex: `[name-of-profile-in-credentials-file]`) in the `~/.aws/credentials` file is populated:
```
[name-of-profile-in-credentials-file]
aws_access_key_id = ...
aws_secret_access_key = ...
aws_session_token = ...
aws_security_token = ...
```

You should be able to do normal aws cli things (ex: `aws s3api list-buckets`).

Once confirmed, you *should* be able to configure bazel-remote to use your credentials file.

PR #496, opened 2021-11-11
ci: update to use latest ubuntu runner and builtin bazelisk

chenrui333:
ref, https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#tools


mostynb:
Thanks- this looks good.

PR #526, opened 2022-02-15
Add a max proxy blob size option

flode:
When the proxy server is only connected with a slow network link, it's sometimes beneficial not to download large blobs and recompile them locally instead.
With the new maxProxyBlobSize setting, the remote blob is not read when it is larger.

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -124,8 +124,9 @@ func New(dir string, maxSizeBytes int64, opts ...Option) (Cache, error) {
 		dir: dir,
 
 		// Not using config here, to avoid test import cycles.
-		storageMode: casblob.Zstandard,
-		maxBlobSize: math.MaxInt64,
+		storageMode:      casblob.Zstandard,
+		maxBlobSize:      math.MaxInt64,
+		maxProxyBlobSize: math.MaxInt64,
 
 		// Go defaults to a limit of 10,000 operating system threads.
 		// We probably don't need half of those for file removals at
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -1044,7 +1044,7 @@ func (c *diskCache) Contains(ctx context.Context, kind cache.EntryKind, hash str
 
 	if c.proxy != nil {
 		exists, foundSize = c.proxy.Contains(ctx, kind, hash)
-		if exists && !isSizeMismatch(size, foundSize) {
+		if exists && size <= c.maxProxyBlobSize && !isSizeMismatch(size, foundSize) {
 			return true, foundSize
 		}
 	}
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -69,12 +69,13 @@ type lruItem struct {
 // diskCache is a filesystem-based LRU cache, with an optional backend proxy.
 // It is safe for concurrent use.
 type diskCache struct {
-	dir           string
-	proxy         cache.Proxy
-	storageMode   casblob.CompressionType
-	maxBlobSize   int64
-	accessLogger  *log.Logger
-	containsQueue chan proxyCheck
+	dir              string
+	proxy            cache.Proxy
+	storageMode      casblob.CompressionType
+	maxBlobSize      int64
+	maxProxyBlobSize int64
+	accessLogger     *log.Logger
+	containsQueue    chan proxyCheck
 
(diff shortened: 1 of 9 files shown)


cache/disk/options.go
```diff
@@ -42,14 +42,15 @@ func WithMaxBlobSize(size int64) Option {
 	}
 }
 
-func WithProxyBackend(proxy cache.Proxy) Option {
+func WithProxyBackend(proxy cache.Proxy, maxProxyBlobSize int64) Option {
```
mostynb, line 45: I think it would probably be nicer to add a `WithMaxProxyBlobSize` function instead of modifying `WithProxyBackend`.

cache/disk/disk.go
```diff
@@ -944,6 +945,9 @@ func (c *diskCache) get(ctx context.Context, kind cache.EntryKind, hash string,
 	if r == nil {
 		return nil, -1, nil
 	}
+	if foundSize > c.maxProxyBlobSize {
```
mostynb, line 948: If we know the size ahead of time, we might be able to avoid the `c.Proxy.Get` call above. Maybe add a check in `availableOrTryProxy` ?

cache/disk/disk.go
```diff
@@ -944,6 +945,9 @@ func (c *diskCache) get(ctx context.Context, kind cache.EntryKind, hash string,
...
+		return nil, -1, nil
```
mostynb, line 949: I think we should call `r.Close()` if it's non-nil and we don't return `r`.

cache/disk/disk.go
```diff
@@ -1040,7 +1044,7 @@ func (c *diskCache) Contains(ctx context.Context, kind cache.EntryKind, hash str
 
 	if c.proxy != nil {
 		exists, foundSize = c.proxy.Contains(ctx, kind, hash)
```
mostynb, line 1046: If we know the size ahead of time, and it's larger than maxProxyBlobSize we could skip this call.

cache/disk/disk_test.go
```diff
@@ -913,16 +914,32 @@ func TestHttpProxyBackend(t *testing.T) {
 		t.Fatal("Expected testCache to be empty")
 	}
 
-	// Add the proxy backend and check that we can Get the item.
+	// Add the proxy backend
 	testCache.proxy = proxy
+	testCache.maxProxyBlobSize = 0
```
mostynb, line 919: maxProxyBlobSize 0 is invalid- could you make this blobSize-1 instead?

cache/disk/disk_test.go
```diff
@@ -913,16 +914,32 @@ func TestHttpProxyBackend(t *testing.T) {
...
+
+	r, fetchedSize, err := testCache.Get(ctx, cache.CAS, casHash, blobSize, 0)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if r != nil {
+		t.Fatal("Expected the Get to fail")
+	}
+
+	// Wet a larger max proxy blob size and check that we can Get the item.
```
mostynb, line 933: s/Wet/Set/

mostynb (commented):
Hi, thanks for the pull request, this sounds like a useful feature.

flode:
Thanks for the review @mostynb ! I've updated the pull request with your suggestions.

cache/httpproxy/httpproxy_test.go
```diff
@@ -122,7 +123,7 @@ func TestEverything(t *testing.T) {
 	}
 
 	diskCacheSize := int64(len(casData) + disk.BlockSize)
-	diskCache, err := disk.New(cacheDir, diskCacheSize, disk.WithProxyBackend(proxyCache), disk.WithAccessLogger(testutils.NewSilentLogger()))
+	diskCache, err := disk.New(cacheDir, diskCacheSize, disk.WithProxyBackend(proxyCache), disk.WithProxyMaxBlobSize(math.MaxInt64), disk.WithAccessLogger(testutils.NewSilentLogger()))
```
mostynb, line 126: I think you can undo this too (and another one below).

mostynb (commented):
Looks pretty good, just a couple of minor things left.

flode:
Good catch, I had to add max proxy blob size to the workaround in disk.go

mostynb:
Thanks for finishing this off- merged.

PR #527, opened 2022-02-19
Implement unambiguous compressed blobs bytestream upload early return

mostynb:
This is an implementation of this REAPI spec update:
https://github.com/bazelbuild/remote-apis/pull/213

Which is part of the solution to this issue:
https://github.com/bazelbuild/bazel/issues/14654


mostynb:
The corresponding spec change was merged, as well as a bazel fix.

PR #530, opened 2022-03-09
Improve GetValidatedActionResult() performance by validating action result blobs in parallel

justfalter:
Speed up `GetValidatedActionResult()` for proxy backends by validating
result blobs in parallel.

Previous behavior would iterate through result blobs sequentially. An
action with hundreds (or thousands) of result blobs to check at 10-20ms
each could easily run into timeouts for the caller.

Really, this is just reusing some of the functionality that already exists in `FindMissingCasBlobs()`.  As a result, this should also improve performance for the local cache lookups since it is using `findMissingLocalCAS()` to check the local cache in batches.

### DIFF
--- a/cache/disk/findmissing.go
+++ b/cache/disk/findmissing.go
@@ -127,13 +127,15 @@ func (c *diskCache) findMissingLocalCAS(blobs []*pb.Digest) int {
 func (c *diskCache) containsWorker() {
 	var ok bool
 	for req := range c.containsQueue {
-		select {
-		case <-req.ctx.Done():
-			// Fast-fail if the context has already been cancelled.
-			c.accessLogger.Printf("GRPC CAS HEAD %s CANCELLED", (*req.digest).Hash)
-			req.wg.Done()
-			continue
-		default:
+		if req.ctx != nil {
+			select {
+			case <-req.ctx.Done():
+				// Fast-fail if the context has already been cancelled.
+				c.accessLogger.Printf("GRPC CAS HEAD %s CANCELLED", (*req.digest).Hash)
+				req.wg.Done()
+				continue
+			default:
+			}
 		}
 
 		ok, _ = c.proxy.Contains(req.ctx, cache.CAS, (*req.digest).Hash)
--- a/cache/disk/findmissing.go
+++ b/cache/disk/findmissing.go
@@ -16,7 +16,7 @@ type proxyCheck struct {
 	wg          *sync.WaitGroup
 	digest      **pb.Digest
 	ctx         context.Context
-	onProxyMiss *func()
+	onProxyMiss func()
 }
 
 // Optimised implementation of FindMissingBlobs, which batches local index
@@ -25,16 +25,21 @@ type proxyCheck struct {
 //
 // Note that this modifies the input slice and returns a subset of it.
 func (c *diskCache) FindMissingCasBlobs(ctx context.Context, blobs []*pb.Digest) ([]*pb.Digest, error) {
-	err := c.findMissingCasBlobsInternal(ctx, blobs, nil)
-	if err != nil {
-		return nil, err
-	}
-
-	return filterNonNil(blobs), nil
+	return c.findMissingCasBlobsInternal(ctx, blobs, false)
 }
 
-func (c *diskCache) findMissingCasBlobsInternal(ctx context.Context, blobs []*pb.Digest, onProxyMiss *func()) error {
+func (c *diskCache) findMissingCasBlobsInternal(ctx context.Context, blobs []*pb.Digest, failFast bool) ([]*pb.Digest, error) {
+	// batchSize moderates how long the cache lock is held by findMissingLocalCAS.
 	const batchSize = 20
+
+	var contextCancel context.CancelFunc = nil
+
+	if failFast && c.proxy != nil {
+		// We want the contains worker to cancel the context on a cache miss, allowing us to fail fast.
+		ctx, contextCancel = context.WithCancel(ctx)
+
+		defer contextCancel()
+	}
 
 	var wg sync.WaitGroup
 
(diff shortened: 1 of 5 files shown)


justfalter:
I've got some tests to fix up in this first

justfalter:
Tests added, should be good to go.

justfalter:
Any thoughts @mostynb? This has made a pretty significant difference in my build environment (bazel-remote + aws s3).

cache/disk/findmissing.go
```diff
@@ -88,21 +119,27 @@ func filterNonNil(blobs []*pb.Digest) []*pb.Digest {
...
 	var exists bool
+	var item lruItem
 	var key string
 	missing := 0
 
 	c.mu.Lock()
 
 	for i := range blobs {
-		if blobs[i].SizeBytes == 0 {
+		if blobs[i].SizeBytes == 0 && blobs[i].Hash == emptySha256 {
```
mostynb, line 129: Just curious, did you hit a client bug here?

cache/disk/findmissing_test.go
```diff
@@ -148,3 +153,208 @@ func TestContainsWorker(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(missing) != 2 {
+		t.Fatal("Expected missing array to have exactly two entries")
+	}
+
+	if !proto.Equal(missing[0], &digest2) {
+		t.Fatalf("Expected missing[0] == digest2")
```
mostynb, line 234: Should we log the actual value of `digest2` here?

cache/disk/findmissing_test.go
```diff
@@ -148,3 +153,208 @@ func TestContainsWorker(t *testing.T) {
...
+	}
+
+	if !proto.Equal(missing[1], &digest4) {
+		t.Fatalf("Expected missing[1] == digest4")
```
mostynb, line 238: And the value of `digest4` here?

cache/disk/findmissing_test.go
```diff
@@ -148,3 +153,208 @@ func TestContainsWorker(t *testing.T) {
...
+	if !proto.Equal(missing[0], &digest2) {
+		t.Fatalf("Expected missing[0] == digest2")
+	}
+
+	if !proto.Equal(missing[1], &digest3) {
+		t.Fatalf("Expected missing[1] == digest3")
+	}
+
+	if !proto.Equal(missing[2], &digest4) {
+		t.Fatalf("Expected missing[2] == digest4")
```
mostynb, line 309: Log digest{2,3,4} here?

cache/disk/findmissing_test.go
```diff
@@ -148,3 +153,208 @@ func TestContainsWorker(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(missing) != 1 {
+		t.Fatalf("Expected missing array to have exactly one entry, got %d", len(missing))
+	}
+
+	if !proto.Equal(missing[0], &digest2) {
+		t.Fatalf("Expected missing[0] == digest2")
```
mostynb, line 358: Here too?

mostynb (commented):
Thanks for this PR- the improvement is a good one. I mostly only have some pedantic comments.

cache/disk/findmissing.go
```diff
@@ -47,27 +62,43 @@ func (c *diskCache) FindMissingCasBlobs(ctx context.Context, blobs []*pb.Digest)
...
 				}
 			}
+		} else if failFast {
+			break
 		}
 	}
 
 	if c.proxy != nil {
 		wg.Wait()
 	}
```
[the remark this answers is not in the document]
justfalter: hmmm.. If we are failing fast with a proxy, then there really is no reason to wait around. The context will be cancelled, so any pending `proxyCheck`'s will do no real work.

I've got a few thoughts while re-reading this code.
1. We should be checking `ctx.Done()` right before attempting to add to `conteainsQueue` since additions could block when the channel is full.
2. We should differentiate between the requestor cancelling `ctx` (returns a grpc Cancelled error) and our canceling in an effort to fast fail (not an error).
3. We should be watching the context(s) while we are waiting for the waitgroup to clear out. This will allow us to respect cancellation via the upstream context, fail fast, *and* let things spin down normally.

cache/disk/findmissing.go
justfalter: No, simply trying to make it consistent with https://github.com/buchgr/bazel-remote/blob/e0f5a5c447e591949d05da040b653e89f9ce8d1d/cache/disk/disk.go#L1029

cache/disk/findmissing.go
```diff
@@ -24,8 +25,22 @@ type proxyCheck struct {
...
 // Note that this modifies the input slice and returns a subset of it.
 func (c *diskCache) FindMissingCasBlobs(ctx context.Context, blobs []*pb.Digest) ([]*pb.Digest, error) {
+	return c.findMissingCasBlobsInternal(ctx, blobs, false)
+}
+
+func (c *diskCache) findMissingCasBlobsInternal(ctx context.Context, blobs []*pb.Digest, failFast bool) ([]*pb.Digest, error) {
+	// batchSize moderates how long the cache lock is held by findMissingLocalCAS.
 	const batchSize = 20
 
+	var contextCancel context.CancelFunc = nil
```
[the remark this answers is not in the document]
justfalter: I've removed the bit that was calling `contextCancel` within this function, so that check is gone. Hopefully the code is a bit more clear, now.

mostynb (commented):
This is looking better and better, but I want to think it through for another day before landing.

mostynb:
Thanks- merged!

mostynb:
I pushed a couple of minor formatting changes that I didn't want to hold up the this review for.

I don't have a test setup for AWS S3, but if you have tried this out and it looks good I can make a new release.

justfalter:
The s3 stuff has been working fine, for me. I'm going to run one more build through a build off master just to make sure everything is jiving.. will report back in a couple hours.

justfalter:
@mostynb All my testing shows that it's working great! s3 no longer hangs for 2 minutes for the timeout (#531), and the validation performance is great! :shipit:

PR #531, opened 2022-03-11
Update minio-go - fix for s3 IAM auth delays

justfalter:
- Update minio-go so that we can get a fix for timeouts encountered when
  trying to retrieve IMDSv2 tokens during IAM s3 authentication. When
  bazel-remote is run within a container on an aws ec2 instance, initial
  attempts to authenticate using IAM will encounter a 120s delay. The
  issue is the result of changes in the ec2 metadata services with
  IMDSv2, which restricts how many network hops may be incurred before
  for PUT statements. When executing directly on the ec2 instance, it
  is 1 hop away, and is allowed. When executing within a docker
  container on an ec2 instance, it is 2 hops away, and is forbidden. The
  change to minio go reduces the timeout from 120s to 1s. Apparently
  this token request is optional, so things continue to work afterward,
  regardless.
  https://github.com/minio/minio-go/pull/1626


mostynb:
Thanks for this fix, merged.

PR #538, opened 2022-04-01
Add missing access log entry

ulrfa:
Add missing access log entry for successfull HTTP Get
of AC with enabled validation.


mostynb (approved):

ulrfa:
Thanks @mostynb

PR #539, opened 2022-04-03
Add support for Prometheus metric categories

ulrfa:
This pull request replaces my previous #350 and #358 PRs (@mostynb: feel free to close them if you want).

What do you think @mostynb? If you would like a subset of this PR, then I can restructure it. If you decide to not merge this PR that is also fine by me, since the context parameters and metrics decorator interface from the previous commits, allows me to maintain this as separate patch without too much effort.


cache/disk/BUILD.bazel
```diff
@@ -19,6 +19,7 @@ go_library(
         "//utils/validate:go_default_library",
         "@com_github_djherbis_atime//:go_default_library",
         "@com_github_prometheus_client_golang//prometheus:go_default_library",
+        "@org_golang_google_grpc//metadata:go_default_library",
```
mostynb, line 22: Moving this down one line might fix the buildifier check?

mostynb:
I'll try to take a look at this in the next day or two...

ulrfa:
> I'll try to take a look at this in the next day or two...

Thanks @mostynb! I'll let you have look before I address the current comments, since I assume reviewing is more convenient with fewer commits and without force-pushed commits.

mostynb:
I have cherry-picked the FindMissingCasBlobs metrics fix to the master branch- good catch.

Re disabling Contains metrics, let's think this through a bit first, if it's not a big change I think it would be better to fix them than to disable them and then fix them. I have an idea for how to do this, that I need to run through in my head a bit. I think this should be dropped from this PR.

Re the metrics categories change, I'll try to take a look tomorrow.

ulrfa:
> I have cherry-picked the FindMissingCasBlobs metrics fix to the master branch- good catch.

Thanks @mostynb! 

> Re disabling Contains metrics, let's think this through a bit first, if it's not a big change I think it would be better to fix them than to disable them and then fix them. I have an idea for how to do this, that I need to run through in my head a bit. I think this should be dropped from this PR.

OK, I removed that commit from this PR.

> Re the metrics categories change, I'll try to take a look tomorrow.

Thanks @mostynb! I now force-pushed the branch to address the comments so far.

README.md
```diff
@@ -610,3 +630,86 @@ To avoid per-prefix rate limiting with Amazon S3, you may want to try using
...
+> increase the amount of data stored. Do not use labels to store
+> dimensions with high cardinality (many different label values),
+> such as user IDs, email addresses, or other unbounded sets of
+> values."
+
+Received headers that match a declared category name, but with a value outside
+the declared allowed value set, is reported with the value "other" to
+Prometheus. This is convenient for categories such as "branch", where it from a
+cache hit ratio perspective often make sense to distinguish between "main"
+branch and "other" branches.
```
mostynb, line 659: I think we should mention here that only lowercase category names are allowed, otherwise it will likely be a common error that people will hit when first trying this feature. Also: are non lowercase values allowed?

cache/disk/metrics.go
```diff
@@ -118,17 +126,97 @@ func (m *metricsDecorator) FindMissingCasBlobs(ctx context.Context, blobs []*pb.
...
+		grpcHeaders = getGrpcHeaders(ctx)
+	}
+
+	for categoryNameLowerCase, allowedValues := range m.categories {
+		// Lower case is canonical for gRPC headers and convention for prometheus.
+		var headerValue string = ""
+		if grpcHeaders != nil {
+			grpcHeaderValues := grpcHeaders[categoryNameLowerCase]
+			if len(grpcHeaderValues) > 0 {
+				// Pick the first header with matching name if multiple headers with same name
```
mostynb, line 178: I wonder if it would be less surprising to pick the last matching header instead of the first?

cache/disk/metrics.go
```diff
@@ -118,17 +126,97 @@ func (m *metricsDecorator) FindMissingCasBlobs(ctx context.Context, blobs []*pb.
...
+				headerValue = grpcHeaderValues[0]
+			}
+		} else if httpHeaders != nil {
```
mostynb, line 181: Just checking, when using grpc is it assumed that we should ignore http headers too?

cache/disk/metrics.go
```diff
@@ -118,17 +126,97 @@ func (m *metricsDecorator) FindMissingCasBlobs(ctx context.Context, blobs []*pb.
...
+			headerValue = httpHeaders.Get(categoryNameLowerCase)
+		}
```
mostynb, line 183: nitpick: add an empty line after this

config/config.go
```diff
@@ -351,6 +352,12 @@ func validateConfig(c *Config) error {
 		return errors.New("'access_log_level' must be set to either \"none\" or \"all\"")
 	}
 
+	for categoryName := range c.MetricCategories {
+		if categoryName != strings.ToLower(categoryName) {
+			return fmt.Errorf("Names in 'metric_categories' must be in lower case: %s", categoryName)
```
mostynb, line 357: nitpick: reword to `"metric_categories names must be lowercase: %q", categoryName`

cache/disk/metrics.go
```diff
@@ -118,17 +126,97 @@ func (m *metricsDecorator) FindMissingCasBlobs(ctx context.Context, blobs []*pb.
...
+	if !ok {
+		return nil
+	}
+	return headers
+}
+
+func getGrpcHeaders(ctx context.Context) metadata.MD {
+	grpcHeaders, _ := metadata.FromIncomingContext(ctx)
+	return grpcHeaders
+}
```
mostynb, line 222: I wonder if it would be a little cleaner to find http/grpc headers in the server package (separately in http and grpc code), and then add them as a single custom value in the context? Then we could simplify addCategoryLabels a little.

mostynb (commented):
Sorry this took so long for me to get to. This change looks good, I only have one substantial suggestion (move http/grpc header processing into the http/grpc code in the server package).

ulrfa:
Sorry, I have not had time to address these comments. I’m going on vacation now and I’m back again in middle of August. Wish you a good summer!

derigel23:
Is this PR going forward?

PR #540, opened 2022-04-12
Remove S3 performance note from README

jcourteau:
Now that `s3.key_version` is deprecated, and defaults to `2`, the performance note at the bottom of the README isn't relevant anymore. Furthermore, it had a typo anyways - it mentioned `s3.key_format` which never existed.


mostynb (approved):
Thanks for the cleanup.

PR #542, opened 2022-04-15
Update objects timestamps on cache hit

PycKamil:
In this PR I would like to introduce option to leverage S3 storage lifecycle to automatically remove not used objects after specified amount of time. 
We [can configure S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-set-lifecycle-configuration-intro.html) to remove objects after certain amount of time from its modification date. To update modification date, on each cache hit we need to copy object in place - that's limitation of s3. We can now pass `s3.update_timestamps` option in config to use it.

### DIFF
--- a/cache/s3proxy/s3proxy.go
+++ b/cache/s3proxy/s3proxy.go
@@ -24,14 +24,15 @@ type uploadReq struct {
 }
 
 type s3Cache struct {
-	mcore        *minio.Core
-	prefix       string
-	bucket       string
-	uploadQueue  chan<- uploadReq
-	accessLogger cache.Logger
-	errorLogger  cache.Logger
-	v2mode       bool
-	objectKey    func(hash string, kind cache.EntryKind) string
+	mcore            *minio.Core
+	prefix           string
+	bucket           string
+	uploadQueue      chan<- uploadReq
+	accessLogger     cache.Logger
+	errorLogger      cache.Logger
+	v2mode           bool
+	updateTimestamps bool
+	objectKey        func(hash string, kind cache.EntryKind) string
 }
 
 var (
@@ -56,6 +57,7 @@ func New(
 	Prefix string,
 	Credentials *credentials.Credentials,
 	DisableSSL bool,
+	UpdateTimestamps bool,
 	Region string,
 
 	storageMode string, accessLogger cache.Logger,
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -415,6 +415,7 @@ func get(ctx *cli.Context) (*Config, error) {
 			AccessKeyID:              ctx.String("s3.access_key_id"),
 			SecretAccessKey:          ctx.String("s3.secret_access_key"),
 			DisableSSL:               ctx.Bool("s3.disable_ssl"),
+			UpdateTimestamps:         ctx.Bool("s3.update_timestamps"),
```
mostynb, line 418: If we end up keeping this option (instead of making it always-on), we'll need to update the README.md file with the new help text and sample config file.

cache/s3proxy/s3proxy.go
```diff
@@ -196,6 +199,22 @@ func (c *s3Cache) Put(ctx context.Context, kind cache.EntryKind, hash string, si
...
+		Bucket: bucket,
+		Object: object,
+	}
+
+	dst := minio.CopyDestOptions{
+		Bucket: bucket,
+		Object: object,
+	}
+
+	_, err := c.mcore.ComposeObject(context.Background(), dst, src)
```
mostynb, line 213: I see that the copying is done server side- nice. Are there any downsides to having this always on?

mostynb (commented):
Hi, thanks for the PR.

cache/s3proxy/s3proxy.go
PycKamil: Good question - I don't see any downsides in terms of performance or storage usage but I'm running self hosted s3 storage. As I understand AWS s3 [pricing documentation](https://aws.amazon.com/s3/pricing/) there is some cost for coping object - `you pay the S3 charges for storage in the selected destination S3 storage classes, for the primary copy, for replication PUT request`. So making this turn on by default could potentially generate costs for people using AWS so maybe it's better to leave it as opt-in?

config/config.go
PycKamil: Let me know what you think about leaving it opt-in, then I will update documentation if is stays this way 👍

cache/s3proxy/s3proxy.go
mostynb: OK, let's make it opt-in to be on the safe side.

mostynb:
Thanks for the contribution- merged.

PR #543, opened 2022-04-17
Make remoteHTTPProxyCache.Contains do HEAD request

noxiouz:
As only headers and response status are important and the body is ignored
doing HTTP HEAD will save some network.

Signed-off-by: Anton Tiurin <atiurin@proton.me>


mostynb (approved):
Thanks, it looks like we used to do this and I accidentally(?) switched from HEAD to GET in ba2800f57adeaa366fe40df2aec890d6cefd0c43.

PR #480, opened 2021-09-30
[disk] limit the number of concurrent file removals

mostynb:
The disk cache's eviction function removes files in a separate goroutine for each, to avoid holding the lock while waiting for the filesystem.

If the disk/filesystem is too slow to keep up with the rate of file removals, we can hit Go's default 10,000 operating system thread limits, and crash. To avoid this, let's try limiting the number of concurrent file removals to half the default limit (5,000).

This might fix #479.

PR #544, opened 2022-04-25
Make Go MaxThreads configurable

SrodriguezO:
We're running into an issue where threads occassionally back up on IO, leading the service to crash with this error:
```
runtime: program exceeds 10000-thread limit
fatal error: thread exhaustion
```

Our host has sufficient spare memory to handle a much larger number of threads during IO hiccups, so we'd like to configure this value higher.

Is this a change you'd be open to upstreaming?

I've validated the MaxThreads limit gets set locally, but I'm open to also adding an automated test if it seems worthwhile (I'd run the service with `BAZEL_REMOTE_MAX_THREADS=1` and expect a core dump with a `runtime: program exceeds 1-thread limit` error. I'm open to other suggestions).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -40,6 +40,7 @@ type Config struct {
 	ProfileAddress              string                    `yaml:"profile_address"`
 	Dir                         string                    `yaml:"dir"`
 	MaxSize                     int                       `yaml:"max_size"`
+	MaxThreads                  int                       `yaml:"max_threads"`
 	StorageMode                 string                    `yaml:"storage_mode"`
 	HtpasswdFile                string                    `yaml:"htpasswd_file"`
 	TLSCaFile                   string                    `yaml:"tls_ca_file"`
@@ -91,7 +92,11 @@ var defaultDurationBuckets = []float64{.5, 1, 2.5, 5, 10, 20, 40, 80, 160, 320}
 
 // newFromArgs returns a validated Config with the specified values, and
 // an error if there were any problems with the validation.
-func newFromArgs(dir string, maxSize int, storageMode string,
+func newFromArgs(
+	dir string,
+	maxSize int,
+	maxThreads int,
+	storageMode string,
 	httpAddress string, grpcAddress string,
 	profileAddress string,
 	htpasswdFile string,
@@ -122,6 +127,7 @@ func newFromArgs(dir string, maxSize int, storageMode string,
 		ProfileAddress:              profileAddress,
 		Dir:                         dir,
 		MaxSize:                     maxSize,
+		MaxThreads:                  maxThreads,
 		StorageMode:                 storageMode,
 		HtpasswdFile:                htpasswdFile,
 		MaxQueuedUploads:            maxQueuedUploads,
@@ -441,6 +447,7 @@ func get(ctx *cli.Context) (*Config, error) {
 	return newFromArgs(
 		ctx.String("dir"),
 		ctx.Int("max_size"),
+		ctx.Int("max_threads"),
 		ctx.String("storage_mode"),
 		httpAddress,
 		grpcAddress,
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -16,6 +16,7 @@ port: 8080
 grpc_port: 9092
 dir: /opt/cache-dir
 max_size: 100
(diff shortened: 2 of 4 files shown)


mostynb:
Which version of bazel-remote are you running? In particular, I wonder if it has this fix or not: #480

SrodriguezO:
We're running 2.3.4, which does indeed have that fix.

Sidenote: would it be worth setting the [file removal semaphore](https://github.com/buchgr/bazel-remote/blob/v2.3.4/cache/disk/disk.go#L134) based on this config too?

mostynb:
> Sidenote: would it be worth setting the [file removal semaphore](https://github.com/buchgr/bazel-remote/blob/v2.3.4/cache/disk/disk.go#L134) based on this config too?

Perhaps. I wonder if you could share a stack trace when the issue occurs? It would be good to see where all the OS threads are being used.

SrodriguezO:
We increased provisioned IOPS/throughput on the EBS device that backs our cache, and IOWait and used file handles both plummeted. I'm pretty sure this will solve our pain point, so maybe making this max thread count configurable isn't necessary - 10,000 OS threads is quite a lot as it is.. that said, having the option might be nice in a pinch, so open to discuss either way.

>I wonder if you could share a stack trace when the issue occurs? It would be good to see where all the OS threads are being used.

I've attached it on this comment (it's rather large). Here's the summary of the goroutine states:
```
{'chan receive, 1 minutes' -> '1147'}
{'chan receive' -> '3357'}
{'chan receive, 4240 minutes' -> '48'}
{'chan send' -> '1'}
{'IO wait, 16 minutes' -> '1'}
{'IO wait, 1 minutes' -> '22'}
{'IO wait' -> '252'}
{'IO wait, 2 minutes' -> '2'}
{'IO wait, 3 minutes' -> '1'}
{'IO wait, 4221 minutes' -> '1'}
{'runnable' -> '8676'}
{'select, 10 minutes' -> '13'}
{'select, 11 minutes' -> '8'}
{'select, 13 minutes' -> '2'}
{'select, 14 minutes' -> '4'}
{'select, 15 minutes' -> '5'}
{'select, 16 minutes' -> '1'}
{'select, 17 minutes' -> '6'}
{'select, 18 minutes' -> '6'}
{'select, 19 minutes' -> '1'}
{'select, 1 minutes' -> '33'}
{'select, 23 minutes' -> '5'}
{'select, 25 minutes' -> '3'}
{'select, 26 minutes' -> '2'}
{'select, 27 minutes' -> '5'}
{'select, 29 minutes' -> '1'}
{'select, 2 minutes' -> '28'}
{'select, 3 minutes' -> '7'}
{'select, 4 minutes' -> '5'}
{'select, 5 minutes' -> '20'}
{'select, 6 minutes' -> '1'}
{'select, 7 minutes' -> '5'}
{'select, 8 minutes' -> '5'}
{'select' -> '9206'}
{'select, 9 minutes' -> '3'}
{'syscall' -> '1246'}
{'syscall, 1 minutes' -> '76'}
```

[d481-crash-dump.tar.gz](https://github.com/buchgr/bazel-remote/files/8566664/d481-crash-dump.tar.gz) (sha256: `7a8329ce781726de03a3a3355d4cefb7bb92b91f4f3eafbb20902c08ba99bf32`)

mostynb:
Thanks for providing the goroutine stack trace data.

I think a lot of these threads are blocked waiting for the file removal semaphore- `grep -c Acquire bazel-remote-d481-dump.txt` gives 6538 hits, which I think means there must have been 5000 in-flight operations that had already acquired the semaphore and were waiting for files to be removed.

> We increased provisioned IOPS/throughput on the EBS device that backs our cache, and IOWait and used file handles both plummeted. I'm pretty sure this will solve our pain point, so maybe making this max thread count configurable isn't necessary - 10,000 OS threads is quite a lot as it is.. that said, having the option might be nice in a pinch, so open to discuss either way.

I think this is a better solution for your users- IMO the cache is less useful the slower it is.

I would prefer not to add new configuration options that might not be necessary, is it OK with you if we close this PR? We can always reopen it later if needed.

SrodriguezO:
Yeah that sounds reasonable; thanks for the help :)

Issue #545, opened 2022-04-29
Example config.yml from the README.md results in "Connection reset by peer" error


LKreutzer:
### Problem:

The example `config.yml` from the `README.md` results in `Connection reset by peer` error when sending requests due to `http_address: localhost:8080`. This seems to be due to `localhost` not being known/resolved properly in the container (?).

### Steps to reproduce:
Do a docker run:
```
docker run -v /some/folder:/data \
                -v /some/other/folder/config:/config \
                -u 1000:1000 -p 9090:8080 -p 9092:9092 \
                buchgr/bazel-remote-cache:latest \
                --config_file=/some/other/folder/config/config.yml
```
with config file `config.yml`:
```
dir: /data
max_size: 50
http_address: localhost:8080
```
or
```
dir: /data
max_size: 50
http_address: 127.0.0.1:8080
```
Sending a request `curl -v  'http://localhost:9090/status'` will result in:
```
*   Trying 127.0.0.1:9090...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9090 (#0)
> GET /status HTTP/1.1
> Host: localhost:9090
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Recv failure: Connection reset by peer
* Closing connection 0
curl: (56) Recv failure: Connection reset by peer
```
Doing the same with `http_address` set to either `0.0.0.0:8080` or just `:8080` in the `config.yml` succeeds and outputs the status.

### Solution:
Either
- adapt the `README.md` (feel free to comment on the draft PR #546) to reflect a working configuration 
or 
- modify the internal logic (in this case feel free to close the draft PR #546).

mostynb:
Thanks for the report, I guess this can be closed now that the README.md PR was merged?

PR #546, opened 2022-04-29
Modify README.md to reflect valid config.yml

LKreutzer:
### Description:

- See issue #545.
- Replace "localhost" by "0.0.0.0" in `README.md` which is a setup that avoids the "Connection reset by peer" error when sending requests.
- Unclear whether the `profile_address: 127.0.0.1:7070` should also be modified in the `README.md`?


mostynb:
Hi this change looks good.

I'm not sure if people consider profiling data to be sensitive, so I think it makes sense for it to be disabled by default and suggesting a more restrictive value when enabling it in the readme makes sense.

Issue #554, opened 2022-06-10
mac: runtime: failed to create new OS thread


shubhindia:
We are using latest release available.

```
2022/06/10 10:46:01 bazel-remote built with go1.18 from git commit f774f6dc82c22540511241a3db58462898edc2be.
```

We are using bazel-remote as a cache backend for ios-builds. It was running fine for about 3 weeks now but now we are getting below error after starting the server.

```
runtime: failed to create new OS thread
```

There aren't any logs available apart from the access logs. Only above error.  We are running it on an m1-mac-mini with 512 GB

Host OS: Mac OS
Version: Mac OS 12.2.1
RAM: 16GB

mostynb:
Hi, a quick web search for "runtime: failed to create new OS thread golang m1 mac" gives me several mentions of people having similar trouble with docker- are you using docker here too? Is that the entire log line, or does it have something like `(have 2 already; errno=22)` at the end?

If you are using docker, I suggest trying the bazel-remote-2.3.7-darwin-arm64 binary from https://github.com/buchgr/bazel-remote/releases without docker.

shubhindia:
Hi, @mostynb  no we are not using docker. And yes, that is the entire log line. There is nothing else before or after that. I even tried with latest merges  (built one locally with latest changes) and still the issue persists. I am pretty certain this issue started happening when we reached the storage limit which we have set for cache. I was hoping  https://github.com/buchgr/bazel-remote/pull/480 would fix it but issue still persists.

mostynb:
Is this a HDD or an SSD?

shubhindia:
@mostynb  its an SSD

shubhindia:
Also the storage metrics which it is reporting are incorrect. I manually cleaned up some old files by running `find . -type f -mtime +9 -delete` and it went down to 304GB, but its still reporting 402GB.

![Screenshot 2022-06-13 at 1 16 56 PM](https://user-images.githubusercontent.com/7694806/173306036-e33c3688-23cb-487e-9a31-281abedf3e66.png)

<img width="869" alt="Screenshot 2022-06-13 at 1 22 30 PM" src="https://user-images.githubusercontent.com/7694806/173306091-2fffb05a-21e1-4b8c-972b-2d19ad6e825b.png">

shubhindia:
That's what I figured. We already had a limit of 300GB and it went above that. Hence, I had to increase the limit.

mostynb:
[...]

PR #556, opened 2022-06-16
[disk] limit the number of concurrent file removals even more on mac

mostynb:
This also adds a startup log line mentioning the value, to help debugging.

This might avoid #554.


shubhindia:
Should we move this behind a flag instead of setting this to 3000 or 5000. People with high config systems should not be limited to 3k or 5k.

mostynb:
I think we can try this without a flag initially, and add one later if needed.

PR #561, opened 2022-07-05
feat: Fix get blob when using azure proxy

bakjos:
This PR fixes the blob download  by returning nil when the blob is not found instead of the error

### DIFF
--- a/cache/azblobproxy/azblobproxy.go
+++ b/cache/azblobproxy/azblobproxy.go
@@ -82,6 +82,12 @@ func (c *azBlobCache) Get(ctx context.Context, kind cache.EntryKind, hash string
 
 	resp, err := client.Download(context.Background(), nil)
 	if err != nil {
+		var stgErr *azblob.StorageError
+		if errors.As(err, &stgErr) && stgErr.ErrorCode == azblob.StorageErrorCodeBlobNotFound {
+			cacheMisses.Inc()
+			logResponse(c.accessLogger, "DOWNLOAD", c.storageAccount, c.container, key, errNotFound)
+			return nil, -1, nil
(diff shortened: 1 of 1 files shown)


bakjos:
@mostynb I just realized I was not returning the right result when the blob was not found and only see the error when running with grpc

cache/azblobproxy/azblobproxy.go
```diff
@@ -82,6 +82,12 @@ func (c *azBlobCache) Get(ctx context.Context, kind cache.EntryKind, hash string
 
 	resp, err := client.Download(context.Background(), nil)
 	if err != nil {
+		var stgErr *azblob.StorageError
+		if errors.As(err, &stgErr) && stgErr.ErrorCode == azblob.StorageErrorCodeBlobNotFound {
+			cacheMisses.Inc()
```
mostynb, line 87: Since we do `cacheMisses.Inc()` in both paths, maybe do that as the first line inside the `if err != nil {` block?

cache/azblobproxy/azblobproxy.go
```diff
@@ -120,12 +126,17 @@ func (c *azBlobCache) Contains(ctx context.Context, kind cache.EntryKind, hash s
 
 	client, err := c.containerClient.NewBlobClient(key)
```
mostynb, line 128: nitpick: remove empty line

mostynb (changes requested):
lgtm with two small suggestions

bakjos:
@mostynb Thanks for the feedback, I fixed both

mostynb (approved):
Thanks, squashed and merged.

PR #591, opened 2022-10-08
Shutdown gracefully on sigint (ctrl-c) and sigterm

mostynb:
For now, the grpc service (if running) is gracefully stopped first, then the http service is gracefully stopped.

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,8 +8,10 @@ import (
 	"net/http"
 	_ "net/http/pprof" // Register pprof handlers with DefaultServeMux.
 	"os"
+	"os/signal"
 	"runtime"
 	"strings"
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -85,6 +89,45 @@ func run(ctx *cli.Context) error {
...
+					err := httpServer.Shutdown(context.Background())
+					if err != nil {
+						log.Println("Error occurred while stopping HTTP server:", err)
+					}
+
+					// The "HTTP server stopped" log comes from the main goroutine.
+				}
+			}
+
+			os.Exit(0)
```
mostynb, line 127: I guess this is a race condition with the main goroutine exiting.

PR #593, opened 2022-10-21
Fix AC key mangling for grpc server

amartani:
https://github.com/buchgr/bazel-remote/pull/339 implemented AC key mangling based on instance name. However, that implementation missed applying the mangling logic on grpc UpdateActionResult method, causing cached entries to not match the ones fetched by GetActionResult. Correctly implement the feature on that method and add test to verify the new behavior.

### DIFF
--- a/server/grpc_test.go
+++ b/server/grpc_test.go
@@ -30,7 +30,7 @@ import (
 	"github.com/buchgr/bazel-remote/cache"
 	"github.com/buchgr/bazel-remote/cache/disk"
 	"github.com/buchgr/bazel-remote/cache/disk/casblob"
-	"github.com/buchgr/bazel-remote/utils"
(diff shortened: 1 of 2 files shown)


server/grpc_test.go
```diff
@@ -92,7 +92,7 @@ func TestMain(m *testing.M) {
 	listener = bufconn.Listen(bufSize)
 
 	validateAC := true
-	mangleACKeys := false
+	mangleACKeys := true
```
mostynb, line 95: This now means that we aren't testing the regular non-mangled AC key case. I will refector the grpc tests in #594 to avoid this.

mostynb (commented):
Thanks for fixing this- I merged your change with some minor reformatting of the commit message. I will follow up with some test refactoring in #594.

PR #595, opened 2022-10-23
always allow health check requests, even when using mTLS

### LINKED ISSUE #590: Unauthenticated health probes and mTLS
With mTLS enabled, `/status` requires a valid client certificate for a successful response, preventing unauthenticated use of `/status` as a health check endpoint. (For example as an http probe in kubernetes.)

Workarounds such as just using a tcp probe, or a command probe that runs curl with a valid client certificate, do work.

Is there a recommended way of doing health checks when mTLS is enabled?

One solution could possibly be to have another endpoint, on another port, that only provides `/status` and does not require mTLS.

mostynb:
Previously if bazel-remote was using mTLS authentication but without `--allow_unauthenticated_reads` then it would use the `tls.RequireAndVerifyClientCert` setting which prevented unauthenticated access to the health check service. We want to always allow access to the health check service regardless of authentication status.

Fixes #590.

### DIFF
--- a/config/tls.go
+++ b/config/tls.go
@@ -27,19 +27,16 @@ func (c *Config) setTLSConfig() error {
 			return fmt.Errorf("Error reading certificate/key pair: %w", err)
 		}
 
-		var cat tls.ClientAuthType = tls.RequireAndVerifyClientCert
-		if c.AllowUnauthenticatedReads {
-			// This allows us to handle some requests without a valid client
-			// certificate, but then we need to explicitly check for verified
-			// certs on requests that we require auth for.
-			// See server.checkGRPCClientCert and httpCache.hasValidClientCert.
-			cat = tls.VerifyClientCertIfGiven
-		}
-
 		c.TLSConfig = &tls.Config{
 			Certificates: []tls.Certificate{readCert},
 			ClientCAs:    caCertPool,
-			ClientAuth:   cat,
+
+			// This allows us to handle some requests without a valid client
+			// certificate (like the grpc health check service), but then we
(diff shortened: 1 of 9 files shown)


config/tls.go
```diff
@@ -27,19 +27,16 @@ func (c *Config) setTLSConfig() error {
...
 			Certificates: []tls.Certificate{readCert},
 			ClientCAs:    caCertPool,
-			ClientAuth:   cat,
+
+			// This allows us to handle some requests without a valid client
+			// certificate (like the grpc health check service), but then we
+			// need to explicitly check for verified certs on requests that
+			// we require auth for.
+			// See server.checkGRPCClientCert and httpCache.hasValidClientCert.
+			ClientAuth: tls.VerifyClientCertIfGiven,
```
mostynb, line 39: This change now exposes metrics (if enabled) and the /status http endpoint to unauthenticated clients. I need to think this through a bit before landing.

main.go
```diff
@@ -237,17 +237,20 @@ func startHttpServer(c *config.Config, httpServer **http.Server,
 		WriteTimeout: c.HTTPWriteTimeout,
 	}
 
+	checkClientCertForReads := c.TLSCaFile != ""
 	checkClientCertForWrites := c.TLSCaFile != "" && c.AllowUnauthenticatedReads
```
mostynb, line 241: These two lines need adjustment.
* checkClientCertForReads should be false if c.AllowUnauthenticatedReads is true.
* checkClientCertForWrites should be true if c.TLSCaFile is set.

Issue #605, opened 2022-11-21
Several openssl and glibc vulnerabilities reported by snyk in the docker image


bu3:
Hi,
I was trying to use the docker image from Dockerhub but the Snyk scanner that runs in my organization raised several vulnerabilities related to `debian:11`.
It sounds like the base image is quite outdated.

I've uploaded the scanner report here:
[snyk.report.zip](https://github.com/buchgr/bazel-remote/files/10056604/snyk.report.zip)

I couldn't find the Dockerfile in the repo, otherwise, I'd have tried to upgrade it by self and then send a PR.

mostynb:
Hi, thanks for the report.

The bazel-remote container images aren't built using a Dockerfile, they're built using the `go_image` target (I'm unsure if that's the right name) from https://github.com/bazelbuild/rules_docker - I'll try upgrading that to see if it helps: https://github.com/buchgr/bazel-remote/pull/606

mostynb:
The snyk report mentions vulnerabilities in openssl and glibc. These libraries seem to come from the gcr.io/distroless/base-debian11 container image that rules_docker uses for go_image's with cgo enabled. So I wonder if the rules_docker upgrade mentioned above upgraded these components- could you try rescanning docker.io/buchgr/bazel-remote-cache:latest ?

Note that even though openssl is included in this image, bazel-remote doesn't use it (go has its own TLS implementation).

mostynb:
I added your changes to the master branch, but as I mentioned in the other PR I don't think it will help this issue.

While I haven't figured out how to upgrade the distroless base image yet, I have found a reference that seems to agree with the diagnosis: https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md confirms that gcr.io/distroless/static is used for go apps that use cgo, and libssl + openssl are included.

It might be that we can avoid the warnings if we can find a suitable base image with recent glibc but without openssl and libssl, and specify that in the `go_image`. If you want a quick fix, you could probably create a minimal Dockerfile with suitable base and use the standalone release binary (https://github.com/buchgr/bazel-remote/releases/download/v2.3.9/bazel-remote-2.3.9-linux-x86_64).

mostynb:
Hmm: there's work-in-progress in distroless to create "base-nossl" images: https://github.com/GoogleContainerTools/distroless/issues/1098

bu3:
Wow. That's good news @mostynb. I might be able to convince security by having them take a look at that issue. Thanks a million

[...]

PR #611, opened 2022-11-26
Try to use a cgo base container image without openssl

mostynb:
Since we enabled cgo, the docker container images switched to use a distroless base image with glibc and openssl. We want to avoid thinking about openssl security issues/upgrades, especially since bazel-remote doesn't use openssl (go has its own TLS implementation).

There is work-in-progress distroless PR to add "base-nossl-debian11" images, which do not include openssl. Let's try it out:
https://github.com/GoogleContainerTools/distroless/issues/1098

Here are two test images, amd64:
https://hub.docker.com/layers/mostynb/bazel-remote-cache/no-openssl-amd64/images/sha256-a66cf03d5ad1b766018bdb90190840e1f174e13aeee89dc9545c97cb88304d74?context=repo
And arm64:
https://hub.docker.com/layers/mostynb/bazel-remote-cache/no-openssl-arm64/images/sha256-dfa958aa54826b3f2d42b9cefe6314c9694099f4a516ee9c6f557c503ddd11e6?context=repo

Relates to #605.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -102,6 +102,7 @@ BAZEL_REMOTE_USER_ID = 65532
 
 go_image(
     name = "bazel-remote-base",
+    base = "@cgo_amd64_base//image",  # Does not include openssl.
     embed = [":go_default_library"],
     goarch = "amd64",
     goos = "linux",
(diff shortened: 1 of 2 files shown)


BUILD.bazel
```diff
@@ -182,8 +184,8 @@ container_push(
     format = "Docker",
     image = ":bazel-remote-image",
     registry = "index.docker.io",
-    repository = "buchgr/bazel-remote-cache",
-    tag = "tmp-amd64",
+    repository = "mostynb/bazel-remote-cache",
+    tag = "no-openssl-amd64",
```
mostynb, line 188: Don't merge these :)

Issue #638, opened 2023-02-10
Reject requests instead of crashing when overloaded


ulrfa:
We experience bazel-remote crashes when local SSD is overloaded by write operations. It would be preferable if bazel-remote would reject requests instead of crashing.

The crash message contains: `runtime: program exceeds 10000-thread limit` and among the listed goroutines, there are almost 10000 stack traces containing either diskCache.Put or diskCache.removeFile, doing syscalls. They originate from both HTTP and gRPC requests.

What do you think about rejecting incoming requests when reaching a configurable number of max concurrent diskCache.Put invocations? Or are there better ways?

We have not experienced overload from read requests, the network interface is probably saturated before that would happen. Therefore, it seems beneficial to still allow read requests and get cache hits even when writes are rejected.

I’m uncertain about proper relation between the semaphore constant (currently 5000) for the diskCache.removeFile, the number of allowed diskCache.Put and Go’s default 10 000 operating system threads. Would it make sense to set one limit for the sum of diskCache.removeFile and diskCache.Put? Should bazel-remote support tuning https://pkg.go.dev/runtime/debug#SetMaxThreads?

It is important to not reject or hinder Prometheus HTTP requests, because the metrics are even more important in overload conditions.

mostynb:
This is an interesting problem.

It would be useful for me (and potentially for other large users) if you could share some details of your storage, the normal load, and the spikes you're seing (either here, or via email if you prefer not to share this info publicly). Does the client handle the service outage gracefully? And do you have bazel-remote configured to restart and hopefully recover from the overload?

The first things to suggest if you want to maintain performance would be:
* Switch to faster storage that might be able to handle the load without falling behind.
* Partition the cache (eg separate cache with separate storage for each platform, or for each release branch, or similar).

If the storage is falling behind due to the number of requests, then I doubt that increasing runtime/debug.SetMaxThreads would help for long (though it probably depends on the duration of the spikes). You're welcome to experiment with this, of course (it would be a one-line patch to try a higher limit).

[...]

PR #639, opened 2023-02-12
Reject high-cost requests instead of creating more OS threads when overloaded

mostynb:
We have been using a file removal semaphore with weight 5,000 (half of Go's default 10,000 maximum OS threads, beyond which Go will crash), in an attempt to avoid crashing when the filesystem/storage layer can't keep up with our requirements.

This change renames that semaphore to `diskWaitSem` and also uses it for disk-write operations. When the semaphore cannot be acquired for disk-writes, we return HTTP 503 (service unavailable) or gRPC RESOURCE_EXHAUSTED error codes to the client.

Relates to #638

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -77,8 +77,9 @@ type diskCache struct {
 	accessLogger     *log.Logger
 	containsQueue    chan proxyCheck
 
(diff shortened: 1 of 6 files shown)


mostynb:
@ulrfa: something for you to try out...

PR #662, opened 2023-05-24
feat(config,cache/s3proxy): add s3 addressing style support

wuhuizuo:
Some S3 providers do not support auto bucket lookup type, need to set it with `path` type or `dns` type. 

Here I added a support to set it in `s3_proxy` filed in configuration to make it reachable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,10 @@ OPTIONS:
    --s3.bucket value The S3/minio bucket to use when using S3 proxy backend.
       [$BAZEL_REMOTE_S3_BUCKET]
 
+   --s3.bucket_lookup_type value The S3/minio bucket lookup type to use when
+      using S3 proxy backend.  (default: auto, types: auto|dns|path)
+      [$BAZEL_REMOTE_S3_BUCKET_LOOKUP_TYPE]
+
    --s3.prefix value The S3/minio object prefix to use when using S3 proxy
       backend. [$BAZEL_REMOTE_S3_PREFIX]
 
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ OPTIONS:
       [$BAZEL_REMOTE_S3_BUCKET]
 
    --s3.bucket_lookup_type value The S3/minio bucket lookup type to use when
-      using S3 proxy backend.  (default: auto, types: auto|dns|path)
+      using S3 proxy backend.  (default: auto, Allowed values: auto, dns, path.)
       [$BAZEL_REMOTE_S3_BUCKET_LOOKUP_TYPE]
 
    --s3.prefix value The S3/minio object prefix to use when using S3 proxy
@@ -466,6 +466,7 @@ http_address: 0.0.0.0:8080
 #  bucket: test-bucket
 #  prefix: test-prefix
 #  disable_ssl: true
+#  bucket_lookup_type: auto
 #
 # Provide exactly one auth_method (access_key, iam_role, or credentials_file) and accompanying configuration.
 #
--- a/cache/s3proxy/s3proxy.go
+++ b/cache/s3proxy/s3proxy.go
@@ -49,6 +49,7 @@ func New(
 	// S3CloudStorageConfig struct fields:
 	Endpoint string,
 	Bucket string,
+	BucketLookupType minio.BucketLookupType,
 	Prefix string,
 	Credentials *credentials.Credentials,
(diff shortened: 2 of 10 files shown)


cache/s3proxy/s3proxy.go
```diff
@@ -49,6 +49,7 @@ func New(
 	// S3CloudStorageConfig struct fields:
 	Endpoint string,
 	Bucket string,
+	BucketLookupType minio.BucketLookupType,
```
mostynb, line 52: Could you add a command line flag for this too, along with the other `--s3.*` options?

mostynb (commented):
Thanks for the contribution. Sorry it took me a few days to look at this.

cache/s3proxy/s3proxy.go
wuhuizuo: Added, PTAL.

wuhuizuo:
> Thanks for the contribution. Sorry it took me a few days to look at this.

I fixed it.

utils/flags/flags.go
```diff
@@ -219,6 +219,12 @@ func GetCliFlags() []cli.Flag {
 			Usage:   "The S3/minio bucket to use when using S3 proxy backend.",
 			EnvVars: []string{"BAZEL_REMOTE_S3_BUCKET"},
 		},
+		&cli.IntFlag{
```
mostynb, line 222: Most of bazel-remotes other flags take string values, would you mind changing this to do that? We could use the values "dns", "path" and "auto". I think that's a bit easier to understand when reading a config file or command line arguments.

mostynb (changes requested):
This feature looks good to me, but I think we should use string flags instead, for improved readability.

I had to do some background reading to understand the use case. This blog post was useful, sharing it here for anyone reading through this PR later: https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

utils/flags/flags.go
wuhuizuo: Yes, your suggestion fine. I updated it.

config/proxy.go
```diff
@@ -78,3 +80,14 @@ func (c *Config) setProxy() error {
...
+
+func parseBucketLookupType(typeStr string) minio.BucketLookupType {
+	valMap := map[string]minio.BucketLookupType{
+		"auto": minio.BucketLookupAuto,
+		"dns":  minio.BucketLookupDNS,
+		"path": minio.BucketLookupPath,
+	}
+
+	// also when not found the type, return "auto" type.
+	return valMap[typeStr]
```
mostynb, line 92: I think this function should probably check if `typeStr` is in the map, and return an error if it isn't. And this error checking should be done in `validateConfig()` in config/config.go

utils/flags/flags.go
```diff
@@ -219,6 +219,12 @@ func GetCliFlags() []cli.Flag {
 			Usage:   "The S3/minio bucket to use when using S3 proxy backend.",
 			EnvVars: []string{"BAZEL_REMOTE_S3_BUCKET"},
 		},
+		&cli.StringFlag{
+			Name:    "s3.bucket_lookup_type",
+			Value:   "auto",
+			Usage:   "The S3/minio bucket lookup type(auto|dns|path) to use when using S3 proxy backend.",
```
mostynb, line 225: nitpick: could you reword this to something like: `The S3/minio bucket lookup type. Allowed values: auto, dns, path.` and update the help text in README.md to match?

README.md
```diff
@@ -263,6 +263,10 @@ OPTIONS:
    --s3.bucket value The S3/minio bucket to use when using S3 proxy backend.
       [$BAZEL_REMOTE_S3_BUCKET]
 
+   --s3.bucket_lookup_type value The S3/minio bucket lookup type to use when
```
mostynb, line 266: github won't let me comment on lines that aren't close to the dirr, but could you also add bucket_lookup_type to the example s3_proxy yaml further down in this file?

config/proxy.go
wuhuizuo: I think we only need to check it in `validateConfig()`.

wuhuizuo: Also I refactored it again.

mostynb:
Thanks for following through with these changes- I made a few minor tweaks and landed this on the master branch.

PR #665, opened 2023-06-16
grpc-asset: fix error format argument order.

PI-Victor:


mostynb (approved):
Thanks- I wish this was a compile error.

PR #676, opened 2023-07-17
Http proxy returns content length when uncompressed

AlessandroPatti:
This is already the case for s3 and azureblob proxy but not for http (and gcs since it's a wrapper around http proxy)


mostynb:
Thanks for the contribution- this looks good.

PR #679, opened 2023-07-24
Support HTTP proxy with mTLS

AlessandroPatti:
Support passing a key and certificate pair to enable mTLS with the http proxy backend

### DIFF
--- a/config/proxy.go
+++ b/config/proxy.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


config/proxy.go
```diff
@@ -32,6 +33,24 @@ func (c *Config) setProxy() error {
...
+			if err != nil {
+				return err
+			}
+
+			config := &tls.Config{
+				Certificates: []tls.Certificate{readCert},
+			}
+
+			tr := &http.Transport{TLSClientConfig: config}
+			httpClient = &http.Client{Transport: tr}
```
mostynb, line 51: Could we do `httpClient.Transport = tr` here instead of creating a new `HttpClient`?

utils/flags/flags.go
```diff
@@ -189,6 +189,18 @@ func GetCliFlags() []cli.Flag {
 			Usage:   "The base URL to use for a http proxy backend.",
 			EnvVars: []string{"BAZEL_REMOTE_HTTP_PROXY_URL"},
 		},
+		&cli.StringFlag{
+			Name:    "http_proxy.key_file",
+			Value:   "",
+			Usage:   "Path to the key used to autheticate with the proxy backend. Enables mTLS.",
```
mostynb, line 195: The help text for these two new options should each mention that the other must also be specified.

mostynb (commented):
Thanks for the feature submission. This looks pretty good, just a few comments...

Once the help text is finalized, we will need to update the help text that's in README.md.

We should also add a test for this feature. We could try to add one to .bazelci/system-test.sh, but it already runs slowly on bazelci/buildkite, so it might be time to refactor that and split it into more jobs. I will try to have a look at that tomorrow.

mostynb:
Merged with some minor adjustments to the help text. I'll have a think about how best to test this, it's probably doable in .bazelci/system-test.sh after some refactoring, now that I have sped it up.

Thanks!

Issue #524, opened 2022-02-11
Can not http proxy to another bazel-remote instance


andrewhamon:
### Steps to reproduce

1. Set up two `bazel-remote` instances, with one http proxying to the other:

**bazel remote 1:**

```
bazel-remote --dir /tmp/bazel_remote_1
```

**bazel remote 2:**

```
bazel-remote --dir /tmp/bazel_remote_2 --http_address localhost:8081 --grpc_address localhost:9093 --http_proxy.url http://localhost:8080
```

2. Run bazel against the second bazel remote:

```
build --remote_cache=grpc://localhost:9093 //path_to:target
```

3. Observe tons of failed requests on the first `bazel-temote`:

<img width="938" alt="image" src="https://user-images.githubusercontent.com/5134584/153542003-7c38172d-52f8-45fd-b504-028fc589ed5d.png">

It seems like when attempting to perform HTTP proxying, bazel-remote tries to read and write to `/cas.v2/`, however `bazel-remote` rejects these as bad requests. All these hashes do already exist on the filesystem, and if I manually do a curl for the same hash but at `/cas/`, then things work as expected.

### Why am I trying this anyway

I'm somewhat interested in setting up a local caching layer for devs at my company, proxying to a shared cache for all developers. This is because a locally running bazel cache, once its cache is filled, is substantially faster.

mostynb:
Hi, this is a (hopefully temporary) limitation of the default compressed storage mode on the instance 2. If you add `--storage_mode uncompressed` to instance 2 (and optionally to instance 1) then I think this should work.

At some point I would like to make this work with compressed storage (and transfer compressed blobs between the two instances).

andrewhamon:
Thanks for the quick reply @mostynb, that makes sense.

This does lead to a few more questions on my part.

The readme says:

> Uploaded CAS blobs are stored in a zstandard compressed format by default, which can increase the effective cache size and reduce load on the server if clients also download blobs in zstandard compressed form.

But I noticed that [`--experimental_remote_cache_compression`](https://docs.bazel.build/versions/main/command-line-reference.html#flag--experimental_remote_cache_compression) in bazel defaults to `false`.

Does that mean that the default `--storage_mode zstd` actually increases, rather than decreases, server load for bazel clients with default configurations, since the server must decompress from storage (rather than pass the bytes straight through)?

mostynb:
[...]

PR #680, opened 2023-07-24
Add grpc proxy backend

AlessandroPatti:
This adds a proxy backend that uses the grpc protocol. When the size of the blobs are unknown (i.e. when using the bazel-remote is used with the REST api), the proxy relies on the asset api to get the blob size, so in order to use another bazel-remote instance as backend `--experimental_remote_asset_api` must be provided to the proxy.
Tested mostly with the added unit test and some local testing done building bazel-remote itself. These are the step used to test locally (each command executed in a different terminal)
```
$ rm -rf /tmp/proxy && bazel run //:bazel-remote -- --dir /tmp/proxy -max_size 5 -port 8080 -grpc_port 9090 --storage_mode uncompressed --experimental_remote_asset_api --tls_cert_file=/tmp/certs/server/cert.crt --tls_key_file=/tmp/certs/server/key.key --tls_ca_file=/tmp/certs/client/cert.crt
$ rm -rf /tmp/cache1 && bazel run //:bazel-remote -- --dir /tmp/cache1 --max_size 5 -port 8081 -grpc_port 9091 --storage_mode uncompressed --grpc_proxy.url localhost:9090 --grpc_proxy.cert_file=/tmp/certs/client/cert.crt --grpc_proxy.key_file=/tmp/certs/client/key.key
$ rm -rf /tmp/cache2 && bazel run //:bazel-remote -- --dir /tmp/cache2 --max_size 5 -port 8082 -grpc_port 9092 --storage_mode uncompressed --grpc_proxy.url localhost:9090 --grpc_proxy.cert_file=/tmp/certs/client/cert.crt --grpc_proxy.key_file=/tmp/certs/client/key.key
$ bazel clean && bazel build //:bazel-remote --disk_cache= --remote_cache=grpc://localhost:9091 --remote_upload_local_results && bazel clean && bazel build //:bazel-remote --disk_cache= --remote_cache=grpc://localhost:9092 --remote_upload_local_results --remote_accept_cached
```


cache/grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,287 @@
...
+
+// Helper function for logging responses
+func logResponse(logger cache.Logger, method string, msg string, resource string) {
+	logger.Printf("GRPC PROXY %s %s: %s", method, msg, resource)
+}
+
+func (r *remoteGrpcProxyCache) UploadFile(item backendproxy.UploadReq) {
+	switch item.Kind {
+	case cache.RAW:
+		fallthrough
```
mostynb, line 71: I think this case deserves a comment, and we can maybe upgrade that to a log message later.

utils/flags/flags.go
```diff
@@ -183,6 +183,24 @@ func GetCliFlags() []cli.Flag {
...
+		&cli.StringFlag{
+			Name:    "grpc_proxy.url",
+			Value:   "",
+			Usage:   "The base URL to use for a grpc proxy backend.",
+			EnvVars: []string{"BAZEL_REMOTE_GRPC_PROXY_URL"},
+		},
+		&cli.StringFlag{
+			Name:    "grpc_proxy.key_file",
+			Value:   "",
+			Usage:   "Path to the key used to autheticate with the proxy backend. Enables mTLS.",
```
mostynb, line 195: The `key_file` / `cert_file` help texts should each mention that using one requires the other.

cache/grpcproxy/BUILD.bazel
```diff
@@ -0,0 +1,41 @@
...
+        "grpcproxy.go",
+        "readcloser.go",
+    ],
+    importpath = "github.com/buchgr/bazel-remote/v2/cache/grpcproxy",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//cache:go_default_library",
+        "//genproto/build/bazel/remote/asset/v1:go_default_library",
+        "//genproto/build/bazel/remote/execution/v2:go_default_library",
+        "//utils/backendproxy:go_default_library",
```
mostynb, line 15: I think this code is missing from your PR...

cache/grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,287 @@
...
+
+	return proxy, nil
+}
+
+// Helper function for logging responses
+func logResponse(logger cache.Logger, method string, msg string, resource string) {
+	logger.Printf("GRPC PROXY %s %s: %s", method, msg, resource)
+}
+
+func (r *remoteGrpcProxyCache) UploadFile(item backendproxy.UploadReq) {
```
mostynb, line 68: I can't see where `item.Rc` being closed anywhere?

mostynb (commented):
Thanks for starting on this feature- it has been on my wishlist for a while :)

I will try to find time for a thorough review tomorrow, but I have left some initial comments in the meantime (I think your PR is missing utils/backendproxy/ btw).

cache/grpcproxy/BUILD.bazel
AlessandroPatti: This is existing code, it is already used by all the proxy implementations

cache/grpcproxy/grpcproxy.go
AlessandroPatti: Great catch! Added

cache/grpcproxy/BUILD.bazel
mostynb: D'oh ;)

cache/grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,316 @@
...
+			logResponse(r.errorLogger, "GetActionResult", err.Error(), digest.Hash)
+			return nil, -1, err
+		}
+
+		logResponse(r.accessLogger, "GetActionResult", "Success", digest.Hash)
+		return io.NopCloser(bytes.NewReader(data)), int64(len(data)), nil
+
+	case cache.CAS:
+		// We don't know the size, so send a FetchBlob request first to get the digest
+		// TODO: consider passign the expected blob size to the proxy?
```
mostynb, line 220: This has been one of the complicating factors that I have been putting off dealing with in order to add a grpc proxy backend...

The first proxy backends were added to bazel-remote before it had gRPC support, and were designed for bazel's simpler remote cache REST API. In particular, the REST api only specifies the hash but not the size of CAS blobs, whereas the gRPC API uses hash/size digests.

One way to handle this, would be to add an optional size argument to the Proxy functions, pass -1 if unknown, and let the proxy backend use the value if non-negative.

Another, probably more complicated option, would be to try refactoring a little, to add an alternate path for gRPC proxy backends from the gRPC code. While the setup might be more complicated, it could probably be more efficient, because there would be fewer calls (eg FindMissingBlobs, BatchUpdateBlobs and BatchReadBlobs wouldn't require multiple calls to the backend). I am  not sure what this would look like- but I think it's worth analysing before accepting this feature.

cache/grpcproxy/grpcproxy_test.go
```diff
@@ -0,0 +1,203 @@
...
+	cc      *grpc.ClientConn
+	server  *grpc.Server
+	clients *GrpcClients
+	cache   disk.Cache
+}
+
+func newFixture(t *testing.T, proxy cache.Proxy) *fixture {
+	port := rand.Intn(65353-1024) + 1024
+	host := fmt.Sprintf("localhost:%d", port)
+	cc, err := grpc.Dial(host, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
mostynb, line 40: Could we use google.golang.org/grpc/test/bufconn for this instead? See the example in server/grpc_test.go

cache/grpcproxy/readcloser.go
```diff
@@ -0,0 +1,57 @@
+package grpcproxy
+
+import "io"
+
+type StreamReadCloser[M DataMessage] struct {
```
mostynb, line 5: Introducing generics to bazel-remote means that I need to find some time to do some background reading before I can understand this code...

config/config.go
```diff
@@ -348,6 +361,17 @@ func validateConfig(c *Config) error {
 		}
 	}
 
+	if c.GRPCBackend != nil {
+		if c.StorageMode != "uncompressed" {
+			return errors.New("--grpc_proxy does not yet support compressed storage mode")
```
mostynb, line 366: Lacking support for this mode is unfortunate, it could make the gRPC proxy backend less efficient than HTTP.

cache/grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,316 @@
...
+		decoded, err := hex.DecodeString(hash)
```
mostynb, line 221: Why decode the string, just to encode it again below?

cache/grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,316 @@
...
+		// is to get the object and discard the result
+		// We don't expect this to ever be called anyways since it is not part of the grpc protocol
+		rc, size, err := r.Get(ctx, kind, hash)
+		rc.Close()
+		if err != nil || size < 0 {
+			return false, -1
+		}
+		return true, size
+	case cache.CAS:
+		decoded, err := hex.DecodeString(hash)
```
mostynb, line 281: Why decode the string, just to encode it again below?

cache/grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,316 @@
...
+		// TODO: consider passign the expected blob size to the proxy?
+		decoded, err := hex.DecodeString(hash)
+		if err != nil {
+			return nil, -1, err
+		}
+		q := asset.Qualifier{
+			Name:  "checksum.sri",
+			Value: fmt.Sprintf("sha256-%s", base64.StdEncoding.EncodeToString(decoded)),
+		}
+		freq := asset.FetchBlobRequest{
```
mostynb, line 229: I find this a bit confusing, why is this using the remote asset API instead of BatchReadBlobs or the bytestream API?

cache/grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,316 @@
...
+			Uris:       []string{},
+			Qualifiers: []*asset.Qualifier{&q},
+		}
+
+		res, err := r.clients.asset.FetchBlob(ctx, &freq)
```
mostynb, line 295: Why use the remote asset API instead of FindMissingBlobs?

mostynb (commented):
Sorry it has taken me a while to get back to this.

I think the most important thing to do next is see my comment about investigating more direct gRPC->gRPC backend proxying.

config/config.go
AlessandroPatti: Yes, although it depends on whether clients use compression or not. Need to spend some time to figure out how to efficiently get the raw blob (including the header) from a proxied bazel-remote instance instead having to recompress it on the fly and did want to block this feature in the meanwhile. Is compression support a prerequisite for this to merge?

cache/grpcproxy/readcloser.go
AlessandroPatti: The generic is not really necessary here since we only use it with one message (`bs.ReadResponse`), I just found it more elegant and easier to reuse. I can remove if you feel this introduces an unwanted precendent.

config/config.go
```diff
@@ -348,6 +361,17 @@ func validateConfig(c *Config) error {
...
+		}
+		if c.GRPCBackend.KeyFile != "" || c.GRPCBackend.CertFile != "" {
+			if c.GRPCBackend.KeyFile == "" || c.GRPCBackend.CertFile == "" {
+				return errors.New("To use mTLS with the grpc proxy, both a key and a certifacte must be provided")
```
mostynb, line 370: nitpick: s/certifacte/certificate/

config/config.go
mostynb: > Is compression support a prerequisite for this to merge?

It's not a prerequisite, but if we don't support it then I think we would need some documentation and/or example setups that explain the issue- in particular, comparing to using a HTTP bazel-remote backend. (And maybe it's better to just land this feature which works optimally, and skip that documentation?)

Ideally I would like things to work nicely when clients use compression, because that seems to be the best tradeoff for performance (CPU time vs download time on the client side, and minimal CPU usage + effectively larger cache size on the server side).

The best case scenario IMO is when the client, frontend bazel-remote and optional proxy backend server all support compression. The frontend bazel-remote instance needs to verify the sha256 of blobs from the client, but it can then store a compressed result from its disk cache (increasing its effective disk cache size) and serve blobs without re-verifying their hashes.

The presence of a proxy backend increases complexity- should the frontend trust results from the backend? If the answer is yes, then the network transfers from the backend to the client are most efficient (a compressed result can be sent from the backend, via the frontend, to the client without recompression). If the answer is no, then CPU load on the frontend increases, because it needs to verify results from the backend.

Circling back to your question of should this PR land without compression support: would this gRPC backend be better than using proxy backend bazel-remote instance using HTTP with compression support? (Or do you have a different setup than I am imagining, is the proxy backend something other than bazel-remote? Could you to describe your use-case for this feature?)

AlessandroPatti: @mostynb Sorry it took me this long to get back on this one. The setup I have in mind is using another bazel-remote as backend, but without compression. From my experience, enabling compression on both client and server (and proxy?) was never better: the overhead from compressing and decompressing each blob on the client side was too much, and the network transfers fast enough to effectively seeing slower builds with compression enabled. Enabling only on server side is also a possibility, which might reduce used storage and increase the effective cache size, but is something that I haven't tried as I suspect it would put too much additional load on the server for too little gain. I guess it all depends on the available resources (Is network fast? Are there a lot of cpus?) and the build profile (how "compressable" are the artifacts?).

Regardless, I've added support for zstd compression, but with a caveat: it does not work with a bazel-remote backend. The reason is that bazel-remote itself expects the backends to return the blobs just like they were pushed by the server (it is actually a bit more subtle, it expects all blobs to start with the [casblob header](https://github.com/buchgr/bazel-remote/blob/8d472a9ea0516720b3f619e2142d878636262bd3/cache/disk/casblob/casblob.go#L32), the content of it can technically change, e.g. use different frame sizes). However bazel remote would always serve the blob after stripping away the header, so it is not behaving as a valid proxy backend. This is true regardless of the protocol used, as of now bazel-remote cannot be used as a http backend for another bazel-remote instace with compression enabled (http additionally suffers from https://github.com/buchgr/bazel-remote/issues/524). 
I though of a few ways to solve this but none of them was great:
  1. Have the grpc (and http?) proxy implementation to decompress and recompress all blobs coming from the backend and add the header, which would probably require writing to file and read back. Overall seems just too much overhead to be an acceptable solution.
  2. Instead of decompressing and recompressing, attempt to recreate the header directly from the compressed stream. The header is mostly used to quickly find the uncompressed offset in the compressed stream, which could be done by decoding each frame and accumulating the [Frame_Content_Size](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#frame_content_size) field of each frame header (provided it is present). This would still require to consume the entire stream and read it back from a temporary storage, but would save recompressing, which is usually the most expensive part.
  3. Somewhat inspired by the above, drop the header all together and find offsets on-request by parsing the frame headers. Not sure if there is something that is currently done that wouldn't be possible this way.
  4. Have bazel-remote stream back the header as well, since they are valid zstd (skippable) frames. For requests where the offeset is != 0, the header should be updated to match the offset of the truncated stream, but that might be fast enough and a rear enough case. The downside is that each request will stream additional bytes, which might be a significant relative increase for small blobs. Maybe this should be put behind a flag so that it can be enabled conditionally for bazel-remote instance that are supposed to be used as backends.

Curious to know if the above makes sense to you or I have completely misread how this works or if you have a better solution in mind. Unless I am missing something obvious, this would need some thinking, and maybe out of scope for this feature request since it would still have feature parity with http?

README.md
```diff
@@ -243,6 +243,19 @@ OPTIONS:
...
+   --grpc_proxy.key_file value Path to a key used to authenticate with the
+      proxy backend using mTLS. If this flag is provided, then
+      grpc_proxy.cert_file must also be specified.
+      [$BAZEL_REMOTE_GRPC_PROXY_KEY_FILE]
+
+   --grpc_proxy.cert_file value Path to a certificate used to authenticate
+      with the proxy backend using mTLS. If this flag is provided, then
+      grpc_proxy.key_file must also be specified.
+      [BAZEL_REMOTE_GRPC_PROXY_CERT_FILE]
+
```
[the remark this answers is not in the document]
mostynb: The bazel-remote <-> proxy backend authentiation is completely separate from the client <-> bazel-remote authentication though. If bazel-remote is the main intended use case for the grpc proxy backend, then I think we should support that. But this would be OK to land in a followup (I can help if needed).

README.md
```diff
@@ -243,6 +243,20 @@ OPTIONS:
       Goroutines to process parallel uploads to backend. (default: 100)
       [$BAZEL_REMOTE_NUM_UPLOADERS]
 
+   --grpc_proxy.url value The base URL to use for a grpc proxy backend, e.g.
+      localhost:9090 or example.com:7070.
+      [$BAZEL_REMOTE_GRPC_PROXY_URL]
```
mostynb, line 248: I think we should probably provide a way to use TLS when talking to the proxy backend and not using mTLS. It looks like this PR uses TLS when authenticating with the proxy backend via mTLS (which makes sense), otherwise TLS is not used at all when talking to the proxy backend?

cache/grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,316 @@
...
+)
+
+type GrpcClients struct {
+	asset asset.FetchClient
+	bs    bs.ByteStreamClient
+	ac    pb.ActionCacheClient
+	cas   pb.ContentAddressableStorageClient
+}
+
+func NewGrpcClients(cc *grpc.ClientConn) *GrpcClients {
```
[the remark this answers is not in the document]
mostynb: Looks like the return value is also used in the config package now too.

README.md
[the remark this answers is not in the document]
AlessandroPatti: I did not realise that the http proxy actualy supports basic auth, just need to pass the credentials in the proxy url. I took a stab at doing the same for the grpc proxy.

mostynb:
> Hi @mostynb, any blocker on merging this?

HI, sorry this is taking me a while. I wanted to make a release before landing this, because I will probably want to make a few changes afterwards. That is done now (after a small hiccup in 2.4.2 and a rushed out 2.4.3 release).

I will try to find time to look at this again tomorrow- thanks for persevering.

AlessandroPatti:
@mostynb any update on this one?

mostynb:
I finally got back to reviewing this, sorry for the delay. I will continue tomorrow.

mostynb:
I landed a squashed version of this- I think it will be easier to iterate on this on master with a few smaller PRs than to drag this one on any longer. Thanks again for being so patient.

PR #691, opened 2023-08-29
Fix serving uncompressed blob as compressed

AlessandroPatti:
The pipe is created but never written to nor closed. Instead, the file is passed as writer to the encoder, causing errors like the following when serving zstd compressed blob when using uncompressed storage:
```
2023/08/29 18:58:44 Error while closing encoder: write /private/tmp/cache/cas.v2/4d/4d42748448b9cf8484bfea1e5df65e33213fcc85bba393ae4f3ebebd627523c0-144750847.v1: bad file descriptor
2023/08/29 18:58:44 Error writing cas/4d42748448b9cf8484bfea1e5df65e33213fcc85bba393ae4f3ebebd627523c0 err: write /private/tmp/cache/cas.v2/4d/4d42748448b9cf8484bfea1e5df65e33213fcc85bba393ae4f3ebebd627523c0-144750847.v1: bad file descriptor
2023/08/29 18:58:44  GET 500       127.0.0.1 /cas/4d42748448b9cf8484bfea1e5df65e33213fcc85bba393ae4f3ebebd627523c0
```

### DIFF
--- a/cache/disk/casblob/BUILD.bazel
+++ b/cache/disk/casblob/BUILD.bazel
@@ -11,4 +11,9 @@ go_library(
 go_test(
     name = "go_default_test",
     srcs = ["casblob_test.go"],
+    deps = [
+        ":go_default_library",
+        "//cache/disk/zstdimpl:go_default_library",
+        "//utils:go_default_library",
+    ],
 )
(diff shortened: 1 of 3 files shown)


mostynb (approved):
Thanks for fixing this!

PR #696, opened 2023-09-01
Improving file eviction performance

ulrfa:


server/grpc_ac.go
```diff
@@ -171,6 +171,12 @@ func (s *grpcServer) maybeInline(ctx context.Context, inline bool, slice *[]byte
 			return nil // Not inlined, nothing to do.
 		}
 
+		if (*digest).SizeBytes <= 0 {
+			// Unexpected corner case?
+			*slice = []byte{}
+			return nil
+		}
+
 		if *digest == nil {
```
ulrfa, line 180: I'm uncertain what your intention was here @mostynb?

It looks suspicious to first use (*digest).SizeBytes and then a few lines further down check for if *digest == nil.

cache/disk/disk.go
```diff
@@ -645,6 +644,16 @@ func (c *diskCache) get(ctx context.Context, kind cache.EntryKind, hash string,
 		return nil, -1, nil
 	}
 
+	// Non-proxied Get requests do not seem to consume any significant amount of OS threads,
+	// and are therefore not throttled. However, it is assumed that proxied Get requests might,
+	// at least when storing the result from the proxy to disk, and perhaps also when
```
ulrfa, line 649: I have been using all the other parts of #696 in production for long time. But not this part 4 commit for proxies, because I'm not using proxies myself.

ulrfa:
@mostynb, what do you think about this?

Is anything more needed from me to land it?

I can add test cases if you are OK with the overall approach.

server/grpc_ac.go
mostynb: It looks your your PR adds this?

cache/disk/disk.go
```diff
@@ -286,11 +311,7 @@ func (c *diskCache) Put(ctx context.Context, kind cache.EntryKind, hash string,
 		}
 		if !ok {
 			c.mu.Unlock()
-			return &cache.Error{
-				Code: http.StatusInsufficientStorage,
-				Text: fmt.Sprintf("The item (%d) + reserved space is larger than the cache's maximum size (%d).",
-					size, c.lru.MaxSize()),
-			}
+			return ErrOverloaded
```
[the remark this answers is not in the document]
ulrfa: Note that the removed code was not reachable, because err!=nil was caught a few lines above and lru.Reserve always returned an err when reservation was rejected.

Also note that diskCache.Put converted all kinds of errors reported by lru.Reserve to the code http.StatusInsufficientStorage, including errors about invalid negative blob size.

I’m not sure where you intend the check for http.StatusInsufficientStorage to be placed? I today added such check inside the gRPCErrCode conversion function in grpc.go and that replaced the need for ErrOverloaded.

Please take a look at my additional commit https://github.com/buchgr/bazel-remote/pull/696/commits/bdeb3dc6ead5e92e306117d91c41b184b1df3007 from today in #696, which aims to address those notes and your comment. You can find more details in the commit message.

Do you think my change of the lru.Reserve function signature is OK, or do you prefer it some other way?

utils/flags/flags.go
```diff
@@ -51,6 +51,12 @@ func GetCliFlags() []cli.Flag {
 			Usage:   "The maximum size of bazel-remote's disk cache in GiB. This flag is required.",
 			EnvVars: []string{"BAZEL_REMOTE_MAX_SIZE"},
 		},
+		&cli.Int64Flag{
+			Name:    "disk_size_limit",
+			Value:   -1,
+			Usage:   "The maximum size of bazel-remote's disk cache, including files queue for eviction. in GiB. Limit is disabled by default.",
+			EnvVars: []string{"BAZEL_REMOTE_DISK_SIZE_LIMIT"},
+		},
```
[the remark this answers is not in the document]
ulrfa: What do you think about a longer usage text like the following?

> The maximum size of bazel-remote's disk cache, including files queued for eviction, in GiB.
>
> If this limit is reached, write requests will be rejected, allowing clients to decide which requests to retry. Enabling this limit is recommended to prevent filling up the disk in overload situations when new files are uploaded faster than old files can be evicted, potentially leading to unexpected outcomes and disk fragmentation.
>
> By default, this limit is disabled for backward compatibility. It is generally advisable to set disk_size_limit to be 5% larger than max_size. The disk_size_limit can be tuned by watching how the prometheus query `max_over_time(bazel_remote_disk_cache_size_bytes[$__interval])` varies between `bazel_remote_disk_cache_size_bytes_limit{type="evict"}` and `bazel_remote_disk_cache_size_bytes_limit{type="reject"}`.

server/grpc_ac.go
mostynb: If we keep this, I think we should move it below the `if *digest == nil {` block.

utils/flags/flags.go
```diff
@@ -51,6 +51,12 @@ func GetCliFlags() []cli.Flag {
 			Usage:   "The maximum size of bazel-remote's disk cache in GiB. This flag is required.",
 			EnvVars: []string{"BAZEL_REMOTE_MAX_SIZE"},
 		},
+		&cli.Int64Flag{
+			Name:    "disk_size_limit",
```
mostynb, line 55: I think we should find a better name for this, disk_size_limit is too similar to max_size. What do you think about max_size_hard_limit ?

utils/flags/flags.go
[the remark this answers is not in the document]
mostynb: How about we split this up, something short for the flag description:
> If enabled, the maximum size of bazel-remote's disk cache including files queued for eviction in GiB, before bazel-remote will reject write requests. A reasonable value might be 5% larger than `--max_size`.

And then in the example configuration in README.md we can add more details:
```
# If specified, write requests will be rejected when this limit is reached.
# Clients can then decide which requests to retry. This setting can be used
# to avoid running out of disk space when new files are uploaded faster than  
# old files can be evicted. A reaonable value might be 5% larger than 
# max_size.
#
# The disk_size_limit can be tuned by watching how the prometheus query 
# max_over_time(bazel_remote_disk_cache_size_bytes[$__interval]) varies 
# between bazel_remote_disk_cache_size_bytes_limit{type="evict"} and 
# bazel_remote_disk_cache_size_bytes_limit{type="reject"}.
#max_size_hard_limit: 105
```

cache/disk/lru.go
```diff
@@ -307,3 +397,54 @@ func (c *SizedLRU) getTailItem() (Key, lruItem) {
 	}
 	return nil, lruItem{}
 }
+
+// Append an entry to the eviction queue. The entry must have been removed
+// from SizedLRU.ll before being sent to this method.
+// Note that this method can be invoked without holding the diskCache.mu mutex,
```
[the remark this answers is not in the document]
mostynb: sgtm

utils/flags/flags.go
ulrfa: I have been thinking about renaming the original `max_size` to something with the word “target” or “eviction” since bazel-remote is evicting to reach that target, but is not 100% enforcing it as a hard max limit. Examples of pairs:

`target_disk_size`   `max_disk_size`

`disk_size_target`   `disk_size_limit`

`eviction_disk_size`   `rejection_disk_size`

The word “disk” could help emphasize that it is not max request size, not max amount stored in proxy backend, etc. 

However, I assume it is not motivated to do such non-backward-compatible renaming and I think your suggested name `max_size_hard_limit` aligns well with the existing `max_size`.

I’m working on adding unit tests, and **I will update according to your suggestions** when I’m ready with the unit tests.

server/grpc_ac.go
ulrfa: After thinking more about this I believe the corner case can only occur if the ActionResult have **both** an inlined slice (with non-zero length) and **also** a contradicting explicit digest specifying zero length. Because otherwise there would be an early return at the `len(*slice) == 0` check above. I assume invalid digests are managed other ways, and I believe the check should not be added so I reverted it in latest commit.

utils/flags/flags.go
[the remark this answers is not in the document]
ulrfa: Updated in lastest commit.

cache/disk/lru.go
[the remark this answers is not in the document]
ulrfa: Updated in latest commit.

ulrfa:
Tuesday next week (24th of June) will be my last day before I go on vacation. Is there anything you'd like me to do before then? Otherwise, I’ll be back in the middle of August.

Feel free to land or make changes to the PR @mostynb if you prefer while I’m away.

mostynb:
> Tuesday next week (24th of June) will be my last day before I go on vacation. Is there anything you'd like me to do before then? Otherwise, I’ll be back in the middle of August.
> 
> Feel free to land or make changes to the PR @mostynb if you prefer while I’m away.

I am hoping to do a release before we land this, since there's an important bug fix. I will try to land this soon afterwards- thanks for persevering all this time, and enjoy your vacation.

mostynb:
This is included in v2.6.0.

PR #701, opened 2023-09-23
fixup kubernetes example

seankhliao:
* remove the misleading references to nginx
* rename the grpc port to grpc
* fix the env to be proper k8s port spec


mostynb:
Thanks for the cleanup.

PR #704, opened 2023-10-11
[metrics] add maximum disk size as metric

martijndegouw:
This allowes for easier monitoring of how much of the assigned space is used.

### DIFF
--- a/cache/disk/lru.go
+++ b/cache/disk/lru.go
@@ -44,6 +44,7 @@ type SizedLRU struct {
 
 	onEvict EvictCallback
 
+	gaugeCacheMaxSizeBytes  prometheus.Gauge
 	gaugeCacheSizeBytes     prometheus.Gauge
 	gaugeCacheLogicalBytes  prometheus.Gauge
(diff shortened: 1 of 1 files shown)


mostynb:
Hi, I'm not sure how useful it is to monitor this value, since you typically set it relative to the filesystem space you have available and then don't need to think about it. What's the use case for adding this?

martijndegouw:
Hi, The reason for adding it is that the bazel_remote_disk_cache_logical_bytes metric has no reference now. Monitoring does not know how much head room is left for the bazel-remote cache instance. If a more dynamic value is preferred this metric could be changed to bazel_remote_disk_cache_logical_bytes_free for example.

ulrfa:
See also https://github.com/buchgr/bazel-remote/pull/696/commits/81961893732633db52e2ed2214229bc21f120dfa from https://github.com/buchgr/bazel-remote/pull/696 that introduces bazel_remote_disk_cache_size_bytes_limit with the labels "evict" and "reject".

PR #705, opened 2023-10-11
[metrics] initialize hits/misses for each cache entry kind

martijndegouw:
This makes sure all hit/miss counters appear on the metrics endpoints and avoids having to fill missing metrics with zero at the client side.


mostynb:
This slipped off my radar, sorry for the delay. I added this to the master branch.

PR #726, opened 2024-01-03
feat: support use s3 sign v2 by config

### LINKED ISSUE #725: chunk-signature in s3 proxy
Hi,
I'm trying to connect S3 to our bazel-remote service, but I find some confusing thing: 

I manually delete an AC file in my local file system:

image

than i build my project again. In my undertanding, bazel-remote will query S3 to find this ac file, if found, it returns to bazel client. But it seems not: our s3 buckt has this ac file, but S3 client still re-build this action and put this ac file:

image

image

Then I download the ac file in s3 bucket. Below is the file i download from s3.

image

And below is the file in my local:

image

Compared to the file in my local, file in s3 has an unexpected 'chunk-signature'.

Then I catch the ac file that finally return to bazel client, and find it still has this 'chunk-signature'. I guess bazel client cannot recognize this, so it finally re-build this action.

But I didn't find any code to do wit this chunk-signature. Then I search history issue and find #162 mentioned this, but seems no conclusion. Maybe minio imported this?

For more information, below is the config in my test env: 

image

So, my question is: is this a bug or a feature? How can i config to avoid this in order to use s3 as our fallback strategy?
thx a lot.

hyphennn:
This PR support to set s3 sign as v2. Resolves #725 
Now user could set a new flag --s3.disable_v4_sign=true when there auth_method is secret_key, which will lead using minio/pkg/credentials.NewStaticV2 as the initialization of s3 credential. This can support some situation when s3 do not support v4 sign well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,10 +317,10 @@ OPTIONS:
       using S3 proxy backend. Applies to s3 auth method(s): access_key.
       [$BAZEL_REMOTE_S3_SECRET_ACCESS_KEY]
       
-   --s3.disable_v4_sign Whether to disable v4 sign and enable v2 sign when 
-      using the S3 proxy backend. Only effect when use access_key as 
-      auth_method. (default: false, is enable v4 sign) 
-      [$BAZEL_REMOTE_S3_DISABLE_V4_SIGN]
+   --s3.signature_type value Which type of aws signature will be used when 
+      using S3 proxy backend. Applies to s3 auth method(s): access_key. 
+      Allowed values: s3v2, s3v4, s3v4streaming, anonymous. (default: s3v4) 
+      [$BAZEL_REMOTE_S3_SIGNATURE_TYPE]
 
    --s3.aws_shared_credentials_file value Path to the AWS credentials file.
       If not specified, the minio client will default to '~/.aws/credentials'.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -316,6 +316,11 @@ OPTIONS:
    --s3.secret_access_key value The S3/minio secret access key to use when
       using S3 proxy backend. Applies to s3 auth method(s): access_key.
       [$BAZEL_REMOTE_S3_SECRET_ACCESS_KEY]
+      
+   --s3.disable_v4_sign Whether to disable v4 sign and enable v2 sign when 
+      using the S3 proxy backend. Only effect when use access_key as 
+      auth_method. (default: false, is enable v4 sign) 
+      [$BAZEL_REMOTE_S3_DISABLE_V4_SIGN]
```
mostynb, line 323: I think it's reasonable to add a configuration option to specify the AWS signature type, but it should be something that is extensible over time for example in case S3 adds a new AWS signature versions.

So instead of adding a `--s3.disable_v4_sign` flag which means "use v2", maybe we could add a `--s3.signature_version` flag which defaults to `"v4"`, but also supports `"v2"` ?

mostynb (changes requested):

README.md
hyphennn: That's a good idea. I've sumitted a new commit in which I replaced disable_v4_sign with signature_type. It's just the same as minio's SignatureType.

mostynb:
Sorry it took me so long to get back to this. I squashed your commits, then made a small change to the values (eg `v2` instead of `s3v2`), and landed these changes on the master branch.

PR #728, opened 2024-01-04
Add an --idle_timer test

mostynb:
This is expected to fail, until #727 lands.


jchorl:
Thanks for adding!

I'm not sure what the test here does (does it exercise any functionality?). If it only runs the server for 1s, I don't think it'll fail. You need to http query it for a blob (even the empty blob works), which will cause infinite recursion/stack overflow.

mostynb:
Updated with a fix, it fails now as expected.

PR #740, opened 2024-04-02
Support LDAP authentication with unauthorized reads

JonasScharpf:
This PR is based on #101 and resolves #100

It provides unauthorized reads when using LDAP, config via command line args and minimal config file testing for the new LDAP section

This is my first GO thing ever, so it might not be perfect

PR #746, opened 2024-05-10
Upgrade azidentity to v1.5.2

gplessis:
I'm working on setting up `bazel-remote` in a Kubernetes cluster, as a proxy on top of an Azure Storage Account. To avoid injecting static credentials into the deployment, I would like to configure [Entra Workload ID](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview) (tl;dr: the resulting pod is mutated and temporary credentials are injected as environment variables).

Unfortunately, the minimum required version of `azidentity` that supports Workload ID via the `default` authentication method is `v1.3.0`.

This PR updates it to it latest version (`v1.5.2`) from `v1.1.0`.


mostynb (approved):
Thanks for the submission- this looks good.

IIRC the go azure sdk made a bunch of breaking changes since the azblob feature was added here, but I am unable to easily update it since I don't have an azure account to test against. If you would like to try upgrading that, PRs would be welcome.

gplessis:
Sounds good 👍 When do you think the next release will be cut?

mostynb:
I'm trying to help get #740 landed this week, once that's done it might be a good time for a new release. And if it looks like that PR will take too long, then I can probably make a release at the end of this week.

PR #747, opened 2024-05-12
chore: cleanup local build rule and docs

hunshcn:
`oci_image` is not executable


mostynb:
Thanks for the contribution- this has now landed on the master branch.

PR #756, opened 2024-06-06
Add a clearer error

dieend:
I spent 1 hour trying to debug what config was changed when upgrading from old version to newer one. This should give enough clues about where to look for.
I was using the config_file parameter, and the YAML didn't specify `bucket_lookup_type`, which caused the app crashes with  unhelpful text `Unsupported value:`

### DIFF
--- a/config/proxy.go
+++ b/config/proxy.go
@@ -184,7 +184,7 @@ func parseBucketLookupType(typeStr string) (minio.BucketLookupType, error) {
 
 	val, ok := valMap[typeStr]
 	if !ok {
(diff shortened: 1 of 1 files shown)


mostynb:
Looks good- thanks for the improvement.

mostynb (approved):

milesj:
I just spent hours debugging this "Unsupported value" error. I'm assuming this hasn't been released yet...

mostynb:
> I just spent hours debugging this "Unsupported value" error. I'm assuming this hasn't been released yet...

Sorry about that. I released version 2.5.0 just now, which has this fix.

PR #759, opened 2024-06-17
More informative error message in case of corrupted blob headers

Andrius-B:
I would love to get a more informative error message here. I keep seeing this happen but since the error does not contain the hash of the CAS item being fetched, it's very tricky to figure out what goes wrong from the access logs.
Specifically the error I'm seeing is this:
```
Warning: expected item to be on disk, but something happened: offset table values should increase: 0 -> 29
```
I'm hoping that with a hash to reference in the logs, I can find out what corrupts the headers

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -476,7 +476,7 @@ func (c *diskCache) availableOrTryProxy(kind cache.EntryKind, hash string, size 
 				}
 
 				if err != nil {
-					log.Printf("Warning: expected item to be on disk, but something happened: %v", err)
+					log.Printf("Warning: expected item to be on disk, but something happened when retrieving %s (compressed: %v, legacy: %v): %v", blobPath, item.legacy, zstd, err)
 					f.Close()
 				} else {
 					return rc, item.size, false, nil


cache/disk/disk.go
```diff
@@ -476,7 +476,7 @@ func (c *diskCache) availableOrTryProxy(kind cache.EntryKind, hash string, size
 				}
 
 				if err != nil {
-					log.Printf("Warning: expected item to be on disk, but something happened: %v", err)
+					log.Printf("Warning: expected item to be on disk, but something happened when retrieving %s-%d: %v", hash, size, err)
```
mostynb, line 479: Thanks for the contribution.

I'm happy to improve the logging here, but perhaps we can improve this a bit, eg:
* It might be worth logging `blobPath` here instead, like the log in the if case one level up.
* Maybe we could also log something different in each of the cases in the if-else block immediately before this one?

Andrius-B: Hi! I have adjusted the error as suggested and added some additional variables that should make it easy to figure out which kind of `ReadCloser` is being fetched.
In the end my problem was that I was running the program with too little resources and that caused it to fail in various interesting ways. Once I gave it some more CPUs, it started running like expected. Although this particular error message was not very informative so it would be cool to make it better

mostynb:
Thanks for the contribution- this has landed on the master branch.

PR #760, opened 2024-06-21
support HTTP headers supported in bazel 7.1

xuzhenglun:
ref: https://github.com/bazelbuild/bazel/pull/21503/files

Bazel 7.1.0 support passthrough HTTP headers to remote downloader. this PR extend `FetchBlob` and make sure those headers be handled correctly.


mostynb:
This looks good- thanks for the contribution, merged.

PR #770, opened 2024-08-23
Return zst Decoder to sync.pool when finished

tobbe76:
zstd Decoder was not returned to sync.pool so a new instance was allocated each time. This reduced performance.
We can now handle higher loads on the bazel-remote server.

### DIFF
--- a/server/grpc_bytestream.go
+++ b/server/grpc_bytestream.go
@@ -15,8 +15,6 @@ import (
 	"github.com/buchgr/bazel-remote/v2/cache"
 	"github.com/buchgr/bazel-remote/v2/cache/disk/casblob"
 
-	"github.com/klauspost/compress/zstd"
-
 	"github.com/buchgr/bazel-remote/v2/utils/zstdpool"
 	syncpool "github.com/mostynb/zstdpool-syncpool"
 )
@@ -367,12 +365,6 @@ func (s *grpcServer) Write(srv bytestream.ByteStream_WriteServer) error {
 	resourceNameChan := make(chan string, 1)
 
 	cmp := casblob.Identity
-	var dec *zstd.Decoder
-	defer func() {
-		if dec != nil {
-			dec.Close()
-		}
-	}()
 
 	go func() {
 		firstIteration := true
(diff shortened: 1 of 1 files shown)


mostynb:
Hi, thanks for the fix- out of curiosity, how did you find this? I wonder if there's a better linter I could be using here.

I propose that we fix this in a slightly different way (see #771) which covers more cases than this PR- in particular, if the `err = dec.Reset(pr)` line returns a non-nill error in this PR then the decoder won't be returned to the pool.

tobbe76:
Thanks @mostynb I'm a golang beginner took a while to understand the sync.pool.

I used http://remote/debug/pprof/heap on the live server and found a huge total count for this object.
<br>
I can also see miljons of allocations here
https://github.com/buchgr/bazel-remote/blob/e83828a19e6860c498a6a8737bcb78cf452dbe65/cache/disk/casblob/casblob.go#L580
I have added a sync pool here also and tested it locally with good results but not yet tested on the production server.
What do you think, is it a good idea?
<br>
Why do a fsync here? It's not good for ssd performance.
https://github.com/buchgr/bazel-remote/blob/e83828a19e6860c498a6a8737bcb78cf452dbe65/cache/disk/casblob/casblob.go#L637

mostynb:
> Thanks @mostynb I'm a golang beginner took a while to understand the sync.pool.
> 
> I used http://remote/debug/pprof/heap on the live server and found a huge total count for this object. I can also see miljons of allocations here
> 
> https://github.com/buchgr/bazel-remote/blob/e83828a19e6860c498a6a8737bcb78cf452dbe65/cache/disk/casblob/casblob.go#L580
> 
> I have added a sync pool here also and tested it locally with good results but not yet tested on the production server.
> What do you think, is it a good idea?

Sounds good- do you want to open a new PR for that?

> Why do a fsync here? It's not good for ssd performance.
> https://github.com/buchgr/bazel-remote/blob/e83828a19e6860c498a6a8737bcb78cf452dbe65/cache/disk/casblob/casblob.go#L637

That is done to reduce the chance of cache corruption. I think this is most important for action cache entries which can't be verified by their hash, large/compressed CAS blobs which would be costly to verify.

mostynb:
Switched to this code in #789.

PR #771, opened 2024-08-23
Ensure that DecoderWrappers are returned to the pool after use in bytestream Writes

mostynb:
This is an alternative fix to #770.

Thanks to Thorbjorn Larsson @tobbe76 for reporting this issue.


mostynb:
Landed this on the master branch.

PR #775, opened 2024-09-02
Use syncpool for chunk buffer

tobbe76:
This reduces the number of gc:s and improves performance.
I have tested this change on our production server with good result.

### DIFF
--- a/cache/disk/casblob/casblob.go
+++ b/cache/disk/casblob/casblob.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"sync"
 
 	"github.com/buchgr/bazel-remote/v2/cache/disk/zstdimpl"
 )
@@ -508,6 +509,14 @@ func (b *readCloserWrapper) Close() error {
 	return f.Close()
 }
 
+// sync pool to reuse large byte buffer between multiple go routines
+var chunkBufferPool = &sync.Pool{
+	New: func() any {
+		b := make([]byte, defaultChunkSize)
+		return &b
+	},
+}
+
 // Read from r and write to f, using CompressionType t.
 // Return the size on disk or an error if something went wrong.
 func WriteAndClose(zstd zstdimpl.ZstdImpl, r io.Reader, f *os.File, t CompressionType, hash string, size int64) (int64, error) {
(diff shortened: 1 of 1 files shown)


cache/disk/casblob/casblob.go
```diff
@@ -508,6 +509,14 @@ func (b *readCloserWrapper) Close() error {
 	return f.Close()
 }
 
+// sync pool to reuse large byte buffer between multiple go routines
+var chunkBufferPool = &sync.Pool{
+	New: func() any {
+		b := make([]byte, defaultChunkSize)
+		return &b
```
mostynb, line 516:
```suggestion
		return make([]byte, defaultChunkSize)
```

cache/disk/casblob/casblob.go
```diff
@@ -577,7 +586,13 @@ func WriteAndClose(zstd zstdimpl.ZstdImpl, r io.Reader, f *os.File, t Compressio
...
 	var numRead int
 
-	uncompressedChunk := make([]byte, chunkSize)
+	chunkBufferPtr := chunkBufferPool.Get().(*[]byte)
+	defer func() {
+		if chunkBufferPtr != nil {
+			chunkBufferPool.Put(chunkBufferPtr)
+		}
+	}()
+	uncompressedChunk := *chunkBufferPtr
```
mostynb, line 595:
```suggestion
	chunkBuffer := chunkBufferPool.Get().([]byte)
	defer func() {
		chunkBufferPool.Put(chunkBuffer)
	}()
	uncompressedChunk := chunkBuffer
```

mostynb (commented):
I wonder if we could simplify this to just return slices, instead of pointers to slices?

tobbe76:
I have simplified it now. And the number of gc allocations is still the same in my stress test.

The reason why i used a ptr was from this thread, but to my understanding now when reading it again, it's a very small optimization using a ptr.
https://github.com/golang/go/issues/16323

mostynb:
Hmm, looks like there's a lint failure that wants us to use pointer types after all: https://staticcheck.dev/docs/checks#SA6002

Sorry- can you change that part back (and drop the third argument to make)?

mostynb:
Landed this change, though I removed the nil check in the defer statement (it seems unnecessary). Thanks again.

BTW, I notice that you're also in Göteborg according to github :) I wonder if you could share some approximate details of your bazel-remote install? I'm always looking for feedback on how people are using it.

PR #780, opened 2024-10-01
Added summary for entries' size

reimai:
Entries could be rather different in size between projects. This is why it's a summary instead of a histogram.


mostynb (approved):
Thanks for the contribution- looks good to me.

mostynb:
Squashed and landed this change on the master branch.

PR #782, opened 2024-10-10
prom: Add option to specify metrics prefix

kellyma2:
It's useful to be able to prefix the metrics that are being reported so as to more easily differentiate them from other applications that are using the same metrics.

This change adds the necessary bits (environment var, config file, command line) to support capturing the prefix and passing it through to the httpmetrics library.

### DIFF
--- a/utils/flags/flags.go
+++ b/utils/flags/flags.go
@@ -473,9 +473,9 @@ func GetCliFlags() []cli.Flag {
 		},
 		&cli.BoolFlag{
 			Name:        "http_metrics_prefix",
-			Usage:       "Prefix to apply to exported http metrics",
+			Usage:       "Whether to prefix http metrics with `bazel_remote` or not",
 			DefaultText: "false, ie no prefix",
-			EnvVars:     []string{"BAZEL_HTTP_METRICS_PREFIX"},
+			EnvVars:     []string{"BAZEL_REMOTE_HTTP_METRICS_PREFIX"},
 		},
 		&cli.BoolFlag{
 			Name:        "experimental_remote_asset_api",
--- a/utils/flags/flags.go
+++ b/utils/flags/flags.go
@@ -471,10 +471,10 @@ func GetCliFlags() []cli.Flag {
 			DefaultText: "false, ie disable metrics",
 			EnvVars:     []string{"BAZEL_REMOTE_ENABLE_ENDPOINT_METRICS"},
 		},
-		&cli.StringFlag{
+		&cli.BoolFlag{
 			Name:        "http_metrics_prefix",
 			Usage:       "Prefix to apply to exported http metrics",
-			DefaultText: "empty string, ie no prefix",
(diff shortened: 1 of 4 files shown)


kellyma2:
@mostynb can you take a look at this?

mostynb:
How disruptive would it be to just hard-code a "bazel_remote" prefix, and skip the configuration setting? And would that be sufficient, if we ignore the distruption for a moment?

kellyma2:
For folks that have dashboards and such depending on this it'd probably be fairly disruptive.  If the prefix is empty it shows up as it does today which reduces disruption.

mostynb:
OK, this sounds reasonable. I just have a couple of questions about the flag...

Is there a recommended value? Would "bazel_remote_" work for anyone who might want to use this setting? If so we could use a boolean flag and if there's ever a bazel-remove v3 we could switch to this behaviour and remove the flag.

kellyma2:
That is a possible option - I was mostly trying to give maximum flexibility.  Any particular reason for preferring that?

mostynb:
I would like to limit the number of unnecessary configuration settings, because the number of settings is growing more than I would like. If we were to remove some configuration settings in a major version upgrade in the future, then the fewer broken configurations the better, since the migration path for users would be simpler. So if there is no plausible need to specify a custom prefix there, let's not add one.

On the other hand, if you can think of a reason why we would need this configurability then that would change things a bit.

kellyma2:
Naw.  That seems totally reasonable.  My initial instinct was that may folks want to prefix this per instance, but after some thought I concluded that probably per application is fine.

utils/flags/flags.go
```diff
@@ -471,6 +471,12 @@ func GetCliFlags() []cli.Flag {
 			DefaultText: "false, ie disable metrics",
 			EnvVars:     []string{"BAZEL_REMOTE_ENABLE_ENDPOINT_METRICS"},
 		},
+		&cli.BoolFlag{
+			Name:        "http_metrics_prefix",
+			Usage:       "Prefix to apply to exported http metrics",
```
mostynb, line 476: This needs to be updated after the refactoring- how about something like "Whether or not to add a 'bazel_remote_' prefix to HTTP metrics" ?

Also, should the prefix end with an underscore?

mostynb (changes requested):
This looks pretty good, I just have a couple of small questions.

utils/flags/flags.go
kellyma2: > This needs to be updated after the refactoring- how about something like "Whether or not to add a 'bazel_remote_' prefix to HTTP metrics" ?

Good catch.

> Also, should the prefix end with an underscore?

No, the underscore between the prefix and the underlying metric gets injected automagically.

mostynb (approved):

mostynb:
Thanks- I squashed this and landed it on the master branch.

Issue #791, opened 2024-11-15
When set disable_http_ac_validation, enable_ac_key_instance_mangling will not work


hyphennn:
Hi, there. I'm using bazel-remote as our cache, my command to run progress is:
```
./bazel-bin/bazel-remote_/bazel-remote  \
  --dir /tmp/bazel-cache/ --max_size 200 --disable_http_ac_validation=true \
  --enable_endpoint_metrics=true --enable_ac_key_instance_mangling=true
```

And I find that the enable_ac_key_instance_mangling doesn't work: no matter how I change instance in .bazelrc, it always hit all caches.

I find in server/http.go Line 215:
```
	if h.mangleACKeys && kind == cache.AC {
		hash = cache.TransformActionCacheKey(hash, instance, h.accessLogger)
	}
```
So, when disable_http_ac_validation==true, the function TransformActionCacheKey will never work.

I forked this repo, and modified these code, it seems work well.

So, here is my question: Is this by design? If not, I am happy to offer my pr to fix it.

mostynb:
Hi, I'm not sure I understand the problem yet, but these two configuration settings should be independent: 
* `--enable_ac_key_instance_mangling` combines non-empty instance names with the original action cache key to form a new/different action cache key. That happens in the block you mentioned: https://github.com/buchgr/bazel-remote/blob/a563ac2ed3630e4c2539934d590d022bdbfe7321/server/http.go#L215
* `--disable_http_ac_validation=true` causes action cache entries to be returned immediately if they're found on disk. The default is to check that the entry is a valid ActionResult protobuf message and that all the CAS blobs that it references also exist in the cache, and return 404 if any of those conditions are false. This happens after the mangling (if that is enabled): https://github.com/buchgr/bazel-remote/blob/a563ac2ed3630e4c2539934d590d022bdbfe7321/server/http.go#L227

Could you provide a test case that shows the problem?

hyphennn:
Sure. I'll give an example, and I think it will be easily for you to reproduct.

Here is my command to start a bazel-remote server locally:
```
./bazel-bin/bazel-remote_/bazel-remote  \
  --dir /tmp/bazel-cache/ --max_size 200 --disable_http_ac_validation=true \
  --enable_endpoint_metrics=true --enable_ac_key_instance_mangling=true
```

Than I run bazel build with this remote cache and instance named 'test'
```
bazel build //:bazel-remote --remote_cache=http://localhost:8080/test
bazel clean
[...]

PR #792, opened 2024-11-18
fix: conflict between enable_ac_key_instance_mangling and disable_http_ac_validation

hyphennn:
Fix the problem mentioned in #791


mostynb:
Oh, I see- nice. Thanks for the fix. Landed on the master branch.

PR #795, opened 2024-12-10
Support session tokens for s3 proxy

### LINKED ISSUE #634: Support AWS session credentials
We use session credentials from the result of an `sts:AssumeRole` call (via the [`configure-aws-credentials`](https://github.com/aws-actions/configure-aws-credentials) action). These credentials take three parts rather than the two of static credentials: `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN`. `bazel-remote` does not seem to consume the `AWS_SESSION_TOKEN` environment variable, so it tries to interpret these credentials as static credentials:
```
WARNING: Remote Cache: io.grpc.StatusRuntimeException: UNKNOWN: The AWS Access Key Id you provided does not exist in our records.
```

Ideally, this common form of expressing AWS credentials would be supported by `bazel-remote`.

References:
* seems very related to #490, but that never concerned itself with `AWS_SESSION_TOKEN` - just with the profile option

There's probably a workaround for this by using a configuration file and a profile flag, but

huonw:
This adds a new arg `--s3.session_token`, to support passing through an AWS session token along with an access key, when using `--s3.auth_method=access_key`. This allows using temporary credentials (as returned by assuming a role) for short-lived invocations of bazel-remote.

Some additional background: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html

Fixes #634

### DIFF
--- a/utils/flags/flags.go
+++ b/utils/flags/flags.go
@@ -342,7 +342,7 @@ func GetCliFlags() []cli.Flag {
 			Name:    "s3.session_token",
 			Value:   "",
 			Usage:   "The S3/minio session token to use when using S3 proxy backend. Optional. " + s3AuthMsg(s3proxy.AuthMethodAccessKey),
-			EnvVars: []string{"BAZEL_REMOTE_S3_SESSION_TOKEN", "AWS_SESSION_TOKEN"},
+			EnvVars: []string{"BAZEL_REMOTE_S3_SESSION_TOKEN"},
 		},
 		&cli.StringFlag{
 			Name:        "s3.signature_type",
--- a/utils/flags/flags.go
+++ b/utils/flags/flags.go
@@ -339,6 +339,12 @@ func GetCliFlags() []cli.Flag {
 			EnvVars: []string{"BAZEL_REMOTE_S3_SECRET_ACCESS_KEY"},
 		},
 		&cli.StringFlag{
+			Name:    "s3.session_token",
+			Value:   "",
+			Usage:   "The S3/minio session token to use when using S3 proxy backend. Optional. " + s3AuthMsg(s3proxy.AuthMethodAccessKey),
+			EnvVars: []string{"BAZEL_REMOTE_S3_SESSION_TOKEN", "AWS_SESSION_TOKEN"},
+		},
+		&cli.StringFlag{
 			Name:        "s3.signature_type",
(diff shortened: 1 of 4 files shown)


huonw:
(I think the failure is unrelated? it seems the latest `master` is failing in a similar way https://github.com/buchgr/bazel-remote/commit/e36a81087c378cbd2da76e78bc2a6570bfe123ea )

utils/flags/flags.go
```diff
@@ -338,6 +338,12 @@ func GetCliFlags() []cli.Flag {
 			Usage:   "The S3/minio secret access key to use when using S3 proxy backend. " + s3AuthMsg(s3proxy.AuthMethodAccessKey),
 			EnvVars: []string{"BAZEL_REMOTE_S3_SECRET_ACCESS_KEY"},
 		},
+		&cli.StringFlag{
+			Name:    "s3.session_token",
+			Value:   "",
+			Usage:   "The S3/minio session token to use when using S3 proxy backend. Optional. " + s3AuthMsg(s3proxy.AuthMethodAccessKey),
+			EnvVars: []string{"BAZEL_REMOTE_S3_SESSION_TOKEN", "AWS_SESSION_TOKEN"},
```
mostynb, line 345: I wonder if we could skip the AWS_SESSION_TOKEN variable here- we don't make use of any of the other environment variables mentioned in https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#credentials

mostynb (commented):
Hi, thanks for the contribution. This looks OK to me, but I'm not sure about adding a new environment variable without the BAZEL_REMOTE_ prefix.

The `bazel run //:gazelle` / `bazel mod tidy` CI failure was unrelated (due to bazel 8.0.0 being released, and changes to the generated MODULE.bazel.lock file.

utils/flags/flags.go
huonw: Ah, yes, I see: I was keying off the reading of `AWS_SHARED_CREDENTIALS_FILE` and `AWS_PROFILE` below, but indeed the `--s3.access_key_id` and `--s3.secret_access_key` args support the corresponding `AWS_...` variable. The last two there are more closely related to this work, so better to follow them!

I've remove that from this PR, and it can be a separate bit of work (if at all) to read those AWS ones.

Thanks for catching 😄

mostynb:
Thanks- I landed this on the master branch.

huonw:
Thank you!

PR #807, opened 2025-03-07
Reject unsupported qualifiers in Remote Asset API

malt3:
The specification requires that any `Fetch` requests containing unsupported qualifiers are rejected with an `INVALID_ARGUMENT` rpc error[^1][^2].

Additionally, the status returned should include a `BadRequest` error detail with `FieldViolation`s indicating the names of unsupported qualifiers[^3]

[^1]: https://github.com/bazelbuild/remote-apis/blob/7f922028fcfac63bdd8431e68de152d9e7a9e2a0/build/bazel/remote/asset/v1/remote_asset.proto#L109-L112
[^2]: https://github.com/bazelbuild/remote-apis/blob/7f922028fcfac63bdd8431e68de152d9e7a9e2a0/build/bazel/remote/asset/v1/remote_asset.proto#L127-L128
[^3]: https://github.com/bazelbuild/remote-apis/blob/7f922028fcfac63bdd8431e68de152d9e7a9e2a0/build/bazel/remote/asset/v1/remote_asset.proto#L139-L143


mostynb:
Hi, thanks for the contribution.

I think the spec is too strict here- for example what if the client provides two cryptographically secure hash qualifiers, and the server only supports one of them. In that scenario I think the server should be allowed to decide if the supported qualifiers are sufficient. I created a spec change PR here: https://github.com/bazelbuild/remote-apis/pull/329

malt3:
As mentioned in [my comment](https://github.com/bazelbuild/remote-apis/pull/329#issuecomment-2709724117) on your PR, I don't think the spec allows for more than one digest right now. Additionally, bazel-remote only understands `checksum.sri` with `sha256` at the moment, so my change shouldn't have any impact on well-formed requests.

malt3:
@mostynb I can update this PR to support parsing `checksum.sri` qualifiers containing more than one checksum separated by spaces (where only sha256 is selected for now). Does that make sense to you, or are there other issues with the PR in the current state you would like me to address?

Issue #675, opened 2023-07-13
Large number of TLS handshakes to the S3 proxy


gabrielrussoc:
Hi all,

I noticed a lot of CPU usage / general slowness coming out of the Bazel Remote and after some digging I was able to pin it down to a very large number of TLS handshakes to our S3 bucket. The problem goes away if I set the `--s3.disable_ssl` flag. The metrics show a drop of FindMissingBlob requests from 4s to 400ms on a p90 level using the same physical resources.

It turns out this issue is not specific to Bazel Remote but rather to minio (the client used to talk to s3). I opened an issue there with reproduction details: https://github.com/minio/minio-go/issues/1855. Unfortunately, the issue might be even lower and actually be on the Go http library itself: https://github.com/golang/go/issues/50984.

I'm exploring whether disabling SSL is feasible for our environment, but it makes the Bazel Remote basically unusable for our volume (we're trying it with a peak of 100k requests / minute, but the real load is much higher). 

I'm using Bazel Remote v2.4.1 on kubernetes using Docker as a runtime.

chris-smith-zocdoc:
I think the handshake bug was fixed in 1.23 https://github.com/golang/go/issues/50984#issuecomment-2311420637

and the remaining issue is that the idle connection pool is too small, I've opened https://github.com/buchgr/bazel-remote/pull/808

PR #808, opened 2025-04-07
configure max idle connections to s3

chris-smith-zocdoc:
Fixes https://github.com/buchgr/bazel-remote/issues/675

This resulted in a significant cpu reduction in our environment


cache/s3proxy/s3proxy.go
```diff
@@ -68,13 +69,23 @@ func New(
 		log.Fatalf("Failed to determine s3proxy credentials")
 	}
 
+	secure := !DisableSSL
+	tr, err := minio.DefaultTransport(secure)
+	if err != nil {
+		log.Fatalf("Failed to create default transport: %v", err)
+	}
+
+	tr.MaxIdleConns = MaxIdleConns
```
chris-smith-zocdoc, line 78: The minio default for `MaxIdleConnsPerHost` is 16, which was causing excessive tls negotiation even with a modest concurrency level (~200 concurrent connections) 

https://github.com/minio/minio-go/blob/1e5fd8a9ce3588100e5be814f564fd78436aa5f4/transport.go#L51-L52

cache/s3proxy/s3proxy.go
```diff
@@ -68,13 +69,23 @@ func New(
 		log.Fatalf("Failed to determine s3proxy credentials")
 	}
 
+	secure := !DisableSSL
+	tr, err := minio.DefaultTransport(secure)
+	if err != nil {
+		log.Fatalf("Failed to create default transport: %v", err)
```
mostynb, line 75: I think this would make it easier to identify where the error is in the code:
```suggestion
		log.Fatalf("Failed to create default minio transport: %v", err)
```

utils/flags/flags.go
```diff
@@ -393,6 +393,12 @@ func GetCliFlags() []cli.Flag {
 			DefaultText: "2",
 			EnvVars:     []string{"BAZEL_REMOTE_S3_KEY_VERSION"},
 		},
+		&cli.IntFlag{
+			Name:        "s3.max_idle_conns",
+			Usage:       "The maximum number of idle connections to use when using the S3 proxy backend.",
+			DefaultText: "1024",
```
mostynb, line 399: Is this correct? The link to minio code shows 256 (and 16 per host).

cache/s3proxy/s3proxy.go
mostynb: Which values did you find the best?

cache/s3proxy/s3proxy.go
```diff
@@ -68,13 +69,23 @@ func New(
 		log.Fatalf("Failed to determine s3proxy credentials")
 	}
 
+	secure := !DisableSSL
+	tr, err := minio.DefaultTransport(secure)
```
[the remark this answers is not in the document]
mostynb: This probably needs some profiling before we consider changing the default. A separate PR would be best.

mostynb (commented):
Hi- thanks for the contribution.

cache/s3proxy/s3proxy.go
chris-smith-zocdoc: In our setup its rare for us to exceed 200 concurrent, as we tend to be very bursty, 0 -> 200 -> 0 within a very small window. I expect this to change as our baseline load increases though. 

We're using Pants and there are two settings that define the max concurrency per client remote_cache_rpc_concurrency / remote_store_rpc_concurrency at 128 each. We'll occasionally have more than one client making requests concurrently, but because they're generally very fast they often don't overlap given our current workload.

utils/flags/flags.go
chris-smith-zocdoc: Do you think I should preserve the existing defaults? In practice I think they're not good for bazel-remote, which I why I changed it here. 

`MaxIdleConnsPerHost: 16` This is the effective limit in place prior to my change, because we're only working with one host, eg `s3.us-east-1.amazonaws.com`

`MaxIdleConns: 256` This would never be used, since we can't exceed MaxIdleConnsPerHost

I can see the argument for compatibility though, and preserving `MaxIdleConnsPerHost: 16`, but that would also mean adding a second flag for `MaxIdleConns: 256` or setting that value to same as MaxIdleConnsPerHost (16)

PR #818, opened 2025-04-25
Add IRSA support on K8s

shaoxt:
To support AUTH_Method "kubernetes_identity"


shaoxt:
Looks like IRSA works by default

PR #819, opened 2025-04-25
Fix validating hash on uncompressed storage

AlessandroPatti:
When storage mode is `uncompressed`, cas blobs are not validated against their sha and accepted regardless. The validation is instead performed when using `zstd` storage mode. This refactors `casbob.WriteAndClose` to write both compressed and uncompress blobs and validate their content.

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -355,7 +355,7 @@ func (c *diskCache) writeAndCloseFile(ctx context.Context, r io.Reader, kind cac
 	if kind == cache.CAS {
 		sizeOnDisk, err = casblob.WriteAndClose(c.zstd, r, f, c.storageMode, hash, size)
 		if err != nil {
-			return -1, annotate.Err(ctx, "Failed to write compressed CAS blob to disk", err)
+			return -1, annotate.Err(ctx, "Failed to write CAS blob to disk", err)
 		}
 		closeFile = false
 		return sizeOnDisk, nil
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -352,7 +352,7 @@ func (c *diskCache) writeAndCloseFile(ctx context.Context, r io.Reader, kind cac
 	var err error
 	var sizeOnDisk int64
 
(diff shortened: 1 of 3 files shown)


cache/disk/disk.go
```diff
@@ -352,7 +352,7 @@ func (c *diskCache) writeAndCloseFile(ctx context.Context, r io.Reader, kind cac
 	var err error
 	var sizeOnDisk int64
 
-	if kind == cache.CAS && c.storageMode != casblob.Identity {
+	if kind == cache.CAS {
 		sizeOnDisk, err = casblob.WriteAndClose(c.zstd, r, f, c.storageMode, hash, size)
 		if err != nil {
 			return -1, annotate.Err(ctx, "Failed to write compressed CAS blob to disk", err)
```
mostynb, line 358:
```suggestion
			return -1, annotate.Err(ctx, "Failed to write CAS blob to disk", err)
```

cache/disk/disk.go
```diff
@@ -352,7 +352,7 @@ func (c *diskCache) writeAndCloseFile(ctx context.Context, r io.Reader, kind cac
 	var err error
 	var sizeOnDisk int64
 
-	if kind == cache.CAS && c.storageMode != casblob.Identity {
+	if kind == cache.CAS {
```
mostynb, line 355: Why isn't this change alone sufficient (and why is the casblob.go refactoring required) ? It's hard to tell in the github UI.

mostynb (commented):
Yikes- thanks for finding and reporting this!

cache/disk/disk.go
AlessandroPatti: The `WriteAndCose` method computed and wrote the CAS blob header ([here](https://github.com/buchgr/bazel-remote/blob/7f72df6ffb0913a753afa8ea3db16a54591c7493/cache/disk/casblob/casblob.go#L530-L555)) even when `Identity` was passed, but `uncompressed` expects the blobs to be plain

PR #830, opened 2025-06-13
Support --experimental_remote_downloader_propagate_credentials

michael-pplx:
Add support for forwarding credentials from client via https://bazel.build/reference/command-line-reference#common_options-flag--experimental_remote_downloader_propagate_credentials

> Whether to propagate credentials from netrc and credential helper to the remote downloader server. The server implementation needs to support the new `http_header_url:<url-index>:<header-key>` qualifier where the `<url-index>` is a 0-based position of the URL inside the FetchBlobRequest's `uris` field. The URL-specific headers should take precedence over the global headers.

### DIFF
--- a/server/grpc_asset.go
+++ b/server/grpc_asset.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"google.golang.org/genproto/googleapis/rpc/status"
@@ -56,7 +57,9 @@ func (s *grpcServer) FetchBlob(ctx context.Context, req *asset.FetchBlobRequest)
 		return nil, errNilFetchBlobRequest
 	}
 
-	headers := http.Header{}
+	globalHeader := http.Header{}
+
+	uriSpecificHeaders := make(map[int]http.Header)
 
 	for _, q := range req.GetQualifiers() {
 		if q == nil {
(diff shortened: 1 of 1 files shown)


server/grpc_asset.go
```diff
@@ -123,8 +153,18 @@ func (s *grpcServer) FetchBlob(ctx context.Context, req *asset.FetchBlobRequest)
...
-		ok, actualHash, size := s.fetchItem(ctx, uri, headers, sha256Str)
+	for uriIndex, uri := range req.GetUris() {
+		uriSpecificHeader := globalHeader
+		if header, found := uriSpecificHeaders[uriIndex]; found {
+			uriSpecificHeader = header
+			for key, value := range globalHeader {
+				if _, found := uriSpecificHeader[key]; !found {
+					uriSpecificHeader[key] = value
+				}
+			}
```
mostynb, line 164: Instead of this loop, could we 1) `uriSpecificHeader := globalHeader.Clone()`, then 2) loop over the URL-specific headers and `uriSpecificHeader.Set(key, value)` each?

mostynb (commented):
lgtm, apart from one nitpick

server/grpc_asset.go
michael-pplx: Ok done!

michael-pplx:
Thank you!

PR #837, opened 2025-07-21
Upgrade azblob module from v0.4.1 to v1.6.1

mostynb:


mostynb:
@bakjos: I wonder if you are able to test this change?

cache/azblobproxy/azblobproxy.go
```diff
@@ -69,22 +70,12 @@ func (c *azBlobCache) Get(ctx context.Context, kind cache.EntryKind, hash string
...
 
-	if err != nil {
-		cacheMisses.Inc()
-		logResponse(c.accessLogger, "DOWNLOAD", c.storageAccount, c.container, key, err)
-		return nil, -1, err
-	}
+	client := c.containerClient.NewBlockBlobClient(key)
 
-	resp, err := client.Download(context.Background(), nil)
+	resp, err := client.DownloadStream(context.Background(), nil)
```
mostynb, line 76: I think we should pass `ctx` here (and similarly in a bunch of other places).

PR #838, opened 2025-07-23
Optimize file removals

mostynb:
* Perform evictions from a single background goroutine that receives files to be removed via a channel.
* Throttle the number of concurrent Put requests and proxied Get requests (because they result in an implicit Put).
* Add a new (optional) max_size_hard_limit configuration setting, which specifies when the server should start rejecting new write requests.

A reasonable starting point for max_size_hard_limit might be ~5% larger than max_size. It can be fine-tined by watching how the prometheus query max_over_time(bazel_remote_disk_cache_size_bytes[$__interval]) varies between bazel_remote_disk_cache_size_bytes_limit{type="evict"} and bazel_remote_disk_cache_size_bytes_limit{type="reject"}.

This helps to avoid crashes due to hitting the go runtime's default limit of 10,000 operating system threads when under high load. It also makes cache eviction ~3 times faster, and gives up to 70% higher write throughput when there are lots of cache evictions.

The cache can grow above max_size when asynchronous file removals cannot keep up with new file writes.

This is a rebase + cleanup of #639 + #696.

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,6 +466,18 @@ OPTIONS:
 dir: path/to/cache-dir
 max_size: 100
 
+# If specified, write requests will be rejected when max_size_hard_limit is
+# reached. Clients can then decide which requests to retry. This setting can
+# be used to avoid running out of disk space when new blobs are uploaded faster
+# than old blobs can be evicted. A reasonable value might be 5% larger than
+# max_size. A higher limit might be needed when using a proxy backend.
+#
+# The max_size_hard_limit can be tuned by watching how the prometheus query
+# max_over_time(bazel_remote_disk_cache_size_bytes[$__interval]) varies
+# between bazel_remote_disk_cache_size_bytes_limit{type="evict"} and
+# bazel_remote_disk_cache_size_bytes_limit{type="reject"}.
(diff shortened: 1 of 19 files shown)


mostynb:
> Any chance we can get a new release cut sometime soon with this change in it? We're seeing some stability issues rn that we think this will resolve, and it'd be much easier if we could just bump our version number 😅

I was hoping to wait a bit for feedback from people with larger setups trying this out from the tip of the master branch. Though @ulrfa has been running a version of this for a couple of years already.

What kind of symptoms are you seeing, and how frequently do they occur? I might be able to make a 2.6.0 release a bit sooner if you can do some stress testing and report back relatively quickly.

nataliejameson:
I think the main symptom / thread that led me here was because we were seeing the 10k thread limit being hit once we flipped on zstd in bazel (and promptly turned it back off when we saw it also OOMing our bazel-remote instances). It's not *super* common, so I don't know that I'd be able to do much to help out on the stress testing side of things.

mostynb:
Released v2.6.0 - if you notice any problems, please create an issue here.

PR #871, opened 2025-11-14
fix: (azblob) Handle 404 responses gracefully

eplightning:
Similarly to how its already done in all other proxy backends, Azure Blob proxy backend needs to be able to handle case of missing blob in its `Get` method.

Without this, certain clients break and do not work over gRPC protocol. Example of one such a client is https://github.com/moonrepo/moon , not sure about bazel itself.


mostynb:
Landed this, with a slightly different commit message. Thanks again.

Issue #684, opened 2023-08-02
OpenTelemetry support


Gorbush:
Hi.

Is there any plan to support traces and metrics for OpenTelemetry?
We use HoneyComb for that and need to get the metrics for BazelRemote instances.

Thank you!

mostynb:
Hi.

Re metrics: bazel-remote currently supports prometheus. I'm not sure of the current status of opentelemetry metrics, is it still kind of immature?

Re tracing: I have been holding off adding this feature because I'm not sure how useful it would be. For bazel-remote installs that only use the disk cache, I suspect that disk iowait metrics are probably more useful than traces for individual requests. Maybe things are different for installs that use s3/gcs/azblob/http backends though?

smocherla-brex:
@mostynb I have an initial PR here https://github.com/buchgr/bazel-remote/pull/873, from initial testing it seems to be working as expected and is strictly opt-in behind a config.

PR #873, opened 2025-11-28
Initial attempt at OpenTelemetry support

smocherla-brex:
Addresses #684 

This PR attempts to add initial support for Opentelemetry using https://github.com/open-telemetry/opentelemetry-go:
1. Aim for support for tracing with OTEL using grpc exporter, disabled by default
2. Only traces for now to keep the scope of this PR limited. No custom spans but ones built-in that come with the SDK.
3. WIll add spans for http calls with http cache, grpc calls with grpc implementation, and for all grpc methods with server interceptors
4. Will also add spans for http calls for the proxies by wrapping minio's http transport with otlp's transport.

We don't have spans for filesystem calls - although we can add them. But I don't think it's needed for that disk/IO metrics can usually help in that situation.

Testing: Some unit tests, I'll share some details once I am able to verify end-to-end after building it here and testing with otel-collector. Below is an example of a trace I generated by running it locally.

for bystream fetch
image

for remote asset api
image


smocherla-brex:
Unfortunately haven't had time to look into this further, so I'll close this.

PR #884, opened 2026-02-13
[zstd] Build zstd library

### LINKED ISSUE #869: Managing glibc versions
👋  very happy users of bazel-remote at Quip/Slack. We run bazel-remote on a variety of Linux and macOS environments, including some that have quite old versions of glibc (2.26).

Recently the project stopped building versions that support glibc 2.26. Couple of requests - 

1. Could we encourage not replacing the binaries on existing releases? We, and hopefully others!, pin specific SHAs, and this replacement is confusing and worrying as it _looks like_ a supply-chain attack.
2. Would y'all be open to building binaries for multiple glibc versions? I would be happy to implement this if it'd be accepted.

lbcjbb:
Build the zstd library to avoid to use the pre-compiled libraries provided by the github.com/valyala/gozstd Go library.

This provides more flexibility to use any cross-compiler to build the bazel-remote binary.

Could resolve #869

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,6 +6,9 @@ bazel_dep(name = "rules_proto", version = "7.1.0")
 bazel_dep(name = "rules_pkg", version = "1.1.0")
 bazel_dep(name = "rules_oci", version = "2.2.6")
 
+# Ensure this version of the libzstd is supported by github.com/valyala/gozstd.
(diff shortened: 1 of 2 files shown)


mostynb:
Hi, I'm not sure how this would help #869 ? IIUC the glibc dependency does not come from the prebuilt gozstd libraries.

PR #886, opened 2026-02-24
grpcproxy: fix CAS writes and AC reads failing against strict gRPC backends

lance-tan-ai:
## Errors

When using the gRPC proxy backend, both CAS write-throughs and AC read-throughs fail against gRPC cache backends that strictly enforce the Remote Execution API spec.

**CAS writes** fail with `finished_writing='false'`:
```
GRPC PROXY WRITE CAS ...: rpc error: code = InvalidArgument desc = Request completed before all data was sent: digest='.../167222', actual_size='167222', finished_writing='false'
```

Then after partially fixing that, with `write_offset=0`:
```
GRPC PROXY WRITE CAS ...: rpc error: code = InvalidArgument desc = write_offset: Current write_offset (0) does not match expected one (4572)
```

**AC reads** fail with an invalid digest:
```
GRPC PROXY GET AC ...: rpc error: code = InvalidArgument desc = action_digest: invalid or missing action digest
```

## Root causes

**1. `finish_write` never set on the final ByteStream `WriteRequest`**

The ByteStream protocol requires the last `WriteRequest` to have `finish_write = true` to signal upload completion. The previous loop never set it — calling `CloseAndRecv()` alone is not sufficient for backends that require an explicit terminal message.

**2. `write_offset` not tracked across chunks**

Each `WriteRequest` must carry the cumulative byte offset of data sent so far. The previous implementation always sent `write_offset = 0`, causing backends to reject multi-chunk uploads.

**3. Action digest `size_bytes` sent as `-1` in `GetActionResult`**

`grpc_ac.go` passes `unknownActionResultSize` (-1) to `cache.Get()` because the ActionResult size is not known at request time. The proxy was forwarding this same `-1` as `size_bytes` in the `GetActionResult` Digest, which backends reject.

The correct value is `ActionDigest.SizeBytes` from the client request (the size of the Action, not the ActionResult). Since the `Proxy.Get` interface doesn't carry the action digest size, it is threaded via `context.WithValue` using a package-level typed key.

## Fix

- Track `finishWrite` from `io.EOF` and set `FinishWrite: true` on the corresponding `WriteRequest`, handling both EOF-with-data (`n > 0`) and EOF-only (`n == 0`).
- Track a running `writeOffset` and include it in every `WriteRequest`.
- In `grpc_ac.go`, store `req.ActionDigest.SizeBytes` in context before calling `cache.Get()`. In the proxy's AC path, read it back to construct a valid `Digest` for `GetActionResult`.

## Validation

After deploying with these fixes and clearing the local disk cache to force proxy fallbacks, all CAS writes and AC reads succeed:
```
GRPC PROXY GET AC 4dfa76214a85a477dc124e75adfe9415d07ce06d6abedcf77513f1dc2b395738: Success
GRPC PROXY CONTAINS CAS ce443496d644a99cc58a1da41a40aaf18d356abec228f4e1864ff72702887105: Success
GRPC PROXY GET AC 7a04428fcdb0388652bc0193c22a9fa5249f54170dca5cb76622c101ee0c6e00: Success
GRPC PROXY GET AC fda4ef0d282774ed830a2492be005785fd7c5e68011cd523c45a9c458bb1d011: Success
GRPC PROXY GET AC 5f82c3f1c16e0bb4a00ad91c9018198d55243af7c4b55e1882759ee53e3fc248: Success
GRPC PROXY GET AC ae8217d7f508d073256d729f7b0b1b14affa20576c2c180904f1679bdd265c42: Success
GRPC PROXY GET AC 31ae1f5e45defefd1ff2338a4ebd96d34823c4030d50ec99f2b959ea78ebf65a: Success
GRPC PROXY GET AC ea8e8bc151329e536b073d202b2af4c701df28b6ce484e0da6ee10cbd6337837: Success
GRPC PROXY GET AC 5aab10c5e7a8f372b250bcc6b69403e85976e257f364d283577d552684a1b41e: Success
```

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -7,6 +7,14 @@ import (
 	"io"
 )
 
+type contextKey int
+
+const (
+	// ActionDigestSizeBytesKey carries the action digest size_bytes to the proxy,
+	// which needs it to issue a valid GetActionResult request.
+	ActionDigestSizeBytesKey contextKey = iota
+)
+
 // EntryKind describes the kind of cache entry
 type EntryKind int
 
--- a/cache/grpcproxy/grpcproxy.go
+++ b/cache/grpcproxy/grpcproxy.go
@@ -297,9 +297,15 @@ func (r *remoteGrpcProxyCache) Get(ctx context.Context, kind cache.EntryKind, ha
 		// is enabled. We can treat them as AC in this scope
 		fallthrough
 	case cache.AC:
+		actionDigestSize := int64(-1)
+		if v := ctx.Value(cache.ActionDigestSizeBytesKey); v != nil {
+			if sz, ok := v.(int64); ok {
+				actionDigestSize = sz
+			}
+		}
 		digest := pb.Digest{
 			Hash:      hash,
-			SizeBytes: size,
+			SizeBytes: actionDigestSize,
 		}
 
 		req := &pb.GetActionResultRequest{ActionDigest: &digest}
--- a/cache/grpcproxy/grpcproxy.go
+++ b/cache/grpcproxy/grpcproxy.go
@@ -201,6 +201,25 @@ func (r *remoteGrpcProxyCache) UploadFile(item backendproxy.UploadReq) {
 			}
 
 			if finishWrite {
+				if n == 0 {
+					// All data was sent in previous iterations without FinishWrite.
+					// io.Reader may return (n>0, nil) for the last chunk then
+					// (0, io.EOF) on the next call. Send a zero-data terminal
+					// message so the server sees finish_write=true.
+					rn := ""
+					if firstIteration {
+						firstIteration = false
+						rn = resourceName
+					}
+					req := &bs.WriteRequest{
+						ResourceName: rn,
+						FinishWrite:  true,
+					}
+					if err := stream.Send(req); err != nil {
+						logResponse(r.errorLogger, "Write", err.Error(), item.Kind, item.Hash)
+						return
+					}
+				}
 				_, err := stream.CloseAndRecv()
 				if err != nil {
 					logResponse(r.errorLogger, "Write", err.Error(), item.Kind, item.Hash)
--- a/cache/grpcproxy/grpcproxy.go
+++ b/cache/grpcproxy/grpcproxy.go
@@ -166,6 +166,7 @@ func (r *remoteGrpcProxyCache) UploadFile(item backendproxy.UploadReq) {
 		resourceName := fmt.Sprintf(template, uuid.New().String(), item.Hash, item.LogicalSize)
 
 		firstIteration := true
+		writeOffset := int64(0)
 		for {
 			n, readErr := item.Rc.Read(buf)
 			if readErr != nil && readErr != io.EOF {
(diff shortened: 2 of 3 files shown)


lance-tan-ai:
@buchgr Mind taking a look at this? We are hitting this issue when trying to write CAS blobs through bazel-remote's gRPC proxy.

fwiw, our setup is the basic bazel-remote setup and just trying to hit the gRPC proxy backend. We are seeing this error appear on every proxy write.

mostynb:
Hi, which server are you seeing these errors with?

lance-tan-ai:
> Hi, which server are you seeing these errors with?

@mostynb We are seeing this with EngFlow servers.

lance-tan-ai:
For some context, this is our `bazel-remote-compose.yaml`:
```
services:
  bazel-remote:
    build:
      context: .
      dockerfile: Dockerfile
    restart: unless-stopped
    logging:
      driver: journald
    environment:
      BAZEL_REMOTE_DIR: /tmp/bazel-remote/
      BAZEL_REMOTE_MAX_SIZE: 5000
      BAZEL_REMOTE_PORT: 8080
      BAZEL_REMOTE_STORAGE_MODE: uncompressed
      BAZEL_REMOTE_ENABLE_ENDPOINT_METRICS: true
      BAZEL_REMOTE_MAX_BLOB_SIZE: 524288000
      BAZEL_REMOTE_MAX_PROXY_BLOB_SIZE: 524288000
      BAZEL_REMOTE_GRPC_PROXY_URL: grpcs://<name>.cluster.engflow.com
      BAZEL_REMOTE_GRPC_PROXY_CERT_FILE: /etc/bazel-remote/engflow.crt
      BAZEL_REMOTE_GRPC_PROXY_KEY_FILE: /etc/bazel-remote/engflow.key
```

mostynb:
I refreshed my memory a bit, bazel-remote's grpc proxy backend is a relatively recent addtion, and should probably still be marked as experimental. It has the best chance of working with another bazel-remote instance as the grpc proxy backend.

The reason for this is historical- bazel-remote originally only supported bazel's older http cache protocol, which does not provide full digest information to the server as the newer grpc protocol does. Some non-trivial refactoring is required to improve the situation, and I suspect that your changes to GetActionResult would not be enough to make the current setup work well. I would prefer to take some time to fix this instead of sneaking extra data through with the context.

However the grpcproxy bytestream changes look OK. Would you be interested in reducing the scope of this PR to focus on this?

lance-tan-ai:
> I refreshed my memory a bit, bazel-remote's grpc proxy backend is a relatively recent addtion, and should probably still be marked as experimental. It has the best chance of working with another bazel-remote instance as the grpc proxy backend.
> 
> The reason for this is historical- bazel-remote originally only supported bazel's older http cache protocol, which does not provide full digest information to the server as the newer grpc protocol does. Some non-trivial refactoring is required to improve the situation, and I suspect that your changes to GetActionResult would not be enough to make the current setup work well. I would prefer to take some time to fix this instead of sneaking extra data through with the context.
> 
> However the grpcproxy bytestream changes look OK. Would you be interested in reducing the scope of this PR to focus on this?

Will descope the PR. I will create an issue to outline all of the issues I was seeing with migrating to a gRPC backend since there were additional patches that I needed to apply for it to work out-of-the-box.

PR #888, opened 2026-04-23
Allow systemd dynamic user

name-snrl:
This hack was originally added in 48feee2. Remove it in favor of using
DynamicUser= in the systemd service.

The problem here, as documented in systemd.exec(5), is that DynamicUser=
implicitly enables RestrictSUIDSGID= and cannot be disabled, which
breaks bazel-remote.

cc @mostynb

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -285,12 +285,6 @@ func (c *diskCache) Put(ctx context.Context, kind cache.EntryKind, hash string, 
 			if err != nil {
 				log.Printf("warning: failed to remove temp file: %q", blobFile)
 			}
-		} else if blobFile != "" {
-			// Mark the file as "complete".
-			err := os.Chmod(blobFile, tempfile.FinalMode)
-			if err != nil {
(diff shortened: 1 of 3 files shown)


mostynb:
The setgid hack was added to prevent incomplete blobs being used after restarting bazel-remote. This PR removes that safety check but doesn't add a replacement, which doesn't sound great.

What is the use case for DynamicUser with bazel-remote?

name-snrl:
> The setgid hack was added to prevent incomplete blobs being used after restarting bazel-remote


I looked at initial commit where it was added and understand why this was done. But as far as I can see, the a GUID check was removed in 15f05f8.

name-snrl:
> What is the use case for DynamicUser with bazel-remote?

It simplifies configuration. All you have to do with DynamicUser=true is enable and start the service. It also apples basic sandboxing

mostynb:
Ah, I see. Since this missing check wasn't noticed for such a long time, it's unlikely to be important, and we may as well drop the unnecessary chmod's.

I landed these changes with some minor changes to the comments and git commit message.

Issue #887, opened 2026-04-14
Add .netrc support for remote asset downloads


vcomito-apexai:
The bazel-remote server can fetch assets:

https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/asset/v1/remote_asset.proto#L90

```
  // Servers *MAY* fetch content that they do not already have cached, for any
  // URLs they support
```

Some assets might be only accessible via username:password (or Personal Access Token) credentials. These are usually stored inside
the .netrc file of the user. bazel-remote can use this file for accessing the protected artifacts

vcomito-apexai:
@buchgr I will open soon a PR for this

PR #892, opened 2026-05-04
[887] add support for netrc in remote asset downloads 

vcomito-apexai:
Addresses https://github.com/buchgr/bazel-remote/issues/887 

with these changes, the bazel-remote server will use the credentials stored in the .netrc file for accessing artifacts that require username:password 

Changes:

- When `--experimental_remote_asset_api` is enabled, bazel-remote now loads credentials from NETRC env variable (if specified) or the process user’s ~/.netrc
- When an HTTP request is made to a domain listed in the .netrc file, the credentials are used as HTTP Basic Auth for asset fetches.
- .netrc files are parsed like curl would do


vcomito-apexai:
I guess the `gazelle & bazel mod tidy on` should be run separately?

Issue #893, opened 2026-05-05
Ghost LRU entries persist indefinitely after file deletion — file-not-found path doesn't evict from index


im-nick-adams:
## Summary

When a CAS blob is deleted from disk (e.g., via the overwrite path in `Add()`), but a concurrent `GET` has already looked up the key in the LRU and released the lock, the subsequent `os.Open` fails with "not exists." The warning is logged, but the stale LRU entry is **never removed**, causing the same warning to fire on every subsequent access until the server is restarted.

## Impact

On a moderately active server (~10 concurrent Bazel users), we observed 20,145 phantom warnings over 2 days (~49/minute), producing persistent `Lost inputs no longer available remotely` failures for Bazel clients using `--remote_download_toplevel`. A restart temporarily resolves the issue by rebuilding the LRU from disk, but phantoms accumulate again within hours.

## Root Cause

In `cache/disk/disk.go`, `availableOrTryProxy()`:

```go
// Slow path retry:
c.mu.Lock()
item, listElem = c.lru.Get(key)
if listElem != nil {
    blobPath = path.Join(c.dir, c.FileLocation(...))
    f, err = os.Open(blobPath)
}
c.mu.Unlock()

if err != nil {
    // WARNING: logs but does NOT evict from LRU
    log.Printf("Warning: expected %q to exist on disk, undersized cache?", blobPath)
}
```

Compare to the decompression-error path ~20 lines later, which correctly self-heals:

```go
if err != nil {
    log.Printf("Warning: expected item to be on disk, but something happened...")
    c.mu.Lock()
    c.lru.RemoveElement(listElem)  // ← correctly removes dead entry
    c.mu.Unlock()
}
```

## Suggested Fix

Add LRU eviction in the file-not-found path, matching the existing decompression-error pattern:

```go
if err != nil {
    log.Printf("Warning: expected %q to exist on disk, undersized cache?", blobPath)
    c.mu.Lock()
    if listElem != nil {
        c.lru.RemoveElement(listElem)
    }
    c.mu.Unlock()
}
```

This makes ghost entries self-healing on first access rather than persisting indefinitely.

## Environment

- bazel-remote commit: `b857daf1f63c641dc3fe6105a674a6e9ed81cf35`
- Go 1.25.6
- Local disk storage (no NFS), zstd compression mode
- ~847K cached files, 126 GB on disk, 1.2 TB max

## Reproduction

1. Start bazel-remote with multiple concurrent clients
[...]

PR #894, opened 2026-05-05
If a cache entry can't be found on disk, remove the cache entry

mostynb:
Initial mitigation for #893.

### DIFF
--- a/cache/disk/disk.go
+++ b/cache/disk/disk.go
@@ -462,23 +462,29 @@ func (c *diskCache) availableOrTryProxy(kind cache.EntryKind, hash string, size 
 
 		if !isSizeMismatch(size, item.size) {
 			var f *os.File
+			fastPath := true
 			f, err = os.Open(blobPath)
 			if err != nil && os.IsNotExist(err) {
 				// Another request replaced the file before we could open it?
 				// Enter slow path.
+				fastPath = false
 
 				c.mu.Lock()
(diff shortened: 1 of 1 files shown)


ulrfa (commented):
Good! I agree it make sense to delete phantom entries to minimize the impact of issues.

However, see suggestion regarding the logging.

ulrfa:
_The github.com web UI gives weird error messages for my review comment. I'm trying to write it here instead:_

If os.Open() in the **fast** path (line 465) is failing with IsNotExist and lru.Get() on line 471 is not finding the key, then I assume it is caused by an undersized cache and the operator needs to address the problem. The existing warning is reasonable for that scenario:

`"Warning: expected %q to exist on disk, undersized cache?"`

But I expect the os.Open() in the **slow** path (line 474) to never fail, and if it does it could indicate either:

- Bug in bazel-remote.
- Something external manipulating the disk storage.
- Several bazel-remote instances incorrectly configured to share same disk storage.
- Other unknown disk storage issue.

Therefore I suggest having separate log messages so that we can distinguish between if failing in fast or slow path. Perhaps also including err in the log message in case of failure in slow path?

mostynb:
> Good! I agree it make sense to delete phantom entries to minimize the impact of issues.
> 
> However, see suggestion regarding the logging.

Updated the log to show if we took the slow path, and the last reported error.

PR #904, opened 2026-06-16
Null-out Android env for non-Android macos builds

ted-xie:
The Intel `macos` deployment in BazelCI does not have the latest Android build tools installed. By default, BazelCI adds `ANDROID*` environment variables to builds, which registers an Android toolchain for any transitively-loaded Android deps, which causes a repo rule-level error.

To prevent this, this PR nulls out the `ANDROID*` environment variables.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -10,6 +10,9 @@ tasks:
       - "..."
   macos:
     platform: macos
+    environment:
+      ANDROID_HOME: ""
+      ANDROID_NDK_HOME: ""
     build_targets:
       # Skip the (linux) container image targets, just build the binary.
       - "//:bazel-remote"


ted-xie:
NOTE: An alternative is to change the BazelCI platform from `macos` to `macos_arm64`.

mostynb:
I can't see any BazelCI errors in the auto-triggered CI jobs on the tip of the master branch.

But switching from macos to macos_arm64 sounds like something we should do anyway.

mostynb:
@ted-xie: I switched to macos_arm64 instead in https://github.com/buchgr/bazel-remote/pull/905.

ted-xie:
Thanks! To answer your previous question, we have an upcoming change in Bazel that would eventually cause `macos` build errors: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/5546#019ecd05-779d-4897-a07e-ea1c23d102c8

PR #475, opened 2021-09-22
[zstd] test low memory options

mostynb:
Let's see how this affects memory usage and performance.

Related to #473.

PR #907, opened 2026-07-13
casblob: reuse the compressed-output buffer to cut upload memory churn

mjonss:
## Problem

Under a burst of concurrent uploads, bazel-remote can suffer a large transient heap overshoot (and OOM on memory-constrained hosts) despite small steady-state memory. `alloc_space` heap profiling attributes the bulk to `zstd.(*Encoder).encodeAll`, reached via `casblob.WriteAndClose`.

## Root cause

The zstd write path compresses each blob in 1 MiB chunks and called `EncodeAll(in, nil)` per chunk, allocating a fresh output slice for every chunk of every upload. The chunk size is exactly `1<<20`, and klauspost's `EncodeAll` only pre-allocates an output buffer when `len(src) < 1<<20` (strictly less-than). A full 1 MiB chunk therefore starts from a `nil` dst and grows it by repeated `append`-doubling — several MiB of transient garbage per chunk.

## Fix

Thread a reusable `dst` through `zstdimpl.EncodeAll` (both the pure-Go and cgo backends already accept one) and have `casblob.WriteAndClose` allocate a single output buffer, sized to `compressBound(chunkSize)`, reused across all chunks of the blob. Sizing to the `ZSTD_compressBound` worst case guarantees `EncodeAll` never has to grow (and reallocate) the buffer, even for incompressible chunks whose compressed output is slightly larger than the input.

This complements #775 (which pooled the *input* chunk buffer) by removing the per-chunk *output* allocation, and is complementary to #475 (which tunes the encoder's internal memory rather than the output buffer).

## Numbers

`BenchmarkWriteAndCloseZstd` (16 MiB blob, incompressible data), added in this PR:

|        |     B/op | allocs/op |
|--------|---------:|----------:|
| before | ~82.8 MB |       215 |
| after  |  ~1.1 MB |        20 |

All existing tests pass.

### DIFF
--- a/cache/disk/casblob/casblob.go
+++ b/cache/disk/casblob/casblob.go
@@ -396,7 +396,8 @@ func GetZstdReadCloser(zstd zstdimpl.ZstdImpl, f *os.File, expectedSize int64, o
 	}
 
 	chunkToRecompress := uncompressedFirstChunk[remainder:]
-	recompressedChunk := zstd.EncodeAll(chunkToRecompress)
+	dst := make([]byte, 0, compressBound(len(chunkToRecompress)))
+	recompressedChunk := zstd.EncodeAll(dst, chunkToRecompress)
 
 	br := bytes.NewReader(recompressedChunk)
 	if chunkNum == int64(len(h.chunkOffsets)-2) {
@@ -517,6 +518,27 @@ var chunkBufferPool = &sync.Pool{
 	},
 }
 
+// compressBound returns the guaranteed maximum compressed size of srcSize
+// bytes. It is a Go port of the ZSTD_COMPRESSBOUND() C macro from libzstd:
+//
+//	#define ZSTD_COMPRESSBOUND(srcSize) \
+//	    ((srcSize) + ((srcSize)>>8) + \
+//	     (((srcSize) < (128<<10)) ? (((128<<10) - (srcSize)) >> 11) : 0))
+//
+// zstd's own C.ZSTD_compressBound() is used by the cgo implementation; this
+// keeps the pure-Go path from depending on any single backend's internals.
+// We use it to size the reusable compressed-output buffer so that EncodeAll
+// never has to grow (and thus reallocate) it, even for incompressible chunks
+// whose output is slightly larger than the input.
+func compressBound(srcSize int) int {
+	const lowLimit = 128 << 10 // 128 KiB
+	margin := 0
+	if srcSize < lowLimit {
+		margin = (lowLimit - srcSize) >> 11
+	}
+	return srcSize + (srcSize >> 8) + margin
+}
+
 // Read from r and write to f, using CompressionType t.
 // Return the size on disk or an error if something went wrong.
 func WriteAndClose(zstd zstdimpl.ZstdImpl, r io.Reader, f *os.File, t CompressionType, hash string, size int64) (int64, error) {
@@ -592,6 +614,11 @@ func WriteAndClose(zstd zstdimpl.ZstdImpl, r io.Reader, f *os.File, t Compressio
 	}()
 	uncompressedChunk := *chunkBufferPtr
 
+	// Reusable output buffer for the compressed chunks. Sized to the maximum
+	// compressed size of a chunk so EncodeAll writes into it without growing,
+	// and reused for every chunk of this blob.
+	compressedChunkBuffer := make([]byte, 0, compressBound(int(chunkSize)))
+
 	hasher := sha256.New()
 
 	for nextChunk < len(h.chunkOffsets)-1 {
@@ -609,7 +636,7 @@ func WriteAndClose(zstd zstdimpl.ZstdImpl, r io.Reader, f *os.File, t Compressio
 			return -1, fmt.Errorf("only managed to read %d of %d bytes: %w", numRead, chunkEnd, err)
 		}
 
-		compressedChunk := zstd.EncodeAll(uncompressedChunk[0:chunkEnd])
+		compressedChunk := zstd.EncodeAll(compressedChunkBuffer[:0], uncompressedChunk[0:chunkEnd])
 
 		hasher.Write(uncompressedChunk[0:chunkEnd])
 
--- a/cache/disk/casblob/casblob.go
+++ b/cache/disk/casblob/casblob.go
@@ -22,6 +22,8 @@ const (
 	Zstandard CompressionType = 1
 )
 
+// If changed to < 128 KiB, WriteAndClose's output-buffer sizing must be updated
+// (see the compressedChunkBuffer comment there).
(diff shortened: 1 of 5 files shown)


mjonss:
I found this while setting up bazel-remote on a small node with limited memory, and it had OOM issues, so after investigating I came up with this PR.

mjonss:
I decided to make a smaller change, instead of using a sync.Pool since the benefit was very good to start with for my use case. If you find it better to also include a sync.Pool also for the `compressedChunkBuffer` as it already is for `uncompressedChunk`, I'm happy to add that too.

cache/disk/casblob/casblob_test.go
```diff
@@ -82,3 +82,76 @@ func TestZstdFromLegacy(t *testing.T) {
 		t.Fatalf("Unexpected content sha %s, expected %s", hs, hash)
 	}
 }
+
+// blobSizeForBenchmark is deliberately several times the 1 MiB chunk size, so
+// that WriteAndClose compresses the blob in multiple chunks. Each chunk in the
+// zstd write path currently allocates a fresh output buffer
+// (zstd.EncodeAll(in, nil)), so the transient garbage produced per Put scales
+// with the blob size. This is the allocation churn that drives the OOM under
+// concurrent upload bursts (see bazel-remote-oom-findings.md).
```
mostynb: bazel-remote-oom-findings.md doesn't exist in the PR, maybe add the PR's URL instead?

cache/disk/casblob/casblob.go
```diff
@@ -592,6 +614,11 @@ func WriteAndClose(zstd zstdimpl.ZstdImpl, r io.Reader, f *os.File, t Compressio
 	}()
 	uncompressedChunk := *chunkBufferPtr
 
+	// Reusable output buffer for the compressed chunks. Sized to the maximum
+	// compressed size of a chunk so EncodeAll writes into it without growing,
+	// and reused for every chunk of this blob.
+	compressedChunkBuffer := make([]byte, 0, compressBound(int(chunkSize)))
```
mostynb: chunkSize is based on a constant, so we could compute compressBound(int(chunkSize)) once, store it in a package variable and use it here

cache/disk/casblob/casblob.go
```diff
@@ -517,6 +518,27 @@ var chunkBufferPool = &sync.Pool{
...
+//	#define ZSTD_COMPRESSBOUND(srcSize) \
+//	    ((srcSize) + ((srcSize)>>8) + \
+//	     (((srcSize) < (128<<10)) ? (((128<<10) - (srcSize)) >> 11) : 0))
+//
+// zstd's own C.ZSTD_compressBound() is used by the cgo implementation; this
+// keeps the pure-Go path from depending on any single backend's internals.
+// We use it to size the reusable compressed-output buffer so that EncodeAll
+// never has to grow (and thus reallocate) it, even for incompressible chunks
+// whose output is slightly larger than the input.
+func compressBound(srcSize int) int {
```
mostynb: Is this guaranteed to work for github.com/klauspost/compress/zstd ?

cache/disk/zstdimpl/gozstd.go
```diff
@@ -65,6 +65,6 @@ func (goZstd) DecodeAll(in []byte) ([]byte, error) {
 	return decoder.DecodeAll(in, nil)
 }
 
-func (goZstd) EncodeAll(in []byte) []byte {
-	return encoder.EncodeAll(in, nil)
+func (goZstd) EncodeAll(dst, in []byte) []byte {
+	return encoder.EncodeAll(in, dst)
```
mostynb: For consistency let's use the argument names src and dst, and in the same order that github.com/klauspost/compress/zstd uses- "src, dst"

cache/disk/zstdimpl/cgozstd.go
```diff
@@ -35,8 +35,8 @@ func (cgoZstd) DecodeAll(in []byte) ([]byte, error) {
 	return gozstd.Decompress(nil, in)
 }
 
-func (cgoZstd) EncodeAll(in []byte) []byte {
-	return gozstd.CompressLevel(nil, in, compressionLevel)
+func (cgoZstd) EncodeAll(dst, in []byte) []byte {
+	return gozstd.CompressLevel(dst, in, compressionLevel)
```
mostynb: Same comment as in gozstd.go ...

cache/disk/zstdimpl/zstdimpl.go
```diff
@@ -39,7 +39,13 @@ type ZstdImpl interface {
...
 	GetEncoder(out io.WriteCloser) (zstdEncoder, error)
 	DecodeAll(in []byte) ([]byte, error)
-	EncodeAll(in []byte) []byte
+
+	// EncodeAll compresses in and appends the result to dst, returning the
+	// updated slice (like the underlying zstd EncodeAll). Passing a dst with
+	// enough spare capacity lets callers reuse a single output buffer across
+	// many chunks instead of allocating a fresh one per call. Pass a nil dst
+	// to let the implementation allocate.
+	EncodeAll(dst, in []byte) []byte
```
mostynb: The argument names/order needs to be changed here too.

mostynb (commented):
Thanks- this looks pretty good, I'm a little worried that the go zstd library might not provide a guarantee on the upper limit of the size of the compressed output. Maybe we should ask Klaus?

cache/disk/casblob/casblob_test.go
mjonss: Oops, that is from my own non-included investigation files, I'll add a PR url instead (pointing to this PR).

cache/disk/casblob/casblob.go
mjonss: Thanks, I will update it to use a new constant, like `maxCompressedChunkSize`, as well as remove `compressBound()` function.

cache/disk/casblob/casblob.go
mjonss: It should be, since it uses the same calculation in a macro. But I will double check.

mjonss: klauspost has even smaller overhead, as well as if the buffer is too small, then it will just extend/re-alloc, so falling back to previous behavior.

cache/disk/casblob/casblob_test.go
mjonss: Done.

cache/disk/casblob/casblob.go
mjonss: I just added the calculation directly in the assignment, with a more elaborate comment.

cache/disk/zstdimpl/gozstd.go
mjonss: Done

mostynb:
I landed a squashed and rebased version of this on the master branch.

Issue #529, opened 2022-03-08
Very high memory usage on v2.3.3 - is this configurable?


SrodriguezO:
We're experiencing severe memory issues w/ the cache since upgrading to v2.3.3 (from v1.1.0). These were asymptomatic for most of January and February, but started causing frequent cache OOMs following our upgrade to Bazel 5 at the beginning of last week. The memory footprint was already significantly higher prior to the Bazel 5 upgrade, however.

Prior to the bazel-remote cache upgrade (which took place 01/01/2022), memory usage was minimal. Following the upgrade, the cache process regularly uses up all the memory on the host (~92g), resulting in the OOM killer killing the cache. 

![remote_usable_memory](https://user-images.githubusercontent.com/44655593/157154059-eb479943-39e3-4a96-959f-3b6e1ecb0819.png)

We noticed that the used file handles count markedly dropped following the cache upgrade as well, which leads us to believe some actions that previously relied on heavy disk usage now occur in-memory. 

![remote_file_handles](https://user-images.githubusercontent.com/44655593/157154077-2cffa98a-963d-4f94-807c-93fb3b8d1e92.png)

--

A very large chunk of the memory usage occurs during cache startup. For example, following a crash at 2:10pm, the cache was holding 70g of memory by 2:29pm, which is when the cache finally started serving requests. You can see the memory usage trend for that OOM/restart (and two prior ones) on this screenshot:

![remote_mem_used_2](https://user-images.githubusercontent.com/44655593/157154103-809fad5e-22db-43d0-8a8f-e3c6c26d1656.png)

The cache logs show
```
<~21:10:00 process starts - logs are truncated, so the exact timestamp is missing, but our service wrapper simply launches this docker container>
…
… <tons of "Removing incomplete file" logs>
2022/03/07 21:24:28 Removing incomplete file: /bazel-remote/cas.v2/ff/f…
2022/03/07 21:24:28 Removing incomplete file: /bazel-remote/cas.v2/ff/f…
2022/03/07 21:24:29 Removing incomplete file: /bazel-remote/cas.v2/ff/f…
2022/03/07 21:24:29 Sorting cache files by atime.
2022/03/07 21:26:26 Building LRU index.
2022/03/07 21:29:41 Finished loading disk cache files.
2022/03/07 21:29:41 Loaded 54823473 existing disk cache items.
2022/03/07 21:29:41 Mangling non-empty instance names with AC keys: disabled
[...]

PR #908, opened 2026-07-13
server: add --max_inflight_bytes to bound concurrent request memory

mjonss:
## Problem

Under a burst of concurrent uploads or downloads, bazel-remote's peak memory scales with client fan-out and can OOM (see #529): each in-flight gRPC bytestream request pins live memory — a per-stream read buffer, a zstd encoder/decoder window, and gRPC send/receive buffers — and there is no limit on how many run concurrently.

A heap profile captured at the peak of a real download burst showed **~5.8 GB resident across ~1000 concurrent `Read` streams** — 48% gRPC send buffers, 29% zstd decoder windows, 20% read buffers — essentially all under `grpcServer.Read`. The working set fully recovers afterward, so it is live memory proportional to concurrency, not a leak.

## Fix

Add a `--max_inflight_bytes` flag (env `BAZEL_REMOTE_MAX_INFLIGHT_BYTES`) that bounds the total logical (uncompressed) size of blobs being served or received concurrently, via a weighted semaphore held for the whole request. Once the budget is reached, further requests block (backpressure) instead of exhausting memory. The weight is clamped to `[1, budget]` so a blob larger than the whole budget is still served (serialized against others) rather than deadlocking. `0` (the default) disables the limit, preserving existing behavior. Client cancellation releases the wait.

This applies the same bounded-concurrency idea that #480 / #556 introduced for disk eviction to request memory, and makes the high memory usage reported in #529 configurable.

## Tests

`TestGrpcInflightLimitBoundsConcurrentReads` verifies the cap deterministically: with a budget of K×blobSize, at most K reads reach the cache concurrently and the rest block until budget frees (verified to fail when the limit is disabled).

This PR also adds a read/decompress benchmark documenting that the read path barely churns (the zstd decoder is pooled) — i.e. the download OOM is driven by *live* per-stream state × concurrency, not allocation churn, which is why the fix is a concurrency bound rather than buffer pooling.

### DIFF
--- a/cache/disk/disk_test.go
+++ b/cache/disk/disk_test.go
@@ -1854,3 +1854,87 @@ func TestResultFromProxyTooLargeToReserve(t *testing.T) {
 
 	close(fakeProxy.getEvents)
 }
+
+// --- Read/decompress hot-path benchmarks ---
+//
+// CAS blobs are stored zstd-compressed (the default storageMode), so serving an
+// uncompressed Get decompresses on the fly. Unlike the write path (which
+// allocated a fresh compressed-output buffer per 1 MiB chunk), the read path
+// gets its zstd decoder from a sync.Pool and reuses it, so steady-state
+// per-request churn is expected to be modest. The memory that drives OOM under
+// download bursts is the *live* per-stream state (pooled decoder window buffers
+// plus gRPC send buffers) multiplied by unbounded concurrency, not per-request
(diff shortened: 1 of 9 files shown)


mjonss:
This is to make it easier to run bazel-remote with gRPC in low memory setups, which was an issue for me. The argument and env variable name was chosen to be more related to memory usage, instead of directly to concurrency (which then needs to be calculated from how much memory would be used per stream).

PR #910, opened 2026-07-22
Land #907

mostynb:
This is a squashed and rebased version of #907, so I can trigger CI on it in that state before landing.


mostynb:
This has landed.

PR #924, opened 2026-09-07
disk cache: cache positive proxy existence checks

### LINKED ISSUE #919: Proxy `Contains()` results are never cached, so the same digests are re-checked on every build
## Summary

When a blob is present in a proxy backend but not on local disk, every `FindMissingBlobs`
that references it issues a fresh `HEAD` to the backend. The result isn't cached, and a
positive result doesn't populate the disk cache — so nothing ever breaks the cycle and the
same digest is re-checked indefinitely, once per build, forever.

On our deployment this makes existence checks **93.8% of all backend requests**, at roughly
**48 `HEAD`s per actual `GET`**.

Is this intended? I couldn't find an existing issue and wanted to check before proposing
anything. #442 optimised the concurrency of this path in 2021, but not the request volume.

## Environment

- bazel-remote v2.6.1 (behaviour verified unchanged in v2.6.2)
- 8 replicas, ~600 GiB disk cache each (~4.8 TiB total), `--storage_mode=zstd`
- GCS proxy backend via `gcs_proxy`, ~40M objects / ~73 TiB
- Bazel CI, requests load-balanced per-blob by consistent hash on the request path

## Measurements

Backend request counts over 7 days, from the storage provider's own metrics:

| Operation | Requests | Share | Sustained |
|---|---:|---:|---:|
| `HEAD` (existence check) | 231,274,027 | **93.8%** | 382/s |
| `PUT` (upload) | 10,502,671 | 4.3% | 17.4/s |
| `GET` (download) | 4,812,766 | 2.0% | 8.0/s |

The repetition is arithmetically provable without further instrumentation:

- ~10.5M of those `HEAD`s returned 404 (each led to a `PUT`), leaving **~220.8M that returned 200**
- A `HEAD` can only return 200 for an object that exists, and the bucket holds **~39.8M objects**
- Therefore **at most 39.8M were distinct → a floor of ~5.5 `HEAD`s per distinct object per week**

That's a lower bound. The realistic figure is far higher, since checks concentrate on the
active working set rather than spreading over the whole bucket.

## Where it happens

`cache/disk/disk.go`, `diskCache.Contains()` (v2.6.2 line 757):

```go
item, listElem := c.lru.Get(key)
exists := listElem != nil
...
if exists && !isSizeMismatch(size, foundSize) {
    return true, foundSize
}

if c.proxy != nil && size <= c.maxProxyBlobSize {
    exists, foundSize = c.proxy.Contains(ctx, kind, hash, size)
    if exists && foundSize <= c.maxProxyBlobSize && !isSizeMismatch(size, foundSize) {
        return true, foundSize        // ← not cached, and disk cache not populated
    }
}
```

and `cache/httpproxy/httpproxy.go`, `remoteHTTPProxyCache.Contains()` issues one
`http.MethodHead` per call.

So a blob that lives in the proxy but not on disk is in a stable state that nothing resolves:
the check succeeds, the client skips its upload, the blob is never fetched to disk, and the
next build repeats the whole exchange.

## Why it matters

**Latency.** These are synchronous backend round-trips on the critical path of every
`FindMissingBlobs`. At 382/s sustained and ~10–30 ms each, this is the dominant cost of a
cache lookup for anything not resident on local disk.

**Request cost.** Object-storage backends bill per operation. At this volume the existence
checks cost several times more than the uploads and downloads combined.

The effect scales with the ratio of proxy size to disk size, so it's most visible on
deployments using a proxy for capacity — which is the case the proxy exists to serve.

## Possible directions

1. **Cache positive `Contains()` results in memory** — a bounded LRU keyed by
   `(kind, hash)` with a short TTL, consulted before hitting the proxy. Positives are stable
   for as long as the backend retains the object; negatives shouldn't be cached, since a
   client may upload at any moment. This is the smallest change and benefits every proxy
   backend rather than just one.

2. **Optionally populate the disk cache on a proxy `Contains()` hit.** A digest being
   checked repeatedly is by definition in demand, and promoting it converts every later
   check into a local hit. Costs bandwidth and disk, so it would want to be opt-in.

3. `max_proxy_blob_size` already bounds proxy use, but by blob size rather than check
   frequency, so it doesn't address this.

Option 1 seems the most contained. Happy to put together a PR if you think it's a reasonable
direction — and equally happy to be told the trade-off was considered and rejected, in which
case it'd be worth a note in the docs, since the amplification isn't obvious from the
configuration.

martinpitt:
Blobs that are not in the local cache get a HEAD against the proxy backend on every request. Nothing remembers the answer, so the same digests are checked over and over. Issue #919 measured this at 94% of all backend requests.

Add an in-memory cache of "the backend has this blob", enabled with `proxy_contains_cache_ttl`. Only positive results are cached: a negative one would hide a blob that another instance uploads later, while a stale positive just fails the following fetch, same as today.

This makes the AC dependency check affordable, and hence allows admins to set a static age expiry rule in an S3 cache bucket. Without the AC check, a `cas.v2/` blob that expires before an `ac/` blob that references it will cause a build failure.

Fixes #919

----

Measured on a fairly small project (for testing), with bazel-remote as S3 proxy in front of SeaweedFS, `--disable_grpc_ac_deps_check` off, one full buck2 build of ~240 actions with ~40 action cache hits. "Cold" is a freshly started bazel-remote with an empty `--dir`; "warm" is the second identical build against the same process.

  | shim state                | `proxy_contains_cache_ttl` | HEADs to the S3 backend |
  |---------------------------|----------------------------|------------------------:|
  | cold, first build         | unset                      |                    1573 |
  | cold, first build         | 10m                        |                    1502 |
  | warm, second build        | unset                      |                     892 |
  | warm, second build        | 10m                        |                       0 |

  The cold saving is small because most digests are only checked once per build. The 892 repeats on a warm shim are blobs the AC dependency check verifies but the client never downloads, so the local disk cache never learns them and they are re-checked on every build.


martinpitt:
I'm still not fully happy with this. Ideally `proxy_contains_cache_ttl` should not exist. It's a knob for "how long should your pain last", as that's the window in which buck builds will fail with "wah wah can't get artifact" if a CAS blob expires (without its ac/ pointer), instead of this being treated as a cache miss. I'll revisit this.

martinpitt:
I retract this. bazel-remote's bucket layout really makes this kind of robustness very hard. My use case also needs to deal with much stronger guarantees wrt. adversarial cloud operators, so we'll need to sign the AC → CAS pointers, and that also needs a layout change (detached signatures are awkward).

